PR #33, opened 2022-04-02
fix: typos

fdelbrayelle:
Fix typos:
- In https://kestra.io/docs/getting-started/ local port is 28080 not 8080
- Some missing s here and there


tchiotludo:
oups didn't see that one... 
The port 8080 is the good one, it's was fixed on the based repository.

You have edited the html source, but in fact the html is generated from master on markdown file. 
Really sorry but I will need a proper contributing.md here ;)

tchiotludo:
I also think there is no more issue on that since I have large proofreading on last commit that must correct this one. 
Feel free to reopen if needed ;)

PR #94, opened 2023-01-10
fix(docs): typos and improvements

loicmathieu:


docs/architecture/README.md
```diff
@@ -26,86 +26,88 @@ More detail on configurations are available [here](../administrator-guide/config
...
-Scheduler will handle most of the [triggers](../developer-guide/triggers) (except the [Flow Trigger handle by executor](../developer-guide/triggers/flow.md)). It will continuously watch all triggers and detect if all the conditions are met to trigger an Execution (submitted to Executor).
+The Scheduler will handle most of the [Triggers](../developer-guide/triggers) (except the [Flow Trigger handle by the Executor](../developer-guide/triggers/flow.md)). It will continuously watch all triggers and, if all the conditions are met, will trigger an Execution (submitted to the Executor).
 
-Internally, a Scheduler has two ThreadPools:
+Internally, a Scheduler has two Thread Pools:
 
-- **ScheduledExecutorService** : This Threadpool will tick every second and try to find an available job.
-- **CachedThreadPool** : This Threadpool will compute in a separate Thread for all the triggers.
+- **ScheduledExecutorService**: This Thread Pool will tick every second and try to find an available job.
+- **CachedThreadPool**: This Thread Pool will compute in a separate Thread for all the triggers.
```
loicmathieu, line 80: @tchiotludo this sentence seems strange. What will be computed by this thread pool ?

loicmathieu:
The architecture document talk about Kafka and Elastisearch a lot whereas it's now possible to make Kestra run without them. It must be updated to reflect that. @tchiotludo maybe I can open an issue to update later the architecture document ?

docs/architecture/README.md
tchiotludo: Clearly, this is not clear here :zipper_mouth_face: 
The thread pool will do long polling for all except the schedule: detect files, execute query and wait for result, fetch a queue and trigger execution when there is result. 
Maybe the word compute is not the good one

tchiotludo:
> The architecture document talk about Kafka and Elastisearch
Go for the issues, it's true that we need to rework that to allow people understanding the both architecture and we never rework this one

docs/architecture/README.md
loicmathieu: Thanks, more clear now. I'll send a small PR.

PR #126, opened 2023-02-02
fix(docs): improve overview and getting started pages

tanay1337:


docs/README.md
```diff
@@ -2,36 +2,26 @@
 order: 0
 ---
 
-# What’s Kestra?
-Kestra is an orchestration & scheduler platform that helps you to build, run, schedule, and monitor complex pipelines.
+# Overview
 
+Kestra is an orchestration and scheduler platform that helps you build, run, schedule, and monitor complex pipelines. It also makes it easy to [create a flow](https://kestra.io/docs/developer-guide/), requiring only a few lines of YAML. It handles all other complexities like deploying, scaling, monitoring, and error handling.
```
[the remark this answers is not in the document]
loicmathieu: Maybe also use workflow here as it's a more general term and we use it later.
Then explain your build a worflow in a yaml called a flow

docs/getting-started/README.md
```diff
@@ -4,39 +4,46 @@ order: 1
...
 
-## Before you begin
-Make sure the followings are already installed:
-- [Docker](https://docs.docker.com/engine/install/)
-- [Docker Compose](https://docs.docker.com/compose/install/)
+## Prerequisites
 
-## Launch Kestra
+Make sure the following is installed:
+- [Docker Desktop](https://docs.docker.com/get-docker/)
```
loicmathieu, line 12: Docker desktop is not needed only Docker is. This should be clearly stated that desktop is to makes things easier

docs/getting-started/README.md
```diff
@@ -4,39 +4,46 @@ order: 1
...
-## Create your first flow
-The default installation contains some examples you can access on the `Flows` menu item.
+### Start your server
 
-We will create a new one:
+Make sure the Docker Desktop app is running. Then, run the following command:
 
-![Flow Create](./assets/flow-2.png)
+```bash
+docker-compose up -d
```
loicmathieu, line 31: docker compose command is the new docker-compose.

docs/troubleshooting/README.md
```diff
@@ -0,0 +1,15 @@
+---
+order: 9
+---
+
+# Troubleshooting
+
+## Memory error
+
+The default amount of memory available for Docker on macOS is often insufficient to get Kestra up and running with all the dependencies. If enough memory is not allocated, it might lead to the Kestra instance continuously restarting. It would be best to allocate at least 4 GB memory for the Docker Engine (ideally 8 GB). You can check and change the amount of memory in [Resources](https://docs.docker.com/docker-for-mac/#resources).
```
loicmathieu, line 9: I'm surprised we need this much memory for the Docker VM. Can someone check if it's still needed with the default JDBC plug-ins?

loicmathieu (changes requested):

docs/README.md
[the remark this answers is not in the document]
tanay1337: I'll revert the scheduling change. We should finalize whether we want to use "workflow" or "flow" as it should be consistent across the docs, marketing copy, and product copy.

docs/getting-started/README.md
tanay1337: I didn't know that. Thanks :)

docs/troubleshooting/README.md
tanay1337: @tchiotludo Thoughts?

docs/getting-started/README.md
tanay1337: I have omitted that in favor of brevity. Is the alternative Docker Engine?

loicmathieu: The alternative is just Docker. On Linux at least.
On non Linux system there is a lot of alternatives

docs/troubleshooting/README.md
tchiotludo: we don't need anymore such amount, 2Go will be sufficient now, it was an history based on kaafka + elastic that need approximately 4go

tanay1337: Should I get rid of the Troubleshooting page + that note, then?

tchiotludo: you are on mac with default config with no issue ? if yes, you can

tanay1337: That's correct. It ran smoother than butter with the default config.

tanay1337: Nice catch @loicmathieu 🙌

PR #133, opened 2023-02-03
Update Concepts - Flow

tanay1337:


docs/concepts/flows.md
```diff
@@ -1,44 +1,67 @@
...
-### Flowable Task
-[Flowable Tasks](../developer-guide/flowable) are tasks that only handle workflow states and trigger new tasks. They will handle the logic of the flow enabling complex workflow like branching, parallel tasks, etc...  
-An example is a `Switch` task that can take any inputs and decide what are the next tasks to run (and to allow branching). 
+## Task
 
-A Flowable Task is handled by `Executors` and can be called very often, so these tasks cannot include intensive computations.
+A task is a single action in a flow. A task can take properties as inputs, perform an action, and produce an output.
+
+There are two kinds of tasks in Kestra:
+- Runnable Task
```
loicmathieu, line 15: Shouldn't we use the plural here ?

docs/concepts/flows.md
```diff
@@ -1,44 +1,67 @@
...
 order: 1
 ---
+
 # Flow
-A [flow](../developer-guide/flow) is a simple list of tasks that describe the behavior of your flow. Flows are grouped by namespaces.
 
-## Task
-A Task is a single action in a flow. 
-A task can optionally take inputs, perform an action, or produce an output.
+A [flow](../developer-guide/flow) is a list of tasks. You create flows in Kestra to automate your processes.
```
loicmathieu, line 7: Here you remove the mention to namespace, in this case the Namesapce section would be better just under.
or you can add something like "Flows are grouped in namespaces"

docs/concepts/flows.md
```diff
@@ -1,44 +1,67 @@
...
+A namespace is like a folder for flows. A namespace is composed of words and letters separated by `.`. The hierarchy depth for Namespaces is unlimited. Here are some examples of namespaces:
+- `projectOne`
+- `com.projectTwo`
+- `test.projectThree.folder`
+
 
 ## Inputs 
-[Inputs](../developer-guide/inputs) are parameters sent to a flow at execution time. It can be anything (string, int, file, ...) that will be sent when the execution is created. The Flow must define possible inputs, and mandatory ones are enforced before the creation of the execution. 
+
+[Inputs](../developer-guide/inputs) are parameters sent to a flow at execution time. It can be anything, for example, a string, integer, file, etc. The flow must define any possible inputs before its execution.
```
loicmathieu, line 43: >The flow must define any possible inputs before its execution. 

The idea behind this sentence is that all inputs must be set before the flow execution, I think it should be rephrased

loicmathieu (changes requested):

docs/concepts/flows.md
tanay1337: The rationale excludes terms that have been introduced or aren't defined immediately after. Also, the definition of flows is independent of namespaces, hence the omission.

docs/concepts/flows.md
tanay1337: Doesn't it resolve to the same meaning?

tanay1337: I understand what it means now. Fixed :)

PR #146, opened 2023-03-01
feat(docs): improve onboarding documentation pages

loicmathieu:


docs/concepts/executions.md
```diff
@@ -4,29 +4,36 @@ order: 2
 
 # Execution
 
-An execution is a single run of a flow. An execution in Kestra can exist in many [states](#state).
+An execution is a single run of a flow. An execution has a [state](#state)](#state).
```
tchiotludo, line 7:
```suggestion
An execution is a single run of a flow. An execution has a [state](#state)].
```

loicmathieu: Fixed

docs/concepts/executions.md
```diff
@@ -4,29 +4,36 @@ order: 2
...
 
 ## Metrics 
 
-Each task can generate metrics data that may be useful in understanding what the current task is doing. For example, `file.size`, `row.count`, etc.
+Each task can generate metrics that may be useful in understanding what the current task is doing. For example: file size, number of rows, etc.
+
+There are two kinds of metrics: duration and counter.
 
-There are two kinds of metrics, duration, and counter. 
+Metrics can be seen in the **Outputs** tab of the **Execution** page.
```
tchiotludo, line 36:
```suggestion
Metrics can be seen in the **Outputs** tab of the **Execution** tabs.
```

docs/concepts/flows.md
```diff
@@ -20,14 +22,14 @@ There are two kinds of tasks in Kestra:
...
 Runnable Tasks handle computational work in the flow. For example, file system operations, API calls, database queries, etc. These tasks can be compute-intensive and are handled by [workers](../architecture/#worker). 
 
-By default, Kestra only includes a few Runnable Tasks. However, many of them are available as [plugins](../../plugins/).
+By default, Kestra only includes a few Runnable Tasks. However, many of them are available as [plugins](../../plugins/) and if you use our default Docker image plenty of them will be already included.
 
 
 ### Flowable Task
 
-[Flowable Tasks](../developer-guide/flowable) only handle flow logic (branching, grouping, parallel processing, etc.) and trigger new tasks. For example, the Switch task takes inputs and decides the next tasks to run. 
+[Flowable Tasks](../developer-guide/flowable) only handle flow logic (branching, grouping, parallel processing, etc.) and new task triggers. For example, the Switch task decides the next task to run based on some inputs.
```
tchiotludo, line 30: take care, don't use the word trigger for everything else than our trigger, start a new tasks ?

docs/concepts/executions.md
loicmathieu: I check and it's the `Executions` page not a tab inside a page.
I use the singular not the plural but it feels more natural as in fact we need to open an execution to have the outputs tab

docs/concepts/flows.md
loicmathieu: I changed to 'start new tasks"

PR #153, opened 2023-03-03
Describe CI/CD-related commands

yuri1969:
The recently added commands seems to lack documentation.

Feel free to move the added section to a more fitting place.


tchiotludo:
seems to be a good place for now since we don't describe any command except the server, thanks :+1:

PR #155, opened 2023-03-06
feat(docs): add a global tasks page

loicmathieu:
@tchiotludo I can add a redirect in case we want to keep current SEO rank for the flowable tasks page


tchiotludo:
Yes please add a redirect for now :+1:

PR #170, opened 2023-03-28
V0.8 new features

loicmathieu:
Maybe wait for the release to merge it, but review can already be done

### DIFF
--- a/docs/developer-guide/cicd/helpers/README.md
+++ b/docs/developer-guide/cicd/helpers/README.md
@@ -4,26 +4,17 @@ order: 4
 
 # Helpers
 
-Kestra has some _helpers_ function that can help during flow development.
+Kestra has some _helper_ functions that can help during flow development.
 
 ::: warning
 These helpers are only available during flow development to test on your local installation.
-You must expand the flow definition before sending it to your server.
+Before sending it to your server, you must expand the flow definition; our CI/CD support will automatically expand it.
 These helpers cannot be used from Kestra's UI.
 :::
- 
-## Expand the flow to be uploaded to the server
-
-There is a convenient command on the Kestra executable that allows validation of the current flow and 
-will output the expanded version of your flow without any helper:
-
-```bash
-./kestra flow validate path-to-your-flow.yaml 
-```
   
 ## `[[> file.txt]]`: Include another file
 
-When doing a large flow, the flow can be messy when many tasks are defined and especially when you have some big text inside the flow (example, sql statement, ...).
+Working on a large flow can become complex when many tasks are defined, especially when you have some big text inside the flow definition (example, SQL statement, ...).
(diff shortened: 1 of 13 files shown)


docs/developer-guide/cicd/helpers/README.md
```diff
@@ -47,17 +38,41 @@ namespace: io.kestra.tests
...
+
+## Validate the flow to be uploaded to the server
+
+There is a convenient command on the Kestra executable that allows validation of the current flow:
+
+```bash
+./kestra flow validate --local path-to-your-flow.yaml 
+```
+::: tip
+If your flow uses a helper function, flow validation must be done locally as the flow cannot be expanded on the webserver.
```
Skraye, line 68: Precise that local kestra must have plugins used in the flow to validate it.

docs/developer-guide/tasks/README.md
```diff
@@ -148,6 +148,39 @@ tasks:
     <a class="btn btn-primary" href="/plugins/core/tasks/flows/io.kestra.core.tasks.flows.Switch">Switch Task documentation</a>
 </div>
 
+### If
+
+This task processes a set of tasks conditionally depending on a condition. 
+The condition must coerce to a boolean. Boolean coercion allows 0, -0, null and '' to coerce to false,  all other values to coerce to true.
```
Skraye, line 154: Is it possible for a value to be undefined?

docs/developer-guide/tasks/README.md
```diff
@@ -261,6 +294,66 @@ tasks:
...
+    cases:
+      case1:
+        - id: case1
+          type: io.kestra.core.tasks.log.Log
+          message: Case 1
+      case2:
+        - id: case2
+          type: io.kestra.core.tasks.log.Log
+          message: Case 2
+      notexist:
```
Skraye, line 325: This example is wrong, if you don't have a default case and don't handle the case, the flow will fail anyway.
To make sense, you should have a default case. That way, you may decide to fail on a specific value and just let any other value pass.

docs/developer-guide/tasks/README.md
loicmathieu: It's Java so undefined didn't exist and will be null

docs/developer-guide/cicd/helpers/README.md
loicmathieu: Done

docs/developer-guide/tasks/README.md
loicmathieu: Well, it's not wrong but it's maybe too simplified.
I added a default, I'l also send a PR to Kestra as the same example is on the task documentation

Skraye (commented):
LGTM

PR #210, opened 2023-05-03
Clarify the difference between Metrics and Outputs 

anna-geller:
+ adding an example

### DIFF
--- a/content/docs/10.plugin-developer-guide/07.outputs.md
+++ b/content/docs/10.plugin-developer-guide/07.outputs.md
@@ -2,35 +2,35 @@
 title: Outputs & Metrics from a script engine
 ---
 
-Kestra can catch outputs and metrics from any scripting language. In the core, [Python](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Python.md) and [Node](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Node.md) plugins, inject a package to help you!
+Kestra can track outputs and metrics from any scripting language. The core [Python](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Python.md) and [Node](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Node.md) plugins provide methods to configure outputs and metrics.
 
-If you are using another language like [Bash](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Bash.md), we don't provide any package to help you, but you can do it with the `echo` command on stdout.
+If you use another language such as [Bash](../../plugins/core/tasks/scripts/io.kestra.core.tasks.scripts.Bash.md), you can leverage the `echo` command on the `stdout`.
 
 ## Script command
 
-Kestra looks at every outputs on standard out (or standard err) and search for the special patterns `::{}::` or `{}` that allow to specify outputs and metrics as JSON objects:
-- `{}` can be used for single line JSON objects.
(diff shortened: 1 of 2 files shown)


content/docs/10.plugin-developer-guide/07.outputs.md
```diff
@@ -2,40 +2,98 @@
...
+          "type": "read",
+          "location": "EU"
         }
     }
 ]
 ```
 
+## When to use metrics and when to use outputs?
+
+### Use cases for outputs
```
loicmathieu, line 41: outputs are also used a lot to pass data between tasks

content/docs/10.plugin-developer-guide/07.outputs.md
```diff
@@ -2,40 +2,98 @@
...
+
+Examples of metadata you may want to track as `outputs`: 
+
+- the **number of rows** processed in a given task that you want to use in subsequent tasks validating the number of rows, 
+- the **accuracy score** of a trained ML model in order to compare this result (output artifact) across multiple workflow executions, 
+- other pieces of **metadata** you want to track across executions of a flow (e.g. a dataset name used within a Python ETL script).
+
+### Use cases for metrics
+
+Metrics are intended to track custom numeric (metric type: `counter`) or duration (metric type: `timer`) attributes that you may want to visualize across tasks. Metrics are typically expressed as `double` data type numerical values.
```
loicmathieu, line 53:
```suggestion
Metrics are intended to track custom numeric (metric type: `counter`) or duration (metric type: `timer`) attributes that you may want to visualize across flow executions. Metrics are typically expressed as `double` data type numerical values.
```

content/docs/10.plugin-developer-guide/07.outputs.md
```diff
@@ -2,40 +2,98 @@
...
+
+Say, you are using the `EachParallel` task to process data across multiple partitions in parallel. You then want to visualize how many rows were processed in each partition and how long this process took. 
+
+Here is a simple flow demonstrating how you can accomplish that:
```
loicmathieu, line 57:
```suggestion
Here is a flow demonstrating how you can accomplish that:
```

I always advise to not use words as 'simple', 'easy', ... in documentation as something that looks easy for you may not look easy for the ready which may be a little frustrated by the fact he didn't understand something that is simple ...

loicmathieu (changes requested):

content/docs/10.plugin-developer-guide/07.outputs.md
anna-geller: yup, I'll make that more explicit

content/docs/10.plugin-developer-guide/07.outputs.md
anna-geller: it's both right?

across task runs and flow executions

loicmathieu: Yes, but I see metrics more for analytics of tasks metadata so looking at different values of the same metric on different executions instead of values of different metrics in the same flow execution.

The metrics dashboard will be at the flow level and will enable selecting a metrics of a task and graph the metric with the values of all flow executions.

anna-geller: Correct, it's on a flow execution level, but in this partition example, we also compare times between mapped child tasks (partitions) so comparing metrics across tasks seems possible

if you feel strongly about excluding tasks here, LMK

loicmathieu: You're the leader on everything documentation related so I trust you on this ;)

Both is OK for me as you said.

loicmathieu (approved):
LGTM

PR #323, opened 2023-06-23
[feature] add airbyte OSS blog post

anna-geller:


tchiotludo:
Does we need a canonical link?

anna-geller:
> Does we need a canonical link?

nice that you think about it. This will go first to Kestra's blog and later we can consider adding it on Medium or sharing it with Airbyte. This is a new post, not the one published on Airbyte post :)

Ben8t (approved):

PR #338, opened 2023-07-04
doc/ee-features

anna-geller:


components/enterprise/Features.vue
```diff
@@ -73,9 +73,9 @@
                         <div class="card-body">
                             <div class="cart-title-container">
                                 <span class="card-icon"><Security /></span>
-                                <h4 class="card-title">Best in class support</h4>
```
MartinRst, line 76: Would be nice to keep a part about Onboarding and training (instead of just support) since it's a feature sold in the EE solution

anna-geller: I agree this is helpful to mention, but this section is about features, support is not a feature. I moved the support more prominently :) 

image

anna-geller: and here 
image

MartinRst: Nice, will need a proper block in a few months to deliver more income of the training (in a section dedicated to this part)

MartinRst (approved):

PR #344, opened 2023-07-07
(blog) Release blog post for 0.10.0

anna-geller:


MartinRst (approved):

content/blogs/2023-07-10-release-0-10-blueprints-worker-groups-scripts.md
```diff
@@ -0,0 +1,299 @@
...
+    format: "{{ secret('SLACK_WEBHOOK_KEY') }}"
+```    
+
+Even though the environment variable name is prefixed with `SECRET_`, you only need to reference the key without a prefix in your flow. For more information, check the detailed [Managing Secrets](../docs/05.developer-guide/10.secrets.md) guide.
+
+
+## Improved Polling Triggers
+
+Kestra 0.10.0 significantly improves polling triggers to enhance performance, strengthen security, and simplify maintenance.
+Until now, the trigger evaluation process was handled directly by the Scheduler. The Worker and the Scheduler needed access to external systems, leading to high memory and CPU consumption.
```
Ben8t, line 242: Ludo gave the following feedback on the release note : 

Polling Triggers
leading to increased memory and CPU consumption.: it's not the right reason, the right reason is
that you have 2 services to secure since they could accept the data
the scaling is more simple since you only have the scale the worker

Maybe changing the last part "leading to high memory and CPU comsuption" 🤔

PR #393, opened 2023-07-29
Edit typo in docs about inputs

khuyentran1401:
I'm using the `kestra/kestra:latest-full` image. Using `default` for under `inputs` doesn't work, but using `defaults` under `inputs` works.


loicmathieu (approved):
LGTM, thanks

PR #409, opened 2023-08-03
(docs) Extend alerting page with flow failure notifications

anna-geller:
closes https://github.com/kestra-io/kestra.io/issues/378 

closes https://github.com/kestra-io/kestra.io/issues/354


content/docs/09.administrator-guide/03.monitoring/index.md
```diff
@@ -1,23 +1,90 @@
...
+tasks:
+  - id: fail
+    type: io.kestra.plugin.scripts.shell.Commands
+    runner: PROCESS
+    commands:
+      - exit 1
+
+errors:
+  - id: slack
+    type: io.kestra.plugin.notifications.slack.SlackIncomingWebhook
```
loicmathieu, line 28: Better to use a SlackExecution and explain that the template can be customized by using a SlackIncommingWebhook.
Each notification plugin comes with 2 tasks, one with an execution message (meaning a message that includes the information of the status of an execution) and one where the message can be customized.

content/docs/09.administrator-guide/03.monitoring/index.md
```diff
@@ -1,23 +1,90 @@
...
+Adding this single flow will ensure that you receive a Slack alert on any flow failure in the `prod` namespace. Here is an example alert notification:
+
+![alert notification](/docs/administrator-guide/alert-notification.png)
+
+
+## Monitoring 
+
+By default, Kestra exposes a monitoring endpoint on port 8081. You can change this port using the `endpoints.all.port` property in the [configuration options](../01.configuration/index.md).
+
+This monitoring endpoint provides invaluable information for troubleshooting and monitoring, including Prometheus metrics and several Kestra's internal routes.
```
loicmathieu, line 74: It may be worth talking directly about `/health` which is the healthcheck as otherwise it's at the end of the page.

content/docs/09.administrator-guide/03.monitoring/index.md
```diff
@@ -36,44 +103,43 @@ Each task type can expose [custom metrics](../../03.concepts/02.executions.md#me
...
 
 #### Indexer
 
-|Metrics|Type|Description|
-|-|-|-|
-|indexer.count|`COUNTER`|Count of indexation sent to repository|
-|indexer.duration|`DURATION`|Duration of indexation sent to repository|
+|Metrics|Type| Description                               |
+|-|-|-------------------------------------------|
+|indexer.count|`COUNTER`| Count of indexers sent to a repository      |
```
loicmathieu, line 115:
```suggestion
|indexer.count|`COUNTER`| Count of index requests sent to a repository      |
```

content/docs/09.administrator-guide/03.monitoring/index.md
```diff
@@ -36,44 +103,43 @@ Each task type can expose [custom metrics](../../03.concepts/02.executions.md#me
...
+|indexer.duration|`DURATION`| Duration of indexers sent to a repository |
```
loicmathieu, line 116:
```suggestion
|indexer.duration|`DURATION`| Duration of index requests sent to a repository |
```

content/docs/09.administrator-guide/03.monitoring/index.md
```diff
@@ -36,44 +103,43 @@ Each task type can expose [custom metrics](../../03.concepts/02.executions.md#me
...
 
-|Metrics|Type|Description|
-|-|-|-|
-|scheduler.trigger.count|`COUNTER`|Count of trigger found|
-|scheduler.evaluate.running.count|`COUNTER`|Evaluation of trigger actually running (aka: number of threads used by the scheduler)|
-|scheduler.evaluate.duration|`TIMER`|Duration of evaluation of trigger|
+|Metrics|Type| Description                                                                                         |
+|-|-|-----------------------------------------------------------------------------------------------------|
+|scheduler.trigger.count|`COUNTER`| Count of triggers                                                                                   |
+|scheduler.evaluate.running.count|`COUNTER`| Evaluation of triggers actually running (equivalent to the number of threads used by the scheduler) |
```
loicmathieu, line 123:
```suggestion
|scheduler.evaluate.running.count|`COUNTER`| Evaluation of triggers actually running|
```

The last part is no longer accurate.

loicmathieu (changes requested):
On the Alerting and Monitoring page, one day we should provide a "Sanity Check" flow and explain that Kestra can be monitored externally by calling periodically this flow (for ex with a webhook). This is a very common way of monitoring and we should provide a template for that.

content/docs/09.administrator-guide/03.monitoring/index.md
anna-geller: yup, the SlackExecution only works in triggers and listeners, that's why here we use IncomingWebhook. It doesn't work on `errors`. Please share a flow example if this is not true. 

Regarding:
> On the Alerting and Monitoring page, one day we should provide a "Sanity Check" flow and explain that Kestra can be monitored externally by calling periodically this flow (for ex with a webhook).

do you want to open a dedicated issue? I was convinced that such monitoring flow is not needed because of the already existing healthcheck endpoint http://localhost:8081/health

content/docs/09.administrator-guide/03.monitoring/index.md
anna-geller: good idea, let me add it

content/docs/09.administrator-guide/03.monitoring/index.md
loicmathieu: > SlackExecution only works in triggers and listeners

OK then.

> do you want to open a dedicated issue? I was convinced that such monitoring flow is not needed because of the already existing healthcheck endpoint http://localhost:8081/health

It tests more than just "the port is open", it can be a complete sanity check that checks DB connectivity, queue functionality, internal storage, ... I'll open an issue for that.

loicmathieu (approved):
LGTM

content/docs/09.administrator-guide/03.monitoring/index.md
anna-geller: ahh I see, that makes sense, thanks so much!

PR #466, opened 2023-08-28
Update UI Guide

Ben8t:


anna-geller (approved):
Nice, thanks for adding that.

It would be nice if you could redo the screenshots as we changed the navigation bar on the left. Other than that 👍

PR #473, opened 2023-08-31
fix(docs) labels syntax

anna-geller:


tchiotludo:
Now the old syntax works again, see kestra-io/kestra#2006, we would prefer to keep the old one on doc I think

PR #534, opened 2023-09-26
feat(): JDBC heartbeat doc

Skraye:
Related to :
https://github.com/kestra-io/kestra/pull/2081

### DIFF
--- a/content/docs/09.administrator-guide/01.configuration/04.others.md
+++ b/content/docs/09.administrator-guide/01.configuration/04.others.md
@@ -280,3 +280,21 @@ kestra:
         }
       </style>
 ```
+
+
+## JDBC Healthcheck
+
+JDBC have a healthcheck that will check every worker at every `frequency`.
+
+* If the last healthcheck is older than `heartbeat-missed * frequency`, the worker will be marked as DEAD.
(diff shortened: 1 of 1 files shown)


content/docs/09.administrator-guide/01.configuration/04.others.md
```diff
@@ -280,3 +280,21 @@ kestra:
         }
       </style>
 ```
+
+
+## JDBC Healthcheck
+
+JDBC have a healthcheck that will check every worker at every `frequency`.
+
+* If the last healthcheck is older than `heartbeat-missed * frequency`, the worker will be marked as DEAD.
```
anna-geller, line 289:
```suggestion
* If the last healthcheck is older than `heartbeat-missed * frequency`, the worker will be marked as unhealthy.
```
is it necessary to say DEAD?

anna-geller (approved):

content/docs/09.administrator-guide/01.configuration/04.others.md
Skraye: Well, it is the exact status, but unhealthy works too I guess

PR #536, opened 2023-09-26
create plugins page

mageexcel:


anna-geller (changes requested):
Plugins and subgroups don't render well atm 

image

mageexcel:
> Plugins and subgroups don't render well atm

Hi, @anna-geller , Actually not getting the correct icon name that why icons are not rendering and also have to do some more fixes

brian-mulier-p:
Closing in favor of #762

PR #603, opened 2023-10-25
fix(docs): avoid mismatched subplugin versions

anna-geller:
@loicmathieu feel free to fix directly on my branch if I misunderstood how that works. This is meant to prevent issues like this one https://github.com/kestra-io/kestra/issues/2379#issuecomment-1778890667


loicmathieu (approved):
LGTM

PR #657, opened 2023-11-23
Document common trigger properties

yuri1969:
Common properties inherited from `AbstractTrigger` were not documented in the main docs nor the generated docs.


anna-geller (approved):
Thanks a lot!

Issue #645, opened 2023-11-16
[UI] Make the Terraform page responsive


anna-geller:
### Feature description

![image](https://github.com/kestra-io/docs/assets/86264395/4fcc1f1e-61f3-4ee7-b10c-782e97bf58a2)

PR #661, opened 2023-11-28
Fix teraform UI

mageexcel:
closes https://github.com/kestra-io/docs/issues/645


anna-geller:
is it meant to close https://github.com/kestra-io/docs/issues/645? cc @MartinRst to verify

PR #672, opened 2023-12-01
Image issue

mageexcel:


components/docs/Container.vue
```diff
@@ -110,6 +110,9 @@
     font-weight: 600;
     line-height: 2.375;
 }
+:deep(img){
+    width: 100%;
+}
```
rldorado, line 115: This will lead to this issue (I've already face it 😄)



I'm thinking another solution. Feel open to share your thoughts.

mageexcel: thanks, They are using same component

PR #694, opened 2023-12-13
Fix typo in 01.getting-started.md

sleungtoast:
Should be using `docker-compose`


anna-geller (approved):
both works, we can add this version, thx  👍

sleungtoast:
Ah I see it's the new syntax in Compose V2 https://docs.docker.com/compose/migrate/

PR #722, opened 2023-12-27
fix(docs): corrected multiple docs and inputs doc file

shrutimantri:
Corrected multiple docs, and major changes are in the 04.inputs.md file.

### DIFF
--- a/content/docs/05.developer-guide/04.inputs.md
+++ b/content/docs/05.developer-guide/04.inputs.md
@@ -66,43 +66,43 @@ inputs:
 
 Inputs in Kestra are strongly typed and validated before starting the flow execution.
 
-Here is a list of supported data types:
+Here is the list of supported data types:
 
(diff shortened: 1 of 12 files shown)


content/docs/05.developer-guide/04.inputs.md
```diff
@@ -258,8 +258,6 @@ with open("/tmp/example.txt", 'rb') as fh:
   mp_encoder = MultipartEncoder(fields={
     "string": "a string",
     "optional": "an optional string",
-    "int": 1,
-    "float": 1.255,
```
shrutimantri, line 262: These 2 lines lead to error:
AttributeError: 'int' object has no attribute 'encode'
AttributeError: 'float' object has no attribute 'encode'

Hence, removed these.

brian-mulier-p: Please use
```
"int": str(1),
"float": str(1.255),
```
instead

brian-mulier-p (changes requested):

PR #711, opened 2023-12-26
added api call for get plugins

t-chatoyan:

PR #762, opened 2024-01-08
fix/plugins-hp

brian-mulier-p:
Cleaner version of PR #711


components/blueprints/Lists.vue
```diff
@@ -94,7 +95,7 @@ watch([currentPage, itemsPerPage, activeTag, searchQuery], ([pageVal, itemVal, t
     }
     timer = setTimeout(async () => {
 
-        const { data } = await useFetch(`https://api.kestra.io/v1/blueprints?page=${(itemVal != oldItemVal) || (tagVal != oldTagVal) ? 1 : pageVal}&size=${itemVal}${Object.keys(tagVal).length && tagVal.name != 'All tags' ? `&tags=${tagVal.id}` : ''}${searchVal.length ? `&q=${searchVal}` : ''}`)
+        const { data } = await useFetch(`${config.public.apiUrl}/blueprints?page=${(itemVal != oldItemVal) || (tagVal != oldTagVal) ? 1 : pageVal}&size=${itemVal}${Object.keys(tagVal).length && tagVal.name != 'All tags' ? `&tags=${tagVal.id}` : ''}${searchVal.length ? `&q=${searchVal}` : ''}`)
```
ridge-kimani, line 98: You could use `querystring` module to build the whole query.

components/plugins/PluginsLists.vue
```diff
@@ -0,0 +1,394 @@
...
+                    </select>
+                </div>
+                <div class="d-flex justify-content-between align-items-center">
+                    <CommonPagination
+                        :totalPages="totalPages"
+                        @on-page-change="changePage"
+                        v-if="totalPages > 1"
+                    />
+                    <div class="d-flex align-items-baseline" v-if="totalPlugins > itemsPerPage">
+                        <span class="total-pages">Total {{ totalPlugins }}</span>
```
ridge-kimani, line 58: How about `{{totalPlugins}} Plugins`?

components/blueprints/Lists.vue
brian-mulier-p: I think it would be noise on this PR, maybe you can create an issue to unify this on every api calls ?

Skraye (changes requested):
image
This only allow search in plugin title, could it be possible to search across content or at least in sub task ? 
Its written search across 400 tasks but in fact it search only plugin group, maybe we could use the search from the global search but filtered on the plugins category ? 

And personal POV : those cards are too white my eyes are bleeding, maybe we could have a darker border for it ?

components/plugins/PluginsLists.vue
brian-mulier-p: image
I think I will add ':' to align with what format we use on the UI

components/blueprints/Lists.vue
ridge-kimani: The querystring would make it a bit easier to destructure.

```js
import querystring from querystring

const qs = querystring.stringify({
  page: currentPage.value,
  size: itemsPerPage.value,
  tags: route.query.tags ? activeTag.value.id : undefined,
  q: route.query.q ? searchQuery.value : undefined,
});

// Output: "page=1&size=10&tags=selectedTagID&q=searchQueryValue"

```

components/content/ProseA.vue
```diff
@@ -25,36 +25,37 @@
...
+            }
+
+            absolutePath = absolutePath + "/";
         }
 
+        link = new URL(link, absolutePath).toString();
+
         if (link.endsWith('/')) {
             link = link.replace(/\/$/, '')
         }
```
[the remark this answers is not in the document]
brian-mulier-p: Fixed :+1: 

Great idea for the global search filtered on plugins trick, will try to do that. I took existing code and didn't notice that the filter was done on front-end side btw so I'll at least switch to a backend filtering :+1:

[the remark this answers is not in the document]
brian-mulier-p: In the end the global search won't fit for that need (category is not a filterable field and it would not fit with other facet types)

anna-geller:
awesome work! initially we wanted 6 plugins per row, can we adjust that?

image

brian-mulier-p:
> awesome work! initially we wanted 6 plugins per row, can we adjust that?
> 
> image

Yes I spoke to Nico and will do that

anna-geller:
it looks like all major issues have been addressed and it looks ready to be shipped!

@Skraye good point about the search, we can tackle it in the next iteration - follow-up issue: https://github.com/kestra-io/docs/issues/772. regarding cards being too white: this will change anyway with website dark mode in the future so we don't need to worry about it for now (totally understand your point though 👍 )

PR #766, opened 2024-01-09
added functionality for get sitemap urls

t-chatoyan:


server/api/sitemap.ts
```diff
@@ -0,0 +1,46 @@
...
+            data: response.results.map(page => ({
+                loc: `${url.pageName}${page.id}-${slugify(page.title)}`,
+            }))
+        };
+    };
+
+    const fetchAllPages = async () => {
+        const allData = [];
+
+        for (const url of urls) {
```
ridge-kimani, line 31: You could use `Promise.all` syntax.

t-chatoyan: Thanks for the review, added.

> You could use `Promise.all` syntax.

brian-mulier-p (changes requested):
Move the dependency to the new one please :+1: 

And I typed everything in the API to remove any warnings, please in the future add types to prevent such warnings :+1:

nuxt.config.ts
```diff
@@ -7,7 +7,9 @@ export default defineNuxtConfig({
       'nuxt-simple-sitemap',
```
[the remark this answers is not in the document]
t-chatoyan: sure, I saw it also, just used it because we had it in the project.

PR #767, opened 2024-01-09
Fix sticky position

t-chatoyan:


ridge-kimani:
It's misplaced in mobile devices.
Screenshot 2024-01-10 at 11 52 38

t-chatoyan:
> It's misplaced in mobile devices. Screenshot 2024-01-10 at 11 52 38

I have fixed this issue

assets/styles/docs.scss
```diff
@@ -142,6 +142,32 @@ $bd-gutter-x: 3rem;
...
+        }
+    }
+    .bd-main {
+        .bd-toc {
+            display: none !important;
+
+            @media (max-width: 992px) {
+                display: block !important;
+                grid-area: toc !important;
+            }
```
ridge-kimani, line 166: Do we need `!important`? I don't see any display overrides that would prevent it from rendering as required.

brian-mulier-p:
close in favor of https://github.com/kestra-io/docs/pull/773 to handle all margins in a better way

PR #770, opened 2024-01-10
Improve Enterprise Edition page

anna-geller:


MartinRst (approved):
LGTM only the listing with Namespace-level in the first frame feel weirds to have the same wording many time maybe change for: Namespace-level management and detail has you did in the table?

PR #833, opened 2024-01-29
Contact us design

t-chatoyan:


brian-mulier-p (changes requested):
Please change the form's font color to white to match Figma + it lacks the two shadowy purple things

Issue #815, opened 2024-01-23
Terraform subpages are missing 


anna-geller:
### Expected Behavior

e.g. https://kestra.io/docs/terraform/data-sources 
![image](https://github.com/kestra-io/docs/assets/86264395/e8dbd180-7726-4378-8761-2ae12a512578)

### Environment Information

...

tchiotludo:
the fix is falling others part of the website, reverting ...

PR #852, opened 2024-02-05
02/05 terraform page

t-chatoyan:


components/docs/NavSideBar.vue
```diff
@@ -51,6 +52,15 @@
                 type: Object,
             },
         },
+        data() {
+          return {
+              disabledPages: [
+                  '/docs/terraform/data-sources',
+                  '/docs/terraform/guides',
+                  '/docs/terraform/resources'
+              ]
```
brian-mulier-p, line 61: Please can you add some screenshots of why you have to do that ? I don't really like the fact that we have arbitrary content-related things in a generic component

brian-mulier-p (changes requested):

components/docs/NavSideBar.vue
t-chatoyan: https://github.com/kestra-io/docs/issues/815
for those pages we don't have content and we need to disable that pages

brian-mulier-p (changes requested):
image
The dropdown disappeared completely while there was some pages underneath (see kestra.io)
image

PR #915, opened 2024-02-21
docs: proxy configuration

anna-geller:


krusaderjake:
@anna-geller from what I could see it looks like Kestra relies on this specific point on the Microsoft VS Code networking notes which might be good to make visible for users in their firewall explorations:

*.vscode-unpkg.net - Used when loading web extensions

^^that path is what Kestra loads resources from for the Editor to work when behind a proxy/firewall combination.

anna-geller:
thanks so much, @krusaderjake, I added a commit based on your feedback. will merge now 👍

PR #918, opened 2024-02-22
feat(docs): creating installation doc for azure aks

shrutimantri:


anna-geller (approved):
Great work, thanks so much!

PR #977, opened 2024-03-01
fixed blog page

t-chatoyan:


Nico-Kestra:
So I have the impression that the original issue, that is to say, having an image that takes up the entire width, hasn't been merged? @anna-geller?

PR #997, opened 2024-03-05
Visual and text changes

t-chatoyan:


components/card/Card.vue
```diff
@@ -48,6 +56,12 @@
             number: {
                 default: null,
             },
+            bottomMenuBar: {
+              default: null,
```
brian-mulier-p, line 60:
```suggestion
              type: Array
              default: [],
```

components/card/Card.vue
```diff
@@ -11,6 +11,14 @@
                 </span>
                 <h4 v-if="title" class="card-title">{{title}}</h4>
             </div>
+            <div v-if="bottomMenuBar && bottomMenuBar.length" class="bottom-menu">
```
brian-mulier-p, line 14:
```suggestion
            <div v-if="bottomMenuBar.length > 0" class="bottom-menu">
```

components/card/Card.vue
```diff
@@ -11,6 +11,14 @@
...
+                <div class="bottom-menu-item" v-for="menuItem in bottomMenuBar" @click="() => menuItem.active = !menuItem.active">
+                    <svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 25 25" fill="none">
+                        <path d="M14.9891 17.3359V13.3359H6.06906L6.03906 11.3259H14.9891V7.33594L19.9891 12.3359L14.9891 17.3359Z" :fill="menuItem.active ? '#CDD5EF' : '#9CA1DE'"/>
+                    </svg>
+                    <span :style="`color : ${menuItem.active ? '#CDD5EF' : '#9CA1DE'}`">{{ menuItem.text }}</span>
```
brian-mulier-p, line 19: Magic colors, please use some classes instead

components/enterprise/Security.vue
```diff
@@ -1,88 +1,249 @@
...
+            @include media-breakpoint-down(md) {
+                :deep(.card) {
+                    width: auto;
+                }
+            }
+            &::before {
+                content: "";
+                position: absolute;
+                width: calc($spacer * 12.5);
+                height: calc($spacer * 12.5);
```
brian-mulier-p, line 235: Example of magic numbers

brian-mulier-p:
I think there is too many magic numbers. Are you sure we don't want to use percentages or other unit on {magicNumbers}*$spacer values ? It looks super hacky

brian-mulier-p (changes requested):
Please reduce the amount of magic numbers in favor of other units (that make sense)

PR #1026, opened 2024-03-11
feat(docs): include output preview in the outputs doc

shrutimantri:
While we will still add "How-to Guide" with a detailed example, add a small section on Render Expression usage in the Outputs doc.


content/docs/06.workflow-components/07.outputs.md
```diff
@@ -242,8 +250,31 @@ Note how the `outputs` are set twice within the `"{{outputs.subflow.outputs.fina
...
+```yaml
+id: json_values
+namespace: dev
+tasks:
+- id: sample_json
+  type: io.kestra.core.tasks.debugs.Return
+  format: '{"data": [1, 2, 3]}'
+```
+
+On running this flow, the **Outputs** tab would have the output for the `sample_json` task as shown.
```
anna-geller, line 270:
```suggestion
When you run this flow, the **Outputs** tab will contain the output for the `sample_json` task, as shown below:
```

content/docs/06.workflow-components/07.outputs.md
```diff
@@ -242,8 +250,31 @@ Note how the `outputs` are set twice within the `"{{outputs.subflow.outputs.fina
...
+
+![json_values](/docs/workflow-components/outputs/json_values.png)
+
+You can select the task from the drop-down, in this case we will select "sample_json", and click on **Render expression**.
```
anna-geller, line 274:
```suggestion
You can select the task from the drop-down menu. Here, we will select "sample_json" and click on **Render expression**:
```

PR #1075, opened 2024-03-20
fixed blueprints page design

t-chatoyan:


anna-geller (changes requested):
wait, there is a regression with trigger topology, can you check?

image

image

anna-geller (approved):
the trigger issue is unrelated to the PR so good to merge 👍

PR #1087, opened 2024-03-22
Feat 0322/bluerpints cards carusel

t-chatoyan:


components/layout/BluerpintsCardsCarusel.vue
```diff
@@ -0,0 +1,173 @@
...
+            </Carousel>
+        </div>
+    </div>
+</template>
+<script setup>
+  const config = useRuntimeConfig();
+  const blueprints = ref([])
+
+  const { data: blueprintsData } = await useAsyncData('blueprints', () => {
+    return $fetch(`${config.public.apiUrl}/blueprints`)
```
tchiotludo, line 36: You need to pass the task type here as query string `types`

tchiotludo (changes requested):

components/layout/BluerpintsCardsCarusel.vue
t-chatoyan: @tchiotludo can you show me how should look query ?

tchiotludo: `${config.public.apiUrl}/blueprints?type=${cls}`
with cls = "io.kestra.plugin.jdbc.pinot.trigger" for examples

PR #1098, opened 2024-03-26
Feat 0326/pricing

t-chatoyan:


Nico-Kestra:
Can you reduce the line spacing between text ?
Capture d’écran 2024-03-26 à 15 16 48
@t-chatoyan

PR #1121, opened 2024-04-01
Feat 0401/optimization

t-chatoyan:


brian-mulier-p (changes requested):
While you're at it please have a look at some SEO things like missing links' attributes 
image

PR #1127, opened 2024-04-02
refactor(docs): Remove duplicate pages in "Getting Started" section

wrussell1999:
I've gone through and removed pages where they're just mirroring content from the index page for the later sections as this doesn't provide any value other than cluttering up the navigation menu as well as being unrelated to just getting started with Kestra. I will come back and dig deeper into the content and structure of this section as a separate PR.


wrussell1999:
This removes all the cards on the docs home page as they use Getting Started to generate them. Worth thinking about a few things here:
- What cards do we want on the home page that isn't overwhelming. Potentially a little less than just exact replicate to the side bar
- Think about how this can be implemented without requiring getting started to be filled with the options

wrussell1999:
https://github.com/kestra-io/docs/issues/1130 to look into the issue faced with "Getting Started"

anna-geller (approved):
🎉

PR #1162, opened 2024-04-09
feat(docs): pebble templating with namespace files doc

shrutimantri:


anna-geller:
I'll merge as is for now but in 0.16 the behavior changes, we'll adjust then

PR #1178, opened 2024-04-10
Feat 0410/blog

t-chatoyan:


pages/blogs/[...slug].vue
```diff
@@ -31,6 +31,11 @@
                                 :src="page.image"
                                 fit="cover"
                             />
+                            <div class="subtitle">
```
Comment (Skraye) [inline pages/blogs/[...slug].vue:34]: Can you add a `v-if=page.description` ? Just in case

Skraye (changes requested):

PR #1195, opened 2024-04-12
refactor(docs): add order to architecture section

wrussell1999:
Added numbers to the architecture files so they display in a more appropriate order rather than in alphabetical order. Order was decided based on the order on the server components page.


anna-geller (approved):
just cross-check broken links as I believe those server components are referenced many times in Configuration Guide and likely many other places so if you rename it, you will need to rename all references

you can try using a tool like https://www.brokenlinkcheck.com/ to spot these

PR #1198, opened 2024-04-12
feat(docs): Migrate Python and Webhook Flow Examples to How to Guides

wrussell1999:
Starting to improve the example and how to guides by combining them and adding more information to help you understand the examples. Also broken down the example into stages so it's easier to digest, introducing new concepts one at a time.

Starting with:
- Python
- Webhooks

Need to add file outputs to the Python example for this to be ready.


anna-geller:
Smart idea to have generic pages on how to use python or webhooks 🧠

PR #1206, opened 2024-04-15
refactor(docs): Migrate plugin section to plugin page in developer guide

wrussell1999:
Moved the plugins section into one page under Developer Guide. As a result of removing a number, I've renumbered the other sections to match. This won't break URLs.


anna-geller:
Maybe you can add a header within each toggle? e.g. the Documentation section is shown on the right sidebar without the parent section "Develop a trigger"

anna-geller (approved):
awesome!

PR #1215, opened 2024-04-16
refactor(docs): Migrate Errors example to main errors page 

wrussell1999:
I was originally going to migrate errors from Flow Examples -> to How-To Guides but determined that this made more sense belonging on the errors page directly.

I will continue to look at other flow examples and determine if they need a dedicated guide or if they should just compliment existing documentation (especially when the example is small and only highlighting one thing, like errors).


content/docs/06.workflow-components/11.errors.md
```diff
@@ -3,29 +3,21 @@ title: Errors
 icon: /docs/icons/components.svg
 ---
 
-The `errors` is a list of tasks set on a flow-level that will be executed when an error occurs. You can add multiple such tasks, and they will be executed sequentially.
+There's multiple ways to manage errors in Kestra, to both help you identify them but also allow flows to continue to operate around errors.
```
anna-geller, line 6:
```suggestion
There are multiple ways to handle errors in Kestra, to both help you identify them and allow flows to continue to operate despite errors.
```

content/docs/06.workflow-components/11.errors.md
```diff
@@ -3,29 +3,21 @@ title: Errors
...
+
+## `errors` Component
+
+`errors` is a list of tasks set on a flow-level that will be executed when an error occurs. You can add multiple tasks, and they will be executed sequentially. This is useful for being able to send alerts when errors occur.
```
anna-geller, line 10:
```suggestion
`errors` is a list of tasks set on a flow level that will be executed when an error occurs. You can add multiple tasks, and they will be executed sequentially. This is useful for sending alerts when errors occur.
```

content/docs/06.workflow-components/11.errors.md
```diff
@@ -3,29 +3,21 @@ title: Errors
...
-      {
-        "channel": "#alerts",
-        "text": "Failure alert for flow {{ flow.namespace }}.{{ flow.id }} with ID {{ execution.id }}"
-      }
+```yaml file=public/examples/flows_errors_1.yaml
+```
+
+## `allowFailure` Property
+
+When you execute a flow and one of the tasks fail, it will no longer proceed stopping later tasks from executing. This can be annoying if the task that failed doesn't impact other tasks, such as sending a message or a log output. We can resolve this by adding the `allowFailure` property to our task which will let it continue and pass the execution as `WARNING` in the Logs.
```
anna-geller, line 20:
```suggestion
When you execute a flow and one of its tasks fails, downstream tasks won't be executed. This may not always be desirable, especially for non-critical tasks. You can resolve this by adding the `allowFailure` property to the task, which will allow downstream tasks to continue despite the error. In this case, the execution will end in a `WARNING` state.
```

anna-geller (approved):

PR #1226, opened 2024-04-17
feat(docs): blog on kestra vs airflow for orchestrating python scripts

shrutimantri:


content/blogs/2024-04-25-kestra-vs-airflow-orchestrating-python-scripts.md
```diff
@@ -0,0 +1,152 @@
...
+description: Print Hello World
+
+tasks:
+  - id: python-script
+    type: io.kestra.plugin.scripts.python.Script
+    script: |
+      print("Hello World")
+```
+
+Not a major difference, isn't it?! But as the complexities creep in, you would start noticing some major differences.
```
Ben8t, line 50: I'll cut that one - to me there is an important difference between the two (Airflow is way more verbose)

Ben8t (commented):
Little adjust, anyway: great stuff 💪 thx!

anna-geller:
misunderstanding on the topic

PR #1237, opened 2024-04-18
Fixed orders.csv uri

sinsniwal:
This pull request is related to issue #1238 
we can see the old URL being used here
https://kestra.io/docs/developer-guide/scripts/outputs-metrics


wrussell1999 (approved):
Looks great, thanks for fixing this!

Issue #1253, opened 2024-04-22
Add links for each feature on the pricing page


wrussell1999:
### Description

Link to each part of the enterprise edition so it's easy to understand what it does.

![Screenshot 2024-04-22 at 13 36 19](https://github.com/kestra-io/docs/assets/34094921/96973115-0f51-4920-8c29-2b473c780241)

PR #1254, opened 2024-04-22
feat: add urls to editions page

wrussell1999:
URLs for features under the enterprise edition. Some links didn't have dedicated pages (because of what they are).

Screenshot 2024-04-22 at 14 30 29


wrussell1999:
Solves: https://github.com/kestra-io/docs/issues/1253

MartinRst:
It's really weird to have links on a table for a pricing page, it's not meant to be linked to the doc and add complexity. Maybe add a tooltip and simplify the table below would be better

wrussell1999:
> It's really weird to have links on a table for a pricing page, it's not meant to be linked to the doc and add complexity. Maybe add a tooltip and simplify the table below would be better

The goal was to allow people to click in and understand what the features actually are. Agree that the docs themselves might be too complicated, but some sort of features page with linkable headers could be a better choice?

PR #1261, opened 2024-04-25
fix(docs): Improve workflow components section

wrussell1999:
Improved the flow and structure of the workflow components section in preparation of the workflow components video series. Home page now includes a high level summary, expandable table will more detail, and 2 expandable examples followed by the list of cards. Wording for the cards has been made more concise too.

Screenshot 2024-04-25 at 12 19 08

## Changes
- Move Flow page into the Index page
- Flow Execution sub pages migrated to Execution Page
- Migrate flow properties page into the Index page
- Combine the trigger section into one page for easier navigation and fix cards on Index page
- Add UI images for concurrency with an additional example  
- Add examples types to Flowable and Runnable tasks to make it clearer
- Add URLs to link concepts/components back to their pages
- General wording tweaks throughout

## Thoughts
- Subflows feels like it would fit more naturally into Concepts as it's the only thing (aside from execution) that isn't a flow property under workflow components, especially because Flowable and Runnable tasks are under Concepts. I think it's fine where it is for now but could be worth moving at a later date.


wrussell1999:
Updated to move cards up

Screenshot 2024-04-25 at 16 10 07

wrussell1999:
Change icons before merging

wrussell1999:
Icon updated
image

anna-geller:
nice!

image

PR #1281, opened 2024-04-30
feat(docs): Reorder Sections and fix URLs

wrussell1999:
As discussed in Slack, docs sections to be re ordered so that workflow components is in a more logical place (both from a GA perspective but also learning journey).

Additionally fix all links to use actual paths rather than markdown file path. This is easier to follow, but also helps avoid outdated links when numbers in filenames are changed to alter the order. Also fixed some dead links from previous structures as well as adding more redirects in.

**Old layout**
Welcome to Kestra
Getting Started
Installation Guide
Tutorial
Architecture
Kestra Enterprise
Workflow Components
Concepts and Expressions

**New layout**
Welcome to Kestra
Getting Started
Installation Guide
Tutorial
Workflow Components
Concepts and Expressions
Kestra Enterprise
Architecture
Screenshot 2024-04-30 at 12 28 33


wrussell1999 (commented):
Cross checked URL changes

PR #1283, opened 2024-04-30
Custom Plugin build step updated in the documentation

phanikiranwtc:
Updated the plugin build step to use kestra:develop-full instead of kestra:latest-full docker as base docker image tag


wrussell1999:
@phanikiranwtc can you explain the reason for this change? I would've thought that the `latest-full` tag would've been preferable for most as it's more stable when using your plugin.

PR #1285, opened 2024-04-30
feat(docs): Add defaults to inputs docs for better clarity

shrutimantri:


content/docs/06.workflow-components/05.inputs.md
```diff
@@ -35,25 +35,38 @@ namespace: example
...
     type: INT
+    defaults: 100
 
   - id: bool
     type: BOOLEAN
+    defaults: true
 
   - id: float
     type: FLOAT
+    defaults: 100
```
wrussell1999, line 54: Can we change this to include a decimal point so it's different to the INT? 100.3 for example

wrussell1999 (commented):
LGTM!

content/docs/06.workflow-components/05.inputs.md
shrutimantri: ohh sure.. looks to be a miss..

wrussell1999 (approved):

PR #1345, opened 2024-05-15
fix(docs): correct the sentence grammatically

shrutimantri:


wrussell1999:
Discussed in Slack but Get Stuck is present tense so is correct for this use case.

PR #1353, opened 2024-05-16
add updates to reverse proxy with context-path

paulgrainger85:


content/docs/10.configuration-guide/url.md
```diff
@@ -86,3 +86,38 @@ location / {
...
+
+```yaml
+micronaut:
+  server:
+    context-path: "/kestra"
+```
+
+Then, modify your above nginx configuration to the following
+
+```nginx
```
wrussell1999, line 101: `nginx` isn't a supported Markdown language so comes up on the website all in one colour. Changing to bash seems to get the right sort of syntax highlighting

```suggestion
```bash
```
Screenshot 2024-05-16 at 14 38 38

wrussell1999 (approved):

PR #1373, opened 2024-05-21
Update Tag page to use Kestra YAML instead of HTML Tags

### LINKED ISSUE #1352: Update Tag page to use Kestra YAML instead of HTML Tags
### Description

> These examples, for example, the [Tags](https://kestra.io/docs/concepts/expression/tag), are like HTML templates. I feel like it would be more useful to be YAMLs, and users would be able to connect the dots more.

https://kestra.io/docs/concepts/expression/tag

shrutimantri:
Closes #1352


wrussell1999 (approved):

wrussell1999:
@shrutimantri going to merge this in but would love to see some examples of where this is used in a flow as well! :D Great work!

PR #1382, opened 2024-05-22
fixed video-container

t-chatoyan:


assets/styles/docs.scss
```diff
@@ -213,10 +213,20 @@ $bd-gutter-x: 3rem;
 
 
     .video-container {
-        padding-top: 56.25%;
         position: relative;
+        padding-top: 35.25%;
+        background-color: $black-3;
```
wrussell1999, line 218: Can we leave the background to match the rest of the page?

Screenshot 2024-05-22 at 15 50 03

wrussell1999 (commented):
Looks good. Can we remove the grey background on larger screens so it's always transparent? Looks odd having a box sometimes but not all the time. Thought it was something selected in the video 😅  @t-chatoyan

PR #1384, opened 2024-05-22
fix(docs): reflect Micronaut 4 config changes

yuri1969:
See [Micronaut Security 4.0 breaking changes](https://micronaut-projects.github.io/micronaut-security/latest/guide/#breaks)


anna-geller (approved):
thx a lot!

PR #1391, opened 2024-05-23
Standardize icons size

t-chatoyan:


Nico-Kestra:
- [ ] Add spacing after the last category.
Capture d’écran 2024-05-23 à 14 04 26

- [ ] The visuals are still pixelated on iPhone.
Capture d’écran 2024-05-23 à 14 06 26

- [ ] Blueprints: visual needs to be repositioned to the left to avoid cutting off the text.
Capture d’écran 2024-05-23 à 14 07 38

PR #1393, opened 2024-05-23
fix(docs): Update outputs tutorial to use `outputFiles` instead of `outputDir`

wrussell1999:
Updated the 3 examples to use `outputFiles` instead of `outputDir`. Also updated the http plugin type to use `core`


anna-geller (approved):
That's perfect 🚀 let's only make sure to push all these changes only to develop.kestra.io until the release

wrussell1999:
Replaced by https://github.com/kestra-io/docs/pull/1397 which will merge into `develop` without conflicts

PR #1411, opened 2024-05-28
Added hideSidebar attribute to  index.md content to hide in nav-menus

t-chatoyan:


anna-geller (approved):
works <3 thx so much!

PR #1467, opened 2024-06-10
fix: Update readme gif video

wrussell1999:
Quick 30 second overview using a simple example flow. Experimenting with the format due to the low resolution and frame rate. We can also make new ones easily in the future.

Next step will be the ones in https://github.com/kestra-io/kestra


anna-geller (approved):

wrussell1999:
@anna-geller added the one for https://github.com/kestra-io/kestra

PR #1510, opened 2024-06-27
Move each plugin index page from Markdown to model-based doc

t-chatoyan:
seems part of https://github.com/kestra-io/kestra/issues/4154


brian-mulier-p:
I'm starting from scratch as the strategy to handle it totally changed since then

PR #1607, opened 2024-07-26
docs: template to see if the new fields are working

anna-geller:


anna-geller:
done in https://github.com/kestra-io/docs/commit/a41d5a97e29150ca6e652501be26a090abbb6c67

PR #1647, opened 2024-08-18
fix(docs): update inputs section

varholak-peter:


anna-geller (approved):
thanks so much for spotting that old name here!

PR #1667, opened 2024-08-26
Fix disabling postgres in kubernetes

pavel-mxsf:
There is a mistake in Kubernetes docs for cloud providers.


anna-geller (approved):
thanks a lot!

PR #1682, opened 2024-08-30
fixed menu

t-chatoyan:


t-chatoyan:
can you check on mobile the menu dropdown the close button does not work (the X does but not clic on close) and we have a huge delay to display the menu links and logo

PR #1683, opened 2024-08-30
fixed header image

t-chatoyan:
something with the new homepage, during the loading, I often see this for few seconds, anything to have a proper layout during the loading ?
image


tchiotludo:
Still some blinking, and it's showing a not found image during few seconds 
[Screencast_20240830_172201.webm](https://github.com/user-attachments/assets/339a0631-d223-4a28-a461-a4b8e17e2d89)

Issue #1694, opened 2024-09-04
Implement a new layout for the How-To Guide on the Website (Docs section)


anna-geller:
### Description

Figma design https://www.figma.com/design/g8Di3M1gbajOjOqp0ZrzXO/02_Kestra.io?node-id=12960-13588&t=qT4Km1eGWoRd9MLL-1 

![image](https://github.com/user-attachments/assets/7fa9afb8-2483-4174-90ce-cd6ae2d42e81)

The frontmatter fields that we'll need for the Filters are: 
- `stage` — a single string that can be one of "Getting Started", "Intermediate" and "Advanced" 
- `topics` — an array accepting multiple string values 

![image](https://github.com/user-attachments/assets/4f5727aa-d7ca-4940-8ae5-0c4203c3b1d3)

Example frontmatter:
```yaml
---
title: Custom Scripts
icon: /docs/icons/plugins.svg
stage: Getting Started 
version: ">= 0.18.0"
editions: ["OSS", "EE"]
topics:
  - DevOps
  - Python 
---
```

wrussell1999:
Solved with https://github.com/kestra-io/docs/pull/1699

PR #1699, opened 2024-09-06
added filtration

t-chatoyan:
Solves https://github.com/kestra-io/docs/issues/1694


wrussell1999:
Can you make the titles on the cards smaller @t-chatoyan ?
Screenshot 2024-09-06 at 14 38 48

wrussell1999:
The search doesn't seem to work either @t-chatoyan

content/docs/15.how-to-guides/index.md
```diff
@@ -3,7 +3,8 @@ title: How-to Guides
 icon: /docs/icons/tutorial.svg
 ---
```
wrussell1999, line 4: We need a way to hide all the sub pages in the side menu. It should only show "How-to Guides" at all times with the only way to change guide by going back to the how-to guide home page and using the filters/search

wrussell1999: `hideSidebar: true` doesn't quite solve this as it hides How-to Guides from the menu all together. Whereas when used for the getting started folder, it kept the "Getting Started" title visible but hide all the sub pages unless selected. We want similar to that but also hide sub pages when on them

anna-geller:
+1 to what @wrussell1999 mentioned and additionally the filters don't work well yet
- [ ] missing option to remove the filter for stage (or select "All")
- [ ] topic filter must be multi-select with the option to remove any option anytime, the same as blueprint tags
- [ ] too large title as Will said

image

wrussell1999 (approved):

PR #1704, opened 2024-09-09
fixed css

t-chatoyan:
for the [css](https://nuxt.com/docs/getting-started/styling#the-css-property), we are using the css props on nuxt and it's clearly defined that it will be inline :

I think you need to move to [import ](https://nuxt.com/docs/getting-started/styling#importing-within-components)inside the layout

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -26,6 +26,8 @@ useHead({
 </script>
 
 <style lang="scss">
+@import "./assets/styles/vendor";
+@import "./assets/styles/app";
 .text-gradient {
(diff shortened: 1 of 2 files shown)


tchiotludo:
it didn't work as expected, the vendor are still inline
image

tchiotludo:
the fix was [here](https://github.com/kestra-io/docs/blob/4427c561c15231ec900f10f6f3b64873ab8d95fb/nuxt.config.ts#L264)

PR #1721, opened 2024-09-12
fix(doc): Logs do not reveal secrets

nicoallgood:
Removed examples that are trying to log secrets. Those are de-facto broken since #3018 .

Also, changed 'password' example to 'myPrivateCode' to avoid user confusion (between variable name and value) since a lot of the content is already about passwords. :)

### DIFF
--- a/content/docs/05.concepts/04.secret.md
+++ b/content/docs/05.concepts/04.secret.md
@@ -73,7 +73,7 @@ If you would add the environment variable to the `kestra` container section in a
       SECRET_MYPASSWORD: bXlQcml2YXRlQ29kZQ==
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/05.concepts/04.secret.md
```diff
@@ -85,25 +70,16 @@ If you would add the environment variable to the `kestra` container section in a
...
-
-```yaml
-id: secretTest
-namespace: company.team
-tasks:
-  - id: hello
-    type: io.kestra.plugin.core.log.Log
-    message: "{{ secret('MYPASSWORD') }}"
-```
+This secret can then be used in a flow using the `{{ secret('SECRET_MYPASSWORD') }}` syntax, and it will base64-decoded during flow execution. Make sure to not include the prefix `SECRET_` when calling the `secret('MYPASSWORD')` function, as this prefix is only there in the environment variable definition to prevent Kestra from treating other system variables as secrets (for better performance and increased security).
```
Ben8t, line 76:
```suggestion
This secret can then be used in a flow using the `{{ secret('MYPASSWORD') }}` syntax, and it will base64-decoded during flow execution. Make sure to not include the prefix `SECRET_` when calling the `secret('MYPASSWORD')` function, as this prefix is only there in the environment variable definition to prevent Kestra from treating other system variables as secrets (for better performance and increased security).
```

Ben8t (approved):

PR #1786, opened 2024-09-28
feat(docs): all links are now relative

brian-mulier-p:


wrussell1999:
Rules going forward to add to the README:
- any reference to something else other than docs (plugins, blogs, examples (anything in public)), use absolute paths (eg. /plugins or /blogs/2024...), images are also passed using absolute paths (it's easier like that as working in relative paths while they have different root looks weird)
- any reference to another markdown file should be done with relative link. Provide the full path, including the order numbers, the .md at the end and if it's the index.md still give it in the path. The benefit of that is that we can CTRL-CLICK any markdown link when editing the doc.

wrussell1999 (approved):

Issue #1819, opened 2024-10-07
Use Cases pages break when reloaded.


wrussell1999:
### Describe the issue

Happens when you refresh any of the following pages:
- https://kestra.io/use-cases/data-engineers
- https://kestra.io/use-cases/software-engineers
- https://kestra.io/use-cases/platform-engineers

Error:
> Cannot read properties of undefined (reading 'name')

https://github.com/user-attachments/assets/be8d3fd5-09d7-43bb-a277-296315de3654

They work fine if you go to them from the menu.

### Environment

- Kestra Version:
- Operating System (OS/Docker/Kubernetes):
- Java Version (if you don't run kestra in Docker):

Vivek-1102:
Hey @wrussell1999 I got this please assign this to me .

wrussell1999:
@Vivek-1102 done! let me know if you have any questions!

Vivek-1102:
Yeah sure it seems to be an client side routing error

Vivek-1102:
@wrussell1999 can you please tell me where client-side routing is defined

Vivek-1102:
@wrussell1999 its challenging for me can you please help me so we can walk through this issue

wrussell1999:
@Vivek-1102 you’re gonna want to look in these locations:

- Pages: https://github.com/kestra-io/docs/tree/main/pages/use-cases
- Components used: https://github.com/kestra-io/docs/tree/main/components/use-cases
Let me know if you need anything else

Vivek-1102:
@wrussell1999  Successfully implemented ! raising the PR soon

PR #1828, opened 2024-10-08
Fix issue #1819 Use Cases pages break when reloaded 

Vivek-1102:
https://github.com/user-attachments/assets/17b47252-8412-48d6-90bf-b5d23a965934

Successfully resolved the reloading issue for some use-cases/* pages and also resolved the browser console errors.


anna-geller (approved):
Thanks so much for contributing and fixing this!

Issue #1843, opened 2024-10-10
Missing some Pebble expression filter and functions in the documentation


loicmathieu:
### Description

The documentation is not exhaustive for all supported Pebble expression filter and functions.

For ex, the `endWith` filter is missing, see https://github.com/kestra-io/kestra/issues/5390

The current list of customization of the templating engine can be found here: https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/runners/pebble/Extension.java

shrutimantri:
Missing Filters
[] endsWith
[] json
[] flatten
[] toIon

Missing Functions
[] json
[] kv
[] encrypt
[] decrypt
[] fromIon

PR #1856, opened 2024-10-12
feat(docs):  add encrypt, decrypt, fromIon and kv functions to pebble function doc

shrutimantri:
Closes https://github.com/kestra-io/docs/issues/1843

### DIFF
--- a/content/docs/05.concepts/expression/04.function.md
+++ b/content/docs/05.concepts/expression/04.function.md
@@ -77,6 +77,37 @@ tasks:
 
 And this works no matter the number of levels of EachSequential tasks used.
 
+## decrypt
+
+The `decrypt` function is used to decrypt the values. AES-256 encryption key is supported for this function.
+
+```twig
+{{ decrypt(secret_key, encrypted_value) }}
+```
+
+## encrypt
+
+The `encrypt` function is used to encrypt the values. AES-256 encryption key is supported for this function.
+
+```twig
+{{ encrypt(secret_key, 'Hello World!') }}
+```
+
+## from Ion
+
+The `fromIon` function will convert any Ion stringto an object allowing to access its properties using the dot notation:
+
+```twig
+{{ fromIon('[1, 2, 3]')[0] }}
+{# results in: '1' #}
+
+{{ fromIon('{foo: [666, 1, 2]}').foo[0] }}
+{# results in: '666' #}
+
+{{ fromJson('{name: "John"}').name }}
+{# results in: 'John' #}
+```
+
 ## fromJson
 
 The `fromJson` function will convert any JSON string to an object allowing to access its properties using the dot notation:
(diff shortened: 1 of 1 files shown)


content/docs/05.concepts/expression/04.function.md
```diff
@@ -96,17 +127,28 @@ The `fromJson` function will convert any JSON string to an object allowing to ac
...
 
-## yaml
+The `kv` function will retrieve the value for the given key from the KV store.
 
-The `yaml` function will convert any string to an object allowing to access its properties.
+```twig
+{{ kv('my_key') }}
+```
+
+By default, it will search for the key within the same namespace to which the flow belongs. And in case, the key is not found in the namespace, it will return the value as `null`.
```
wrussell1999, line 138:
```suggestion
By default, it will search for the key within the same namespace as the flow. If the key is not found in the namespace, it will return the value as `null`.
```

content/docs/05.concepts/expression/04.function.md
```diff
@@ -96,17 +127,28 @@ The `fromJson` function will convert any JSON string to an object allowing to ac
...
+
+In OSS, key will be searched in the same namespace as the flow. However, in EE, you can specify the namespace which should be used to retrieve the value for the key. You can provide namespace as a parameter to the `kv` function.
```
wrussell1999, line 140:
```suggestion
In OSS, the key will be searched in the same namespace as the flow. However, in EE, you can specify the namespace which should be used to retrieve the value for the key. You can provide namespace as a parameter to the `kv` function.
```

content/docs/05.concepts/expression/04.function.md
```diff
@@ -568,4 +610,15 @@ The above example will:
...
 
+## yaml
+
+The `yaml` function will convert any string to an object allowing to access its properties.
+
+```twig
+"{{ yaml('foo: [666, 1, 2]').foo[0] }}"
+{# results in: '666' #}
+
+{{ yaml(yaml_object) | jq(...) | yaml }}
```
wrussell1999, line 621:
```suggestion
"{{ yaml(yaml_object) | jq(...) | yaml }}"
```

wrussell1999: Can you actually give this example some dummy data so it works? replace `yaml_object` and the `...` in `jq`.

wrussell1999 (changes requested):

content/docs/05.concepts/expression/04.function.md
wrussell1999: @shrutimantri tagging for visibility

wrussell1999 (approved):

shrutimantri:
> @shrutimantri can you explain your last commit? Looks like you removed the yaml example for a different kv example.

Resolved on a offline discussion.

PR #1860, opened 2024-10-13
fix(examples) - kestra header line in the KESTRA_CONFIGURATION section

yilativ:
Fixed missed `kestra` header line in the `KESTRA_CONFIGURATION` section.


wrussell1999 (approved):
Thanks for contributing!

PR #1886, opened 2024-10-20
feat(docs): debezium how to guide

### LINKED ISSUE #1434: Improve Debezium documentation (how to guide explaining how the plugin works with Realtime Triggers and batch/polling triggers)
### Description

Debezium triggers require a lot of setup.

Often, users expect that they can just point the plugin to their desired DB, and CDC will work out of the box to trigger the flow anytime there is a new row in the DB, like so:

```yaml
id: hello-world
namespace: company.team

tasks:
  - id: hello
    type: io.kestra.plugin.scripts.shell.Commands
    commands:
      - cat "{{trigger.uris}}"

triggers:
  - id: listen-debezium
    type: io.kestra.plugin.debezium.postgres.Trigger
    hostname: host.docker.internal
    port: "5432"
    username: postgres
    password: "kestra"
    database: taxi
    pluginName: PGOUTPUT
    snapshotMode: INITIAL
    format: INLINE
    interval: PT3S
```

We should make it clear in the documentation:
- any prerequisites on the Debezium configuration side
- using the Realtime Trigger to capture each row/change separately
- using the batch/polling trigger to capture multiple rows that changed within a given interval
- explain the difference in how changed data is captured by the plugin e.g. URI of an internal storage file for the polling trigger and raw data for the realtime trigger

shrutimantri:
Closes #1434 

Reference is taken from the http://debezium.io documentation.

### DIFF
--- a/content/docs/15.how-to-guides/debezium.md
+++ b/content/docs/15.how-to-guides/debezium.md
@@ -8,7 +8,7 @@ topics:
 
 How to enable databases to leverage Debezium tasks and triggers.
 
-In order to ensure the change data capture works seamlessly and get the Debezium tasks and triggers running, you need to perform the necessary setup on the database. We will look into the setup steps in detail.
+In order to ensure the change data capture works seamlessly and to get the Debezium tasks and triggers running, you need to perform the necessary setup on the database. Below are instructions on how to set this up.
 
 ## Debezium with MySQL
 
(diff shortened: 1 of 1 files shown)


content/docs/15.how-to-guides/debezium.md
```diff
@@ -0,0 +1,526 @@
+---
+title: Debezium
+icon: /docs/icons/tutorial.svg
+stage: Intermediate
+topics:
+  - Integrations
+---
+
```
wrussell1999, line 8: Needs a short description

content/docs/15.how-to-guides/debezium.md
```diff
@@ -0,0 +1,526 @@
+---
+title: Debezium
```
wrussell1999, line 2: Based on the issue, this is specifically for the triggers. Can we rename the title to reflect this?

wrussell1999: While this guide is generic, we should specify what this enables you to do in Kestra

wrussell1999 (approved):

PR #1905, opened 2024-10-24
Blog Post: Code in any Language

wrussell1999:
Work in progress blog post to go with dedicated video


content/blogs/2024-10-25-code-in-any-language.md
```diff
@@ -0,0 +1,258 @@
...
+description: "Chose the right tool for the job by keeping your business logic and orchestration logic separate."
+date: 2024-10-25T12:00:00
+category: Solutions
+author:
+  name: Will Russell
+  image: "wrussell"
+image: /blogs/2024-10-25-code-in-any-language.jpg
+---
+
+There are many different programming languages, all with different key benefits that make them the right tool for a particular job. That’s why at Kestra we believe it’s important that your business logic is separate from your orchestration logic so you can pick the right tool for each job. Many orchestrators have your business logic and orchestration logic together, for example using Python. This restricts your options if you requirements change in the future.
```
anna-geller, line 12:
```suggestion
There are only two kinds of programming languages: the ones people complain about and the ones nobody uses. Each language has its own pros and cons. That's why at Kestra, we offer you the flexibility to code in any language. This functionality is possible because Kestra separates your business logic from the glue code needed for orchestration. 

In this post, we'll look at how you can simultaneously use multiple popular programming languages to define your tasks and how you can tie them together into an end-to-end workflow using a simple YAML configuration. Let's dive in!
```

I believe you could cut other explanation and the pandas code snippet from the intro, end the intro here and continue the post with perhaps one section per language you want to discuss and one section tying them all together + conclusion, WDYT? or perhaps did you intend this post to be more conceptual without presenting any code/guide?

wrussell1999: Been thinking on this - thinking that I want this blog post to be more around introducing you to the subject, and then making a how-to guide that compares different languages and also covers stuff like inputs and outputs. The goal of this post is really to show that there’s a few ways to write code in Kestra and it works with any language. However, with this in mind, I will tweak a few bits to help send that message across!

PR #1842, opened 2024-10-10
Release 0.20 docs

Skraye:

PR #1907, opened 2024-10-24
docs: 0.20.0 release highlights

anna-geller:


content/docs/06.enterprise/apps.md
```diff
@@ -0,0 +1,273 @@
...
+
+Build custom UIs to interact with Kestra from the outside world.
+
+## Overview
+
+Apps let you use your Kestra workflows as the backend for custom applications while Apps serve as frontend, allowing anyone to interact with your workflows regardless of their technical background. Business users can manually approve workflows that are paused, submit data to automated processes using simple forms, send requests, and view the execution results.
+
+You can build custom applications that trigger new workflow executions, resume paused workflows waiting for approval, or interact with Kestra’s API. In short, Apps allow you and your users to interact with Kestra from the outside world.
+
+---
```
wrussell1999, line 16: FYI - you don't need to add `---` when the next section is a Heading 2 as it adds them too.

anna-geller: I'm aware that this is the design we use on the website atm (we shouldn't rely on this though). I prefer to keep using dividers to keep the "raw markdown" that will work the same way everywhere regardless of how kestra's website renders it. If you feel strongly about it, I can remove those though, np  👍

wrussell1999: Ah smart thinking - let's keep it as!

anna-geller:
moved to https://github.com/kestra-io/docs/pull/1842

PR #1929, opened 2024-11-04
feat(docs): mention host HA config parameters

yuri1969:
Simply listed the kernel parameter required for handling HA environment, such as AWS OpenSearch cluster upgrades.

Feel free to place the note to a more suitable place.


anna-geller (approved):
thx a lot!

PR #2031, opened 2024-12-13
Rework the How-To Guide to allow 3 guides in a single row

### LINKED ISSUE #2016: Rework the How-To Guide to allow 3 guides in a single row instead of just 2
### Issue description

https://www.figma.com/design/g8Di3M1gbajOjOqp0ZrzXO/02_Kestra.io?node-id=14308-14278&t=8d3SMuwgLPhpAX2I-4

image

d3bvstack:
Fixes #2016

Change the class in `components/content/GuidesChildCard.vue` to display 3 guides in a single row instead of 2.

* Modify the class from `col-12 col-md-6` to `col-12 col-md-4` for each guide in the `NuxtLink` component.

### DIFF
--- a/components/content/GuidesChildCard.vue
+++ b/components/content/GuidesChildCard.vue
@@ -34,7 +34,7 @@
                 </div>
             </div>
         </div>
-        <NuxtLink :href="item._path" class="col-12 col-md-6 mb-lg-4 mb-2" v-for="item in navigation" :key="item._path">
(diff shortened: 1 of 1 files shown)


anna-geller (approved):
good idea! let's see how this will look like — if this breaks some display, we'll revert and add a new design with our internal product designer. thanks for contributing again!

PR #2033, opened 2024-12-17
docs(configuration): add environment variables

MarcoSaba:
Describing how to populate env vars instead of config files.

### DIFF
--- a/content/docs/configuration/index.md
+++ b/content/docs/configuration/index.md
@@ -36,6 +36,27 @@ kestra:
   url: "http://localhost:8080/"
 ```
 
+Injecting OS environment variables is also a possible way to configure Kestra. These environment variables take precedence over configuration files that have been loaded.
+
+To populate an environment variable with the right property, replace any special character with `_` (underscore). Below a comprehensive table on how properties can be translated in variables.
+
+| Configuration Value | Resulting Properties                                                                                                               |
+|---------------------|------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


content/docs/configuration/index.md
```diff
@@ -36,6 +36,27 @@ kestra:
   url: "http://localhost:8080/"
 ```
 
+Injecting OS environment variables is also a possible way to configure Kestra. These environment variables take precedence over configuration files that have been loaded.
+
+To populate an environment variable with the right property, replace any special character with `_` (underscore). Below a comprehensive table on how properties can be translated in variables.
```
wrussell1999, line 41:
```suggestion
To populate an environment variable with the right property, replace any special characters with `_` (underscore). Below is a comprehensive table on how properties can be translated in variables.
```

content/docs/configuration/index.md
```diff
@@ -36,6 +36,27 @@ kestra:
...
+
+```yaml
+datasources:
+  postgres:
+    username: kestra
+```
+
+to environment variable:
+
+`DATASOURCES_POSTGRES_USERNAME=kestra`
```
wrussell1999, line 58:
```suggestion
to environment variable: `DATASOURCES_POSTGRES_USERNAME=kestra`
```

wrussell1999 (commented):
Looks good. Couple tweaks suggested

wrussell1999 (approved):

PR #2046, opened 2024-12-20
feat: add docId for in-app docs

anna-geller:
@elevatebart I'm not sure if I got this one right, is this the correct syntax when a docs page can be fetched in multiple UI pages?

```yaml
docId:
  - new
  - flowEditor
```


content/docs/05.concepts/editor.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Editor
+icon: /docs/icons/concepts.svg
+docId:
+  - new
+  - flowEditor
```
elevatebart, line 6: @brian-mulier-p this will not work at all right?

brian-mulier-p: No, it should be the other way around I think, the doc id should be set to the same value for both UI pages, I'm not sure we want to handle multiple ids there

elevatebart: If the same doc works for 2 different places, I do not see a counter argument to do it from a design point of view.

I am not aware of the way we index the pages in Elastic. If the limitation is technical then let's discuss.

anna-geller: The docId needs to be changed in the kestra UI app code

we change it here to just flowEditor

PR #2079, opened 2025-01-14
fix(docs): amend v0.17 renamed plugins

yuri1969:
The `io.kestra.core.tasks.flows.Flow` was aliased to `io.kestra.plugin.core.flow.Subflow`.


anna-geller (approved):
thx Yuri! <3

PR #2093, opened 2025-01-16
(chore): quality edits for getting started docs

aj-emerich:
Went through our Getting Started section of the documentation to refresh and made some quality, consistency, and clarity edits.

### DIFF
--- a/content/docs/01.getting-started/01.quickstart.md
+++ b/content/docs/01.getting-started/01.quickstart.md
@@ -3,7 +3,7 @@ title: Quickstart
 icon: /docs/icons/quickstart.svg
 ---
 
-Start Kestra in a Docker container and create your first flow.
+In this guide, get started with Kestra quickly and create your first flow.
 
 <div class="video-container">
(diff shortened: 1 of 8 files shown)


content/docs/01.getting-started/01.quickstart.md
```diff
@@ -3,7 +3,7 @@ title: Quickstart
 icon: /docs/icons/quickstart.svg
 ---
 
-Start Kestra in a Docker container and create your first flow.
+In this guide, get started with Kestra quickly and create your first flow.
```
wrussell1999, line 6: For context, the first sentence in these pages is what gets displayed in the cards. I tried to shorten them so it didn't overload the card, and if I wanted a better longer line, I'd repeat underneath. This is kinda a mess so maybe this is something we can look into. If you think this is fine, then no problem! Just thought I'd explain some of the thinking here.

Screenshot 2025-01-16 at 13 00 18

aj-emerich: Let's leave it the way it is, I forgot that is what is getting pulled in 👍

PR #2117, opened 2025-01-22
(fix): update Cluster to Instance

aj-emerich:
Cluster has been changed to Instance and needed to be updated in a couple of places 👍


aj-emerich:
Do not merge yet: there are some screenshots that require an update pending another issue 👍

PR #2127, opened 2025-01-22
Revert "feat: json-schema-based plugin doc with redesign"

brian-mulier-p:
Reverts kestra-io/docs#2125


brian-mulier-p:
@elevatebart meanwhile we investigate the bundle size issue I revert my change...

PR #2159, opened 2025-02-02
feat(docs): add v0.17 JSON-related migration guide

yuri1969:
The JSON Object serialization change got a strong impact on pre-v0.17 expression logic. Added a migration guide:

* mentioning the necessity to adapt the affected expressions
* listting a few examples


anna-geller (approved):
Thanks so much, you're totally right, we should have added it here. So far, we only focused on https://kestra.io/docs/how-to-guides/null-values

PR #2254, opened 2025-02-19
fix(docs): list `runIf` and `allowWarning` props

yuri1969:
Recently added task properties were missing from the list.


Skraye (approved):
LGTM

PR #2266, opened 2025-02-22
test: close init request before init is over to see if this is the problem

elevatebart:


elevatebart:
turns out it is. the problem

PR #2338, opened 2025-03-12
fix(docs): combine namespace mgt and centralized page

wrussell1999:
Combining two very similar pages after realising they don't need to be separate.


content/docs/06.enterprise/02.governance/07.namespace-management.md
```diff
@@ -60,7 +147,7 @@ The video below shows how you can create a namespace from the Kestra UI. After c
 
 For detailed instructions, refer to [centralized task configuration](../02.governance/08.centralized-task-configuration.md).
```
aj-emerich, line 148:
```suggestion
```

aj-emerich (approved):

content/docs/06.enterprise/02.governance/07.namespace-management.md
aj-emerich: We could reroute this link, but I think it is not adding anything and can just be removed.

PR #2369, opened 2025-03-25
fix: typo in architecture doc

fdelbrayelle:


aj-emerich:
@fdelbrayelle Thank you for spotting that and making the fix, LGTM 👍

PR #2380, opened 2025-03-26
fix(docs): add missing typescript install step

fdelbrayelle:
Hi :wave: 

I followed the doc to run locally the docs on a new machine and I had to install typescript globally so I added this as a mandatory step before to run `npm install`.

I also had to `npm install d3` because of these errors:

image

### DIFF
--- a/content/docs/15.how-to-guides/local-flow-sync.md
+++ b/content/docs/15.how-to-guides/local-flow-sync.md
@@ -42,13 +42,13 @@ File created locally should use `<namespace>.<flow_id>.yml` or `<namespace>_<flo
 Flow created inside the UI will be created at the root of the first path supplied in the configuration.
 
 
-::alert{type="note"}
+::alert{type="info"}
 If you are using the docker-compose installation, you will need to mount a volume so Kestra container can access your local folder.
 
 ```yaml
     volumes:
       # ... other volumes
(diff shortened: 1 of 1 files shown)


aj-emerich:
@elevatebart Would you mind giving a quick review? The md looks good, but I don't know about the package-lock.

elevatebart:
Hello @fdelbrayelle,

Thank you for taking this one on. 
We really enjoy seeing people involved in the docs.

Can you say a bit more about the environment where you saw these errors?

OS + version
Node version
NPM version (or other package manager + version)

Notes: 
- I have never installed TypeScript globally. 
- `d3` is a dependency of mermaid that we use for some of the schematics. 
It should be installed without you having to pin down a version since we only use it transitively.

Thank you again

fdelbrayelle:
Hi @elevatebart :wave: 

Here are my infos:

```
➜  docs git:(main) cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.2 LTS"
➜  docs git:(main) node -v             
v20.14.0
➜  docs git:(main) npm -v              
10.7.0
```

Thank you for your help!

Issue #2393, opened 2025-04-02
Add a Docs-Contributor Guide page explaining all frontmatter and Vue properties


anna-geller:
If somebody wants to contribute to docs/blog, they will not understand how without having to explain it thoroughly. We use so many custom implementations - some ideas that could be explained in a "Docs-Contributor Guide" (name TBD):
- ChildCard
- docId to enable contextual in-app docs
- editions + version in the frontmatter
- all types of alerts e.g. `::alert{type="warning"}`
- expand-collapse e.g. `::collapse`
- custom divs e.g. `::div`
- how icon property is used and where icons are stored
- the `release` frontmatter property used only by migration guide and how migration guide can be added
- sizing of video container
- how we store and reference **images** -- we could show it based on an example for a blog post (one main image referenced in the frontmatter and all other images in a dedicated folder for the blog post)
- usage of the index.md along with folders to create nesting if needed
- (optional) explain custom setup to add another "Kestra vs." easily
- bonus: how to build docs locally (+ perhaps how to fork the repo, create a PR and view deploy preview from CloudFlare)
- how we add a branch for the new release and merge it only once the release is done
- how we use `DO_NOT_MERGE` label to avoid merging docs for the next release
- how blueprints and plugn documentation are managed elsewhere outside of the main docs repo and how you can contribute to plugin docs + blueprints
- how to contribute new blueprints + how to report bugs in existing ones
- how to contribute new plugin examples + how to report bugs in existing ones
- how to get docs for your specific version — explain in-app docs, how they work (linking between UI pages and docs pages via `docId` + show how the in-app docs can be opened from the sidebar)
- how to report missing docs vs bug on the website (ideally also explaining the labels)
- likely other things I forgot.

In sth like a "Docs-Contributor Guide", we could document it for both our internal team and for external docs contributors, explaining all Vue and frontmatter properties like docId, editions, version etc.

aj-emerich:
I'm closing for now because the bulk of this is added, and I wanted to get it out. We can add the smaller pieces as future amendments. The list can extend on and on 😄

PR #2437, opened 2025-04-16
doc(local-flow): fix info block

wrussell1999:
Change from note to info. Also make the paths clearer by using different names.

Currently unreadable:
Screenshot 2025-04-16 at 11 19 17


aj-emerich:
Is this a global issue for admonition (i.e., do we have other instances of `::note`)?

aj-emerich (approved):
LGTM, I answered my own question and found that `type="note"` wasn't used anywhere else.

aj-emerich:
> > LGTM, I answered my own question and found that `type="note"` wasn't used anywhere else.
> 
> Yeah I've never seen notes used. Probably worth adding a note somewhere of what the types are. I know we have `info`, `warn` and `error` I think

I will address it solving this issue https://github.com/kestra-io/docs/issues/2393

PR #2439, opened 2025-04-16
docs(inputs): mandatory param is id not name

fdelbrayelle:


aj-emerich (approved):
LGTM. Good catch, thank you!

PR #2454, opened 2025-04-18
 fix: the image size problem on kestra.io

elevatebart:
Since we want to use cloudflare images provider (for money reasons) we need to have it only done in prod.
I have tested the generated urls in theory in production and they work.

It might be a bit of a pickle for a while, but the perf gain is worthe the work.

Total page weight for blog posts might be divided by 1000.


brian-mulier-p (approved):
LGTM

PR #2517, opened 2025-05-02
fix: repair windows dev setup

elevatebart:
The main problem is 2 fold:
- There is no error management in our MDC setup (meh)
- in the `remark-code-import` plugin, we use `node:path`. When this is bundled by vite for the browser, it assumes node premises (it has to assume something). Unfortunately, this means `path.isAbsolute()` is falsy for absolute windows paths. I removed the protection since the config is already done.


Piyush-r-bhaskar:
Seems working for me 🚀

Thank You so much @elevatebart 😊

Piyush-r-bhaskar (approved):

Issue #2489, opened 2025-04-28
Doc Alert info boxes are impacted by Plugin doc change


aj-emerich:
### Describe the issue

The information admonition components in the docs are likely impacted by a recent change to the plugin boxes (https://github.com/kestra-io/docs/pull/2446). They now appear with double icons:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/75b0ed49-d179-41e9-a7e3-a3f5d38f5f8e" />

For example, see the [Worker Group](https://kestra.io/docs/enterprise/scalability/worker-group) page. This does not affect other admonition types (e.g., Warning, etc.).

Issue #2410, opened 2025-04-08
Beta flag is unreadable on Plugin Docs


aj-emerich:
### Describe the issue

https://kestra.io/plugins/plugin-graalvm/js/io.kestra.plugin.graalvm.js.filetransform
https://kestra.io/plugins/plugin-graalvm/js/io.kestra.plugin.graalvm.js.eval

Also both Ruby and Python tasks.

Same on both Brave and Chrome browser.

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/8eba761a-365b-4ce6-bd5a-a98af448b08d" />

starlightknown:
Hello, would love to take this up, once my issue of page not loading gets solved :")

aj-emerich:
@starlightknown Hi again, could you access the page after navigating to it from a different plugin, as I mentioned on [Slack](https://kestra-io.slack.com/archives/C05S0242T4G/p1744812375860259)?

starlightknown:
> [@starlightknown](https://github.com/starlightknown) Hi again, could you access the page after navigating to it from a different plugin, as I mentioned on [Slack](https://kestra-io.slack.com/archives/C05S0242T4G/p1744812375860259)?

Hi no :/

Nico-Kestra:
Hello, we can use this exact `alert-info` design.

<img width="856" alt="Image" src="https://github.com/user-attachments/assets/2bdf6e2d-9ba0-43a2-a012-58cc3cf6b37b" />

starlightknown:
Also I'm not sure how is this working but not

https://github.com/kestra-io/docs/issues/2442#issue-2999917246

I inspected and the alert is coming inside <p> tags

PR #2519, opened 2025-05-05
chore(core): fixes the double icon appearing in alert box.

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/docs/issues/2489

### DIFF
--- a/assets/styles/app.scss
+++ b/assets/styles/app.scss
@@ -91,45 +91,6 @@ table.table {
     > *:last-child {
         margin-bottom: 0;
     }
-
-    &.alert-info {
-        display: flex;
-        gap: 12px;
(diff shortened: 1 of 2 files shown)


MilosPaunovic (approved):
Code-wise, this is good.

@Piyush-r-bhaskar could you just double check if this issue https://github.com/kestra-io/docs/issues/2410 which actually produced the one you're trying to amend could still be considered as solved.

Piyush-r-bhaskar:
Yeah @MilosPaunovic ...issue was we were having issue on visibility of message like `This plugin is currently in beta....`   which is coming from `ui-libs`

MilosPaunovic:
Meaning that one is showing properly, as well? If, so, I'm okay with you merging this PR.

PR #2522, opened 2025-05-05
update pricing list

MartinRst:


components/price/Table.vue
```diff
@@ -433,7 +439,7 @@
         {
           id: 7,
           title: "Custom Plugins",
-          openSourceText: 'Priority in plugin roadmap',
+          openSourceText: false,
```
anna-geller, line 442:
```suggestion
          openSourceText: true,
```
this is not true, OSS users can create custom plugins, no limits here

PR #2526, opened 2025-05-05
feat: add scalar API DOC instead of rapidoc

brian-mulier-p:


anna-geller:
@brian-mulier-p can check with the latest API version where we fixed a bunch of issues, might be good to merged in this cycle

anna-geller:
@brian-mulier-p @tchiotludo what are the next steps here? can we merge by the end of the cooldown maybe?

PR #2537, opened 2025-05-06
fix(plugins): fix plugin filtering logic

### LINKED ISSUE #2535: Tag filters are not working on the plugins page
### Describe the issue

https://github.com/user-attachments/assets/f847113e-d949-4184-a212-0aec2a3bb6fe

elevatebart:
closes #2535

### DIFF
--- a/components/plugins/PluginsLists.vue
+++ b/components/plugins/PluginsLists.vue
@@ -72,7 +72,7 @@
         if (!props.plugins) {
             return new Set();
         }
-        
+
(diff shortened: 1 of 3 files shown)


MilosPaunovic (changes requested):
When filter tag is selected, it does show proper cards, but the pagination stays unchanged, and clicking through the pages seems to have no effect.

Maybe cover that one also as part of this PR?

MilosPaunovic (approved):

Issue #2497, opened 2025-04-29
Document difference between a Flow's Triggers tab and Admin's Triggers tab


aj-emerich:
### Describe what's missing in our docs

In the app, a flow's **Triggers** tab shows Flow type and Scheduler-related triggers. In the Administration Triggers tab, Flow triggers are not shown; only triggers related to the schedule are displayed. 

Add a line in the documentation to clarify this specification and explain why this is the case.

Per internal [Slack thread](https://kestra-io.slack.com/archives/C052XEFFC90/p1745937374435719)

Admin Triggers:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/b4768c38-90ea-4cd2-b573-b11c5793e36b" />

Flow Triggers:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/b27c9aff-8ae8-45d2-943c-6053e1536af2" />

PR #2538, opened 2025-05-06
docs(triggers): note about webhook triggers not displaying in the UI or API

wrussell1999:
Check that this is correct, but noticed we don't talk about the changes made [here](https://github.com/kestra-io/docs/issues/2497) in the triggers page.

We had a user who was trying to fetch triggers from the API and didn't get their webhook one which I assume is related to this. Also doesn't display in UI.


aj-emerich (approved):
Good catch! Let's add it

PR #2548, opened 2025-05-12
fix: count of plugins

elevatebart:


components/plugins/PluginsLists.vue
```diff
@@ -75,10 +75,18 @@
     const route = useRoute();
     const router = useRouter();
 
-    const totalPlugins = computed(() => {
-        const pluginsList = props.plugins.map(p => `${p.group}-${p.title}-${p.description}`);
-        return new Set(pluginsList).size;
-    });
+    const GROUPS = ["tasks", "triggers", "conditions", "taskRunners"];
```
MilosPaunovic, line 78: Are we certain that these 4 are the only options for groups? If so, then we're good.

MilosPaunovic (approved):

PR #2574, opened 2025-05-15
feat(docs): improve property default example

yuri1969:
Add `@Builder.Default` to the defaulted property since the guide explicitly mentions `@SuperBuilder`. Also nuke whitespace prefixes.


anna-geller (approved):
Thanks a lot!

PR #2576, opened 2025-05-15
feat(docs): add validation example

yuri1969:
Property validation must be performed at its inner type.


anna-geller (approved):
Thanks a lot!

PR #2598, opened 2025-05-22
docs(keycloak): fix guide

wrussell1999:
- [x] correct issuer should be https://{{keyCloakServer}}/realms/{{yourRealm}}  without /auth in it.
- [ ] The screenshot contains localhost:8081 which is not the default port as Kestra uses 8080 by default
- [x] no syntax highlighting on bottom code block


wrussell1999:
Duplicate of https://github.com/kestra-io/docs/pull/2598

PR #2608, opened 2025-05-23
feat(docs): mention reserverd plugin property name

yuri1969:
Let's make `version` property-related compilation failures clear.


aj-emerich (approved):
Thank you Yuri!

PR #2757, opened 2025-07-01
Update 13.timeout.md with runnable task limitation

sjoerdsmink:
As long as https://github.com/kestra-io/kestra/issues/6739 is open, it might be good to mention the limitation that timeout doesn't work for flowable tasks.


aj-emerich:
@sjoerdsmink That makes sense, and I will make a note to update when the issue gets closed.

PR #2781, opened 2025-07-04
docs(json): json guide

wrussell1999:
Guide on using JSONs via expressions.


content/docs/15.how-to-guides/json.md
```diff
@@ -0,0 +1,130 @@
...
+
+However, if the body is large, we may only want to access a certain part of it. To do this, `jq` is required as the expression returns a string. Using `jq`, the JSON can be parsed and accessed:
+
+```yaml
+{{ outputs.request.body | jq('.title') | first }}
+```
+
+This will access the key `title` from the JSON and return that inside of an array. The additional function `first` is added to return it as a string, removing it from the array.
+
+We can put that into the example
```
aj-emerich:
```suggestion
We can put that into the example:
```

aj-emerich (approved):
LGTM!

PR #2965, opened 2025-08-22
Create notion-webhook.md

aj-emerich:
Part of https://github.com/kestra-io/kestra/issues/8557

### DIFF
--- a/content/docs/15.how-to-guides/notion-webhook.md
+++ b/content/docs/15.how-to-guides/notion-webhook.md
@@ -192,6 +192,10 @@ tasks:
         messageText: "New task assigned!"
 ```
 
+::alert{type="info"}
+Keep in mind that the above examples are additional tasks to add to the flow and not standalone flows. You need to add `id` and `namespace` properties to execute them standalone.
+::
+
 ## Security considerations
 
 - Use strong, randomly generated webhook keys


content/docs/15.how-to-guides/notion-webhook.md
```diff
@@ -0,0 +1,208 @@
...
+3. Select **"Add connections"**
+4. Find and select your integration
+5. Click **"Confirm"** to grant access
+
+## Set up secrets in Kestra
+
+Store your sensitive credentials as [secrets](../05.concepts/04.secret.md):
+
+1. Navigate to your namespace in the Kestra UI
+2. Go to the **Secrets** tab
```
wrussell1999: This assumes using EE. Do we want to also refer to KV Store?

content/docs/15.how-to-guides/notion-webhook.md
```diff
@@ -0,0 +1,208 @@
...
+        type: io.kestra.plugin.notifications.slack.SlackIncomingWebhook
+        url: "{{ secret('PRODUCT_SLACK_WEBHOOK_URL') }}"
+        messageText: "New product task assigned!"
+    else:
+      - id: general_notification
+        type: io.kestra.plugin.notifications.slack.SlackIncomingWebhook
+        url: "{{ secret('GENERAL_SLACK_WEBHOOK_URL') }}"
+        messageText: "New task assigned!"
+```
+
```
wrussell1999, line 194: Thoughts on adding an ID and namespace to make them full flows? I know that means they’re missing the triggers but we have had it in the past where people copy and paste this stuff directly into the editor and don’t understand why it doesn’t work straight away. 

Saying that, don’t remove tasks as that would make it more confusing.

wrussell1999 (approved):

content/docs/15.how-to-guides/notion-webhook.md
aj-emerich: Ah I added it in other places but not here 👍 Good catch

content/docs/15.how-to-guides/notion-webhook.md
aj-emerich: I prefer to leave them as tasks, and I'll just add a note that they are not full flows. The platform editor will say they are missing those properties anyway.

PR #3017, opened 2025-09-09
chore: update ui-libs with recent mdc

elevatebart:
This updates MDC inside the website.
I had some trouble with i18n management compatibility but everything should be peachy now.

Please review.


brian-mulier-p (approved):
LGTM

PR #3021, opened 2025-09-09
docs(docker-compose): link back to tags

aj-emerich:


content/docs/02.installation/03.docker-compose.md
```diff
@@ -90,6 +90,10 @@ Next, update the `kestra` service in the `docker-compose.yml` file to mount this
         condition: service_started
 ```
 
+::alert{type="info"}
+Check out all of our avaiable [Docker image tags](02.docker.md#docker-image-tags) to see which one is best for your use case.
```
wrussell1999: Does this need a ./ in front of it?

aj-emerich: It still works. However, it jumps to a weird spot on the page where the section is at the bottom. I am not sure if that is caused by this or something else.

PR #3082, opened 2025-09-17
refactor: generalize top menu to avoid nuxt dependencies

elevatebart:


package-lock.json
MilosPaunovic, line 1: Is this expected change?

elevatebart: Indeed

PR #3202, opened 2025-10-14
docs(workers): add redirect from deleted admin page

aj-emerich:


nuxt.config.ts
```diff
@@ -1,5 +1,6 @@
 import * as sass from "sass";
 import { CollectionNames } from "./content.config.names";
+import redirect from "./server/middleware/redirect";
```
wrussell1999: Assumed this would've already been here?

aj-emerich: Back to what it should be. I think that was an accident on my end from autocomplete.

Issue #3220, opened 2025-10-16
Plugin Docs Bugs & Enhancement Proposal for the Plugins website


wrussell1999:
### Describe the issue

- [ ] When you close the expanded definition, the page jumps back to the top. It should not scroll.
- [ ] Sometimes it takes two clicks to expand the definition. It should take one.
- [ ] Expanding properties in a definition causes that property to go off the top of the screen when it auto scrolls. It should not scroll.
- [ ] Sometimes the property expands to nothing when clicked.
- [ ] When closing the expanded property, it scrolls to the top. It should not auto scroll.

On this page specifically but will probably exist on other docs pages too: https://kestra.io/plugins/plugin-script-shell/io.kestra.plugin.scripts.shell.commands

https://github.com/user-attachments/assets/a30dd3c4-0969-4acf-a83d-20db0804f3c4

https://github.com/user-attachments/assets/90e82a3b-070b-4ee7-8f30-889b2fa1c6a9

https://github.com/user-attachments/assets/8b70396a-b072-4280-9771-080fc512d8df

https://github.com/user-attachments/assets/d35a314d-b5b4-4481-895f-232341373cb4

Vishnu-Varthan1:
I am working on this issue and will submit a PR soon.

yuvrajkarna2717:
Hi @wrussell1999 and @Piyush-r-bhaskar, Is this still open ? Would love to contribute, please assign this issue to me. Thanks.

MilosPaunovic:
Hi @yuvrajkarna2717,  

You're already assigned to another issue in our repository, and we usually assign only one at a time to new contributors. Let's make sure to wrap that one up before jumping onto another.  🚀

Raviikumar001:
@MilosPaunovic working on the issue, pls do give some time, as suggested i'm working on resolving the second issue,
i found its in ui-libs part, working on a solution.

PR #3228, opened 2025-10-18
Fix expand/collapse scroll bug for property cards (#3220)

vishnu-varthan1:
Closes https://github.com/kestra-io/docs/issues/3220.

 Fixes the unwanted scroll-to-top and double-click bug when expanding or collapsing property cards in the plugin docs.
- Uses a simple boolean toggle (`isExpanded`) for expand/collapse.
- Removes any scrollIntoView or scrollTo logic from property card toggles.
- Ensures Table of Contents navigation still scrolls correctly.
Testing
- Single-click to expand/collapse each property card works as expected, without scrolling the page.
- Table of Contents navigation still scrolls to sections correctly.


MilosPaunovic:
Just a small note for the future, if you could use **GitHub Keywords** when opening PRs to connect it to the proper issue (to automate the entire process a bit more), it would be super helpful to us. You can read a bit more about it [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) (I've added closing keyword for this particular issue).

components/content/ConfigPropertyCard.vue
```diff
@@ -33,11 +42,24 @@ export default defineComponent({
...
+
+        function toggleProperty() {
+            isExpanded.value = !isExpanded.value;
+        }
+
+        return {
+            isExpanded,
+            toggleProperty
+        };
+    }
```
Piyush-r-bhaskar, line 58: why `setup` ?  maybe it's easier to move to `<script setup lang="ts">` if not can you handle with methods ?

Piyush-r-bhaskar:
Closing as we have no response and only changes here would not check all the points from the issue.

PR #3236, opened 2025-10-20
docs: fix missing {tenant} in Webhook trigger URL 

D-Prasanth-Kumar:
fixes #11939


aj-emerich (commented):
Hello @D-Prasanth-Kumar , the original issue is referring to the in-app documentation for the Webhook trigger, which is actually handled [here](https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java). That's where you will want to add the `{tenant}` placeholder like in this PR. In these examples, `main` is the open-source tenant, so no need to replace.

PR #3238, opened 2025-10-20
Add clear comments for Europe detection and consent logic in cookie consent plugin

swagatika60:
### What's new ###
->This PR makes the cookie consent plugin smarter and easier to understand.

->Added inline comments that explain how Europe detection works.

->Made the logic for showing GDPR consent modal crystal clear.

->Keeps analytics & marketing flows safe and unchanged.

### Why This Matters ###

Previously, the code had great functionality but lacked explanation. Now, anyone jumping into the project can instantly understand how consent is handled without guessing.

GDPR behavior is transparent
Developers can confidently maintain or extend analytics logic

###Check Out the Improvement ###

```javascript
// Check if the user is in Europe
const isEurope = Intl.DateTimeFormat().resolvedOptions().timeZone.indexOf("Europe") === 0;

/*
If user is in Europe, show cookie consent modal
Otherwise, enable analytics and marketing automatically
*/
if (!isEurope) {
    enabledAnalytics();
    enabledMarketing();
}
``` 

### Tested & Verified ###

-.Code behavior remains unchanged.

->Comments are informative and improve maintainability.

->Works as expected for users inside and outside Europe.

### DIFF
--- a/plugins/cookieconsent.client.ts
+++ b/plugins/cookieconsent.client.ts
@@ -4,8 +4,19 @@ import axios from "axios";
 import identify from "../utils/identify";
 
 export default defineNuxtPlugin(nuxtApp => {
-    const isEurope = Intl.DateTimeFormat().resolvedOptions().timeZone.indexOf("Europe") === 0;
-    const runtimeConfig = useRuntimeConfig()
+    //const isEurope = Intl.DateTimeFormat().resolvedOptions().timeZone.indexOf("Europe") === 0;
+    let isEurope = false;
+
+try {
+  const res = await fetch('https://ipapi.co/json/');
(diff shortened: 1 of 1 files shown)


elevatebart:
Thank you @swagatika60 for your interest in this. 
UI improvements are always a good idea.

I understand that you would like to detect through IP the location of the current user to present or not the cookie consent banner. Doing this through an API call might be lengthy and brittle. 
It also makes the code harder to read

Since I am not competent to know if this is relevant, I would prefer that you create an issue first. 
Write the issue you are trying to fix. Marketing & Legal can discuss it prior to any implementation.

 Thank you in advance

PR #3330, opened 2025-11-10
docs(quickstart): clean up unnecessary sections

wrussell1999:
I reviewed our quick start and determined that there was too much unnecessary information in it.

- There was information about creating a user in your config, but the quickstart instructions don't include a config, instead just a "run this command and spin up Kestra" so this is unnecessary. Instead, focus on the UI instructions. Nothing that happens on this instance is permanent.
- Add the extra macOS command into a collapse section. Make it more focused. Also it works fine for me on macOS with the original command.

We should look into a way we can track the completion of this quickstart to confirm that people aren't giving up with it

### DIFF
--- a/content/docs/01.getting-started/01.quickstart.md
+++ b/content/docs/01.getting-started/01.quickstart.md
@@ -27,7 +27,8 @@ docker run --pull=always --rm -it -p 8080:8080 --user=root -v /var/run/docker.so
 ```
 
 
-:::collapse{title="If you're on macOS, you may need to add `-e JAVA_OPTS="-XX:UseSVE=0"`"}
+:::collapse{title="macOS troubleshooting"}
+If you're on macOS, you may need to add `-e JAVA_OPTS="-XX:UseSVE=0"`
 
 ```bash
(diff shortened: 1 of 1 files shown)


content/docs/01.getting-started/01.quickstart.md
```diff
@@ -26,11 +26,13 @@ Once Docker is running, start Kestra with a single command. (*If you are using W
 docker run --pull=always --rm -it -p 8080:8080 --user=root -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp kestra/kestra:latest server local
 ```
 
-If you're on macOS, you may need to add `-e JAVA_OPTS="-XX:UseSVE=0"`:
+
+:::collapse{title="If you're on macOS, you may need to add `-e JAVA_OPTS="-XX:UseSVE=0"`"}
```
wrussell1999:
```suggestion
:::collapse{title="macOS troubleshooting"}
If you're on macOS, you may need to add `-e JAVA_OPTS="-XX:UseSVE=0"`
```

PR #3349, opened 2025-11-13
docs(docker-compose): update standalone pg example

aj-emerich:


content/docs/02.installation/03.docker-compose.md
```diff
@@ -116,9 +116,9 @@ networks:
 
 services:
   postgres:
-    image: postgres
+    image: postgres:18
     volumes:
-      - postgres-data:/var/lib/postgresql/data
+      - postgres-data:/var/lib/postgresql/18/data
```
wrussell1999: Think it's suppose to be docker now?

```suggestion
      - postgres-data:/var/lib/postgresql/18/docker
```

Issue #3353, opened 2025-11-14
Add documentation for deploying multiple operators


loicmathieu:
### Describe what's missing in our docs

Enhance the Kubernates Operator guide with information on how to deploy it multiple times inside a single Kubernetes cluster to manage multiple Kestra instances
See https://kestra.io/docs/version-control-cicd/cicd/kubernetes-operator

---

The Kestra operator is bound to a Kestra instance.
So if you have multiple Kestra instances in a single Kubernetes cluster, you need to deploy multiple instances of the operator.

For that to work, you then need to configure each instance of the operator with the list of Kubernetes namespace to watch for resources using either the config property `quarkus.operator-sdk.namespaces` or the enviroment variable `QUARKUS_OPERATOR_SDK_NAMESPACES`.

PR #3385, opened 2025-11-24
docs(k8s-operator): add instruction for multiple operators

aj-emerich:
Closes https://github.com/kestra-io/docs/issues/3353

### DIFF
--- a/content/docs/version-control-cicd/cicd/07.kubernetes-operator.md
+++ b/content/docs/version-control-cicd/cicd/07.kubernetes-operator.md
@@ -70,7 +70,7 @@ kestra-standalone-677474499f-4r5ft   1/1     Running   2 (5h10m ago)   2d23h
 
 ### Managing multiple operators in one cluster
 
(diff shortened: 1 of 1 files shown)


content/docs/version-control-cicd/cicd/07.kubernetes-operator.md
```diff
@@ -68,6 +68,29 @@ kestra-postgresql-0                  1/1     Running   1 (2d23h ago)   3d
 kestra-standalone-677474499f-4r5ft   1/1     Running   2 (5h10m ago)   2d23h
 ```
 
+### Managing multiple operators in one cluster
+
+Each operator instance manages a single Kestra instance. If you run multiple Kestra deployments in the same Kubernetes cluster, deploy one operator per Kestra instance and scope each operator to the namespaces that contain that instance’s CRDs.
```
loicmathieu:
```suggestion
Each operator instance manages a single Kestra instance. If you run multiple Kestra deployments in the same Kubernetes cluster, deploy one operator per Kestra instance and scope each operator to the namespaces that will contain that instance’s custom resources.
```

The CRD are the definition of the custom resources

PR #3472, opened 2025-12-11
docs(requirements): update postgres version

aj-emerich:


content/docs/09.administrator-guide/00.requirements.md
```diff
@@ -25,7 +25,7 @@ Kestra Enterprise Edition (EE) provides two options:
 
 | Kestra Edition    | Database                 | Required version          | Note           |
 |-------------------|--------------------------|---------------------------|----------------|
-| OSS / Enterprise    | **PostgreSQL**           | >=14 && <= 16.3             | Default `latest` |
+| OSS / Enterprise    | **PostgreSQL**           | >=14 && <= 18.0             | Default `latest` |
```
wrussell1999: Do we even need the less than? 18 is latest so as long as you’re running above 14, you should be good?

```suggestion
| OSS / Enterprise    | **PostgreSQL**           | >=14             | Default `latest` |
```

aj-emerich: The minimum requirement seems sufficient to me too, but I'm not sure if we can guarantee 18.1...so that's why it was like that?

aj-emerich: It's better for us to not have to maintain an upper limit for no reason

wrussell1999: I think it’s probably safe to assume that the latest version of Postgres should work. And the team will make an effort to make sure it’s supported when these things are released.

aj-emerich: I agree, let's take this route

PR #3478, opened 2025-12-12
refactor(core): use blueprint count from endpoint

Piyush-r-bhaskar:


server/api/blueprints/related-blueprints.ts
```diff
@@ -0,0 +1,71 @@
+import { slugify, buildPluginMappings } from "@kestra-io/ui-libs"
```
Piyush-r-bhaskar, line 1: this file will be refactor by using new endpoint as well once available. It's temporary.

PR #3483, opened 2025-12-12
docs(editions): remove editions when available in all

wrussell1999:
As per Amara's suggestion, removing the edition label when it's supported in all editions.


aj-emerich:
Was on my list! Nice! There's a sub issue to this, where I think we should remove all the "As of Kestra 0.20" etc. Another topic for another PR.

aj-emerich (approved):

PR #3522, opened 2025-12-17
docs(variables): fix FAQ

aj-emerich:
So Variables can be used in Triggers. Inputs cannot be used in a Trigger without a default value. A variable can be used in a trigger to set an input.

Hard to follow Slack thread: https://kestra-io.slack.com/archives/C04HTFM3VL6/p1765962037837089

### DIFF
--- a/content/docs/04.workflow-components/04.variables.md
+++ b/content/docs/04.workflow-components/04.variables.md
@@ -144,7 +144,7 @@ To ensure that a block of code won't be parsed by Pebble, you can use the `{% ra
 
 [Inputs](./05.inputs.md) are resolved first, before the execution starts. If a flow has an invalid input value, the execution will not be created.
 
-Therefore, you can use inputs within variables, but you cannot use variables or Pebble expressions in most contexts (Check out [Dynamic Inputs](../04.workflow-components/05.inputs.md#dynamic-inputs) for more information) within inputs.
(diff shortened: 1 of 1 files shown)


content/docs/04.workflow-components/04.variables.md
```diff
@@ -144,11 +144,11 @@ To ensure that a block of code won't be parsed by Pebble, you can use the `{% ra
 
 [Inputs](./05.inputs.md) are resolved first, before the execution starts. If a flow has an invalid input value, the execution will not be created.
 
-Therefore, you can use inputs within variables, but you cannot use variables or Pebble expressions within inputs.
+Therefore, you can use inputs within variables, but you cannot use variables or Pebble expressions in most contexts (Check out [Dynamic Inputs](../04.workflow-components/05.inputs.md#dynamic-inputs) for more information) within inputs.
```
wrussell1999: I think this link also works but will need to check. Both in the same directory. Also will this break with the getting started reshuffle?
```suggestion
Therefore, you can use inputs within variables, but you cannot use variables or Pebble expressions in most contexts (Check out [Dynamic Inputs](./05.inputs.md#dynamic-inputs) for more information) within inputs.
```

aj-emerich: That's the next step. I need to check links within the set. I don't think too many things link back to getting started.

wrussell1999 (approved):

PR #3563, opened 2026-01-05
docs(redirects): fixes for 404s in enterprise docs/pages

akeller:
Came up on a Google Search Console report, but this only includes fixes for the 404s from the enterprise docs and website pages. 

Also, I added two comments, one for "releases" when I couldn't find an enterprise specific page to redirect to and another for "system flows" which looks like a topic that's in the general concepts area of the docs. Please confirm these are correct!

I think these will 301 by default, but we should manually validate the fix with GSC.


wrussell1999 (approved):

nuxt.config.ts
```diff
@@ -320,22 +320,28 @@ export default defineNuxtConfig({
         '/docs/concepts/task-runners': {redirect: '/docs/task-runners'},
         '/docs/enterprise/enterprise-edition': {redirect: '/docs/enterprise/overview/enterprise-edition'},
         '/docs/enterprise/setup': {redirect: '/docs/enterprise/overview/setup'},
-        '/docs/enterprise/releases': {redirect: '/docs/enterprise/overview/releases'},
+        //'/docs/enterprise/releases': {redirect: '/docs/enterprise/overview/releases'}, 404s
```
akeller, line 323: @aj-emerich we seemed to have a separate enterprise releases page at one point. Should this just redirect to https://kestra.io/docs/releases? Or is there some place that has enterprise specific info?

nuxt.config.ts
```diff
@@ -344,8 +350,17 @@ export default defineNuxtConfig({
...
+        '/enterprise/instance/dashboard': {redirect: '/docs/enterprise/instance/dashboard'},
+        '/enterprise/instance': {redirect: '/docs/enterprise/instance'},
         '/docs/enterprise/maintenance-mode': {redirect: '/docs/enterprise/instance/maintenance-mode'},
         '/docs/enterprise/governance/centralized-task-configuration': {redirect: '/docs/enterprise/governance/namespace-management'},
+        '/enterprise/auth-rbac-user-management/api-tokens': {redirect: '/docs/enterprise/auth/api-tokens'},
+        '/docs/enterprise/auth-rbac-user-management/api-tokens': {redirect: '/docs/enterprise/auth/api-tokens'},
+        '/docs/enterprise/auth-rbac-user-management/service-accounts': {redirect: '/docs/enterprise/auth/service-accounts'},
+        '/docs/enterprise/enterprise/service-accounts': {redirect: '/docs/enterprise/auth/service-accounts'},
+        //no longer an enterprise only topic? 
+        '/docs/enterprise/concepts/system-flows': {redirect: '/docs/concepts/system-flows'},
```
akeller, line 362: @aj-emerich did system flows become both an OSS and enterprise topic?

aj-emerich: Yeah, it is both since I joined.

nuxt.config.ts
aj-emerich: The main releases page is good. There are no separate pages per edition.

aj-emerich: https://kestra.io/docs/releases

nuxt.config.ts
akeller: https://github.com/kestra-io/docs/pull/3569

nuxt.config.ts
akeller: https://github.com/kestra-io/docs/pull/3570

PR #3564, opened 2026-01-06
docs(backfill): improve backfill vs replay vs recoverMissedSchedules

aj-emerich:


wrussell1999 (approved):

content/docs/06.concepts/08.backfill.md
```diff
@@ -34,12 +34,10 @@ triggers:
 
 This flow run every 30 minutes. However, imagine that your source system had an outage for 5 hours. The flow will miss 10 executions. To replay these missed executions, you can use the backfill feature.
 
-Ensure the backfill’s start and end dates encompass every missed schedule, so the trigger can replay each execution.
+Ensure the backfill’s start and end dates encompass every missed schedule, so the trigger can replay each execution. Note that Backfill does not only replay missed executions in the time window. If there are successful executions, then these are also replayed. To target specific executions, rather than a time window, to avoid duplication use [Replay](./10.replay.md).
```
wrussell1999, line 37: Do we want to instead say "duplicate" instead of replay to avoid confusion with Replay itself?

aj-emerich: I would be hesitant to say "duplicate" as the primary action because, to my understanding, Backfill is a time-window specified bulk replay, and "duplicate" carries an unintentional connotation with it. So I am comfortable with "replay", but my comprehension could be off.

wrussell1999: Let's stick with this for now. It is clear, just wondering if we can avoid any confusion as this is just creating executions without caring about existing ones but that is also clarified throughout.

PR #3593, opened 2026-01-08
(faq): add general orchestration questions

akeller:
I added several questions on general orchestration topics, and linked to a general orchestration blog.


MartinRst:
Good spot, the FAQ at the end of the page was a design placeholder, questions where prepared but not integrated. Easy to fix and we can add more questions!

PR #3746, opened 2026-01-23
refactor: vue to astro for infra

Piyush-r-bhaskar:
Used `Image` component and moved all components to `.astro` components and no scoped in style block as those are automatically scoped by default.


public/landing/infrastructure/tilt-logo.png
elevatebart, line 1: Could we have moved all those files in the `src/component/infrastructure` folder to make a complete separation of concern?

src/components/common/Link.astro
elevatebart, line 1: What is this and what is it used for?

elevatebart: text should be a `<slot/>`

src/components/infrastructure/Built.astro
elevatebart, line 1: you could have kept the vue components here to avoid a change in architecture.

elevatebart (approved):
good job on this one.
A few nitpikcs if you have the time.
If not they can be done later.

src/components/common/Link.astro
Piyush-r-bhaskar: the Link one is used for the chevron to arrow animation on hover. Because this kind of link interaction is being reused across our design,  I made it reusable and currently being used in `infra-automation` and `use-cases/stories` but more to come.

https://github.com/user-attachments/assets/ae5cb384-0c9f-4fb2-a35b-199c08ba88b4

PR #3772, opened 2026-01-23
(docs): explain uri vs. value in outputs of plugins

akeller:
Since this behavior is expected (URI and value are not available together in outputs), I wanted to put this somewhere we could reference it for several plugins. This was the best place I could find, but I'm open to suggestions of other places.

This was also a bit tricky to explain. The execution error comes across as a user error, which is why I wanted to explicitly mention this in the docs.

In the most simplistic way, if `store: true` then outputs does NOT include `value` and only includes a `uri`. If `store: false` (the default value) then outputs does NOT include `uri` and only includes `value`. 

Related internal conversation - https://kestra-io.slack.com/archives/C07KA8R1K0F/p1769182856745799

### DIFF
--- a/src/contents/docs/14.best-practices/0.flows/index.md
+++ b/src/contents/docs/14.best-practices/0.flows/index.md
@@ -54,6 +54,10 @@ This feature is best suited for small datasets, such as querying a few rows to f
 For large data volumes, use the `stores` property instead. Stored outputs are written to Kestra’s internal storage, and only the file URL is referenced in the execution context.
 :::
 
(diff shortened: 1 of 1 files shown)


fdelbrayelle (approved):

wrussell1999 (commented):
Looks good! Is it worth mentioning tasks they have a fetchType that let you decide how you want the data aka URI or output variable?

aj-emerich (approved):

wrussell1999 (approved):

PR #3867, opened 2026-02-03
docs(plugin-developer-guide): update Property docs)

aj-emerich:
Part of https://github.com/kestra-io/kestra/issues/14387

Attempting to correct Property documentation guidance for Plugin Developer Guide. Will let the plugin guys review for any other corrections to suggest :)


src/contents/docs/plugin-developer-guide/07.document/index.md
```diff
@@ -172,7 +172,7 @@ Many tasks can take input from multiple sources on the same property. They usual
     description = "Can be an internal storage URI, a list of Pub/Sub messages, or a single Pub/Sub message.",
     anyOf = {String.class, Message[].class, Message.class}
 )
-private Object from;
+private Property<Object> from;
```
fdelbrayelle: No we need to keep it this way here, because it's dynamic rendering:

```suggestion
private Object from;
```

Issue #3859, opened 2026-01-30
Backfill API not correct in docs


wrussell1999:
### Describe what's missing in our docs

Comment from Zoomcamp suggests the API path on this page is inaccurate: https://kestra.io/docs/concepts/backfill#trigger-backfill-via-an-api-call

> Thanks, Brian and Will. I tried following the example in the docs, but it didn’t work for me. After some digging, I realized the curl endpoint differs in the open-source version: it should be /api/v1/main/triggers (not /api/v1/main/release/triggers, which seems to be for the Enterprise/Release feature)

PR #3873, opened 2026-02-03
docs(backfills): fix urls

wrussell1999:
Closes https://github.com/kestra-io/docs/issues/3859


wrussell1999:
I couldn't find any more instances of that URL in the docs repo but could be elsewhere.

PR #3885, opened 2026-02-04
docs(server-components): link to services status

aj-emerich:


akeller (commented):
I appreciate the level of detail here! I noticed the mermaid diagram didn't really render in the preview site, but I don't know if that's expected.

aj-emerich:
I think I could do it in Nuxt, but maybe I am misremembering. I'll check it.

PR #3904, opened 2026-02-05
fix(home): move dynamic gradient from vue to index.astro

elevatebart:


MilosPaunovic:
Is this still valid @elevatebart?

PR #4091, opened 2026-02-26
docs(flow): fix sample flow that is invalid

aj-emerich:


src/contents/docs/05.workflow-components/01.flow/index.md
```diff
@@ -63,7 +63,6 @@ labels:
 inputs:
   - id: my-value
     type: STRING
-    required: false
     defaults: "default value"
     description: This input is optional.
```
wrussell1999: Technically it's not optional
```suggestion
    description: This input is has a default value.
```

aj-emerich:
Good catch. Didn't read the value 😄

PR #4094, opened 2026-02-26
vs/Dagster minor copy fixes

elliotgunn:
The CTA text is "Teams replace vRO, Rundeck, Cron, Jenkins, and manual scripts with Kestra" when it should be "See how Kestra can simplify your data pipelines—and scale beyond them."

The "try for free" button goes to the demo page when it should go to the quickstart docs


Nico-Kestra (approved):

src/components/versus/Intro.astro
```diff
@@ -39,7 +39,7 @@ const {
                 {description}
             </p>
             <div class="d-flex gap-3 mb-lg-0 mb-4 flex-column flex-md-row">
-                <Link href="/demo" class="btn btn-primary text-white">
+                <Link href="https://kestra.io/docs/quickstart#start-kestra" class="btn btn-primary text-white">
```
wrussell1999: Should it not be this?

```suggestion
                <Link href="/docs/quickstart#start-kestra" class="btn btn-primary text-white">
```

PR #4095, opened 2026-02-26
io.kestra.plugin.notifications.slack.SlackIncomingWebhook as prompted by the flow code linting

akeller:
In 1.3.0-snapshot, I was prompted to adjust the following by the flow code linting:

`type: io.kestra.plugin.slack.SlackIncomingWebhook`

to

`type: io.kestra.plugin.notifications.slack.SlackIncomingWebhook`

### DIFF
--- a/src/contents/docs/03.tutorial/06.errors/index.md
+++ b/src/contents/docs/03.tutorial/06.errors/index.md
@@ -42,7 +42,7 @@ tasks:
 
 errors:
   - id: alert_on_failure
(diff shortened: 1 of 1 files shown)


wrussell1999 (changes requested):
The editor has a bug at the moment that's prompting the old types. It does it for logs and other core tasks too. Notifications is old. Slack is new

wrussell1999:
The bug can be found here https://github.com/kestra-io/kestra/issues/14737

PR #4223, opened 2026-03-12
refactor: avoid use of iframe for animation

elevatebart:
When receiving an animation HTML from claude we have to 
- remove all HTML overhead `<html> `, `<head>` and `<body>` wrappers + titles and meta
- remove all style acting on `body {}`
- remove the script section to play/pause 
- rename the `card` class since it is conflicting with a bootstrap class


Piyush-r-bhaskar (approved):
Very nice one 🎉

Issue #4159, opened 2026-03-09
[EPIC] Blueprints Page Post-QA Website Refresh


Pradumnasaraf:
### Describe the issue

Issue realted to Blueprints Page Post-QA Website Refresh (https://github.com/kestra-io/docs/issues/3975): https://kestra.io/blueprints

Piyush-r-bhaskar:
As most of issue is already done and we have follow up for one remaining so Closing this EPIC.

PR #4239, opened 2026-03-13
fix(blueprints): swap topology and code widths and refit topology on expand

wrussell1999:
The flow code was being squashed and forced to wrap a lot. Also the topology had tons of white space as most examples are long lines. I thought I would swap around the columns and it looks better IMO. WDYT?

Before:
Screenshot 2026-03-13 at 09 56 09

After:
Screenshot 2026-03-13 at 09 56 13

This also re-fits the topology when "see all lines" is clicked.

Before:
Screenshot 2026-03-13 at 10 06 52

After:
Screenshot 2026-03-13 at 10 06 58

Part of https://github.com/kestra-io/docs/issues/4159

### DIFF
--- a/src/components/blueprints/Detail.vue
+++ b/src/components/blueprints/Detail.vue
@@ -48,7 +48,7 @@
 </script>
 
 <style lang="scss" scoped>
-
+    $compact-panel-height: 660px;
 
     section {
         padding: 0 $rem-1;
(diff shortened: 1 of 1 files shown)


wrussell1999:
Before vs After. I think this increase in height is good, but doesn't push the description too far off the page. Ideally we don't want to have the topology look bad.

Screenshot 2026-03-13 at 12 23 43


Screenshot 2026-03-13 at 12 23 47

PR #4282, opened 2026-03-18
[Redirects] Fix 404 for https://kestra.io/docs/quickstart/contributing

akeller:


wrussell1999 (approved):

src/contents/redirects/docs.yml
```diff
@@ -86,3 +86,5 @@
   to: "/docs/tutorial/scripts"
 - regexp: "^/docs/enterprise/apps"
   to: "/docs/enterprise/scalability/apps"
+- regexp: "/docs/quickstart/contributing"
```
aj-emerich:
```suggestion
- regexp: "^/docs/quickstart/contributing"
```

aj-emerich: I think you need this.

aj-emerich (approved):

src/contents/redirects/docs.yml
akeller:
```
- regexp: "/docs/getting-started/contributing"
  to: "/docs/contribute-to-kestra"
```

I followed the pattern from this example a few rows up. It also 404s currently, because it redirects to this fix, but I wasn't sure if I needed the `^`. Happy to add it!

PR #4342, opened 2026-03-24
feat/cache

tchiotludo:
- **fix(website): add explicit cache control public header**
- **fix(website): use caching by default on all api calls**

### DIFF
--- a/src/pages/plugins/[...slug].astro
+++ b/src/pages/plugins/[...slug].astro
@@ -134,9 +134,15 @@ const [
 if (isEePlugin) {
     try {
         const eeData = await $fetchApiCached<{
-            results: { version: string; kestraVersion: string; releaseDate: string | null }[]
+            results: {
+                version: string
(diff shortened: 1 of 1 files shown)


src/utils/fetch.ts
```diff
@@ -32,30 +26,48 @@ export async function $fetchNoCache<T = any>(
...
+}
+
+export async function $fetch<T = any>(
+    url: string,
+    init: RequestInit = {},
+): Promise<T> {
+    const response = await internalFetch(url, init)
+
+    return await response.json()
+}
```
elevatebart, line 39: How about just one function with cloudflare cache by default and if you do not want it you pass `{}`.

It would make the code more readable and the PR more reviewable.

```suggestion
export async function $fetch<T = any>(
    url: string,
    init: RequestInit = cloudflareCache,
): Promise<T> {
    const response = await internalFetch(url, init)

    return await response.json()
}
```

PR #4346, opened 2026-03-25
ci: add SVG embedded image check

AcevedoR:
## Summary
- Adds a new workflow (`.github/workflows/check-svg.yml`) that fails if any `.svg` file contains an embedded image (`data:image/...`)
- Triggers only on pushes/PRs that touch `.svg` files, so there's no overhead on unrelated commits
- Covers all raster formats (PNG, JPEG, GIF, WebP, etc.) via the `data:image/` pattern

## Test plan
- [ ] Push a PR with a clean SVG → check passes
- [ ] Push a PR with an SVG containing a base64-encoded image → check fails with a clear error message


.github/workflows/check-svg.yml
```diff
@@ -0,0 +1,28 @@
+name: Check SVG files
+
+on:
+  push:
+    paths:
+      - "**.svg"
+  pull_request:
+    paths:
+      - "**.svg"
```
elevatebart: PR only on push to avoid double whammy

```suggestion
  push:
    paths:
      - "**.svg"
```

elevatebart (approved):

PR #4382, opened 2026-03-27
fix: no more collection in ssr pages (build size smaller)

elevatebart:
This PR was initially merged with #4376 but since part of it is failing, we need to fix it without blocking the rest.

It will also be easier to review


src/pages/use-cases/stories/[id]-[slug].astro
```diff
@@ -1,63 +1,31 @@
...
+        const storyFiles = await fs.readdir(path.join(storiesPath, storyFolder))
+        const storyFile = storyFiles.find((file) => file.endsWith(".md"))
+        if (storyFile) {
+            const storyContent = await fs.readFile(
+                path.join(storiesPath, storyFolder, storyFile),
+                "utf-8",
+            )
+            const titleMatch = storyContent.match(/title:\s*(.+)/)
+            if (titleMatch) {
+                // redirect to the correct slug
```
Comment (Piyush-r-bhaskar) [inline src/pages/use-cases/stories/[id]-[slug].astro]: story id opened is giving 500

/use-cases/stories/31-modernizing-mission-critical-e-commerce-integrations-with-kestra

PR #4435, opened 2026-03-31
refactor: animated word

Piyush-r-bhaskar:
https://github.com/user-attachments/assets/93c805f8-7ad2-4967-b377-3915140231b2


src/components/home/AnimatedWord.vue
elevatebart, line 1: Wondering if we could use a Vue transition instead.
Less code, clearer options, standard optimization.

Piyush-r-bhaskar: we can yes. Thank you

PR #4571, opened 2026-04-13
refactor: use marked to render html instead of mdc in browser code

elevatebart:


elevatebart:
too old, too far behind

PR #4590, opened 2026-04-14
fix(security): increase HSTS max-age to 1 year

vfanucci:
## Summary

Increases `Strict-Transport-Security` max-age from `15552000` (~180 days) to `31536000` (365 days).

## Why

- **HSTS preload eligibility** — browsers (Chrome, Firefox, Safari) maintain a [preload list](https://hstspreload.org/) of sites that must always be loaded over HTTPS. The minimum requirement is `max-age >= 31536000`. At 15552000 we don't qualify.
- **No downside** — kestra.io is HTTPS-only behind Cloudflare with no reason to ever serve HTTP. The only risk of a longer max-age is being locked into HTTPS for longer, which is a non-issue.

## Change

`src/middlewares/worker.ts` line 62:
```diff
- response.headers.set("strict-transport-security", "max-age=15552000")
+ response.headers.set("strict-transport-security", "max-age=31536000")
```

## Verification

- [ ] Deploy preview → `curl -sI https://<preview-url> | grep strict-transport-security` → `max-age=31536000`


vfanucci:
Very low priority on this one

PR #4596, opened 2026-04-14
Add SLA breach to the `CANCELLED` execution state docs

akeller:
There might be other reasons the execution state can be `CANCELLED` but we only mentioned the concurrency limit previously, and nothing about SLA. I've added it and made a few other small tweaks.


aj-emerich (approved):
LGTM thank you!

wrussell1999 (approved):

PR #4573, opened 2026-04-13
docs(cloudrun): add more information about cloudrun task runner

aj-emerich:
---
  Changes

  Corrections

  - Clarified working directory behaviour: the container only starts in the root directory when no file operations are
  configured; when a bucket is set, the workingDir path is applied to the container automatically
  - Fixed termination description: "container is terminated" → "job is deleted" to match the actual delete: true behaviour
  in code

  Expanded existing sections

  - Termination behavior: added delete: false (keep job for post-failure inspection) and resume: false (force a new job on
  every attempt), with example
  - IAM permissions: already covered; no change

  New sections

  - Syncing the full working directory: documents syncWorkingDirectory: true for downloading the entire working directory
  when output file names are not known in advance
  - Container resources: documents resources.cpu and resources.memory limits, including support for Pebble expressions to
  size containers dynamically from flow inputs
  - Timeout and polling: reference table for waitUntilCompletion (default PT1H), completionCheckInterval (default PT5S),
  and waitForLogInterval (default PT5S)
  - Job retries: documents maxRetries (default 3), clarifying these are Cloud Run task-level retries distinct from
  Kestra-level task retries
  - VPC networking: two subsections — VPC Access Connector (vpcAccessConnector + vpcEgress) and Direct VPC Egress (network
  + subnetwork), with mutual-exclusivity constraint noted
  - Mounting GCS buckets as volumes: documents volumes with bucket, mountPath, and readOnly fields; clarifies this is
  independent of the bucket property used for file transfer
  - Service account configuration: reference table distinguishing serviceAccount (API auth + fallback runtime identity),
  runtimeServiceAccount (container execution identity), and impersonatedServiceAccount (API impersonation), with example

  Removed

  - Empty ## Run tasks on Google Cloud Run H2 that restated the page title with no content
  - Stray --- horizontal rule between the examples section and the setup guide
  - Markdown heading (### Note on termination behavior) inside an alert block
  - Version number from "Before you begin" prose (already captured in frontmatter)

PR #4599, opened 2026-04-14
docs(google-batchrunner): add properties and missing info

aj-emerich:
Changes:
  - Removed the redundant partial containerImage/taskRunner snippet before the
  full flow example
  - Moved the networkInterfaces info callout to sit directly above the full flow
   example
  - Added a Required IAM roles table as a dedicated section before the setup
  steps, consolidating information that was previously buried in step 14

  New content:
  - Expanded the concepts list from 3 to 5 items, adding Compute resources
  (computeResource with a YAML example) and Task retries (maxRetryCount +
  lifecyclePolicies)
  - Added a :::alert{type="warning"} block explicitly calling out that tasks run
   from the root directory, not the working directory — and why {{ workingDir }}
   is mandatory (This was explicit in the source code)
  - Added an environment variables reference table covering WORKING_DIR,
  OUTPUT_DIR, and BUCKET_PATH alongside their Pebble expression equivalents
  - Added the {{ outputDir }} / OUTPUT_DIR pattern as a fourth bullet in the
  "how it works" list
  - Added an info callout explaining worker-restart / job-resume behavior
  - Added a note on the delete / resume defaults and their operational
  implications

### DIFF
--- a/src/contents/docs/task-runners/04.types/06.google-batch-task-runner/index.md
+++ b/src/contents/docs/task-runners/04.types/06.google-batch-task-runner/index.md
@@ -13,11 +13,21 @@ Run tasks as containers on Google Cloud VMs.
 
 The Google Batch task runner deploys a container for each task on a specified Google Cloud Batch VM.
 
-To launch tasks on Google Cloud Batch, you should understand three main concepts:
+To launch tasks on Google Cloud Batch, you should understand five main concepts:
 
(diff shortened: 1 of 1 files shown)


fdelbrayelle (approved):
@aj-emerich Do you also plan to cover the second task runner from plugin-ee-gcp which is `CloudRun`?

aj-emerich:
@fdelbrayelle Yes in a separate PR already :) https://github.com/kestra-io/docs/pull/4573

PR #4606, opened 2026-04-15
seo: decouple H1 from SEO title (5-page sample)

vfanucci:
## Context

The SEO audit identified that `<title>` and `<h1>` are currently identical on every docs page. Best practice is to keep the SEO title optimised for SERP click-through while the H1 is written for the reader landing on the page.

## What this PR does

- **Adds an `h1` optional field** to the docs Zod schema (`content.config.ts`)
- **Updates both doc page templates** (`[...docsPath].astro` and `docs/index.astro`) to render `h1 ?? title` as the visible heading — zero behaviour change for pages that don't set the field
- **Updates 5 pages as a sample** for review before rolling out the full batch (~470 pages)

## Pages updated

| Page | Change |
|------|--------|
| `docs/workflow-components/flow` | H1 added |
| `docs/workflow-components/triggers/schedule-trigger` | H1 added |
| `docs/how-to-guides/alerting` | H1 added |
| `docs/concepts/kv-store` | title colon style + description expanded + H1 added |
| `docs/installation/docker` | H1 added |

## How to review

- Check that the rendered H1 differs from the browser tab title on the 5 pages above
- Confirm no visual regression on pages that don't have the `h1` field (fallback to `title`)
- If approved, the remaining ~464 pages will be applied in a follow-up PR

### DIFF
--- a/src/contents/docs/05.workflow-components/01.flow/index.md
+++ b/src/contents/docs/05.workflow-components/01.flow/index.md
@@ -1,5 +1,6 @@
 ---
 title: Flows in Kestra – Define Orchestration Units
+h1: "Understand Flows: The Core Units of Kestra Orchestration"
 description: Understand Kestra Flows, the fundamental units of orchestration. Learn to define tasks, inputs, outputs, and logic to automate your business processes.
 sidebarTitle: Flow
 icon: /src/contents/docs/icons/flow.svg
--- a/src/contents/docs/05.workflow-components/01.flow/index.md
+++ b/src/contents/docs/05.workflow-components/01.flow/index.md
@@ -1,6 +1,6 @@
 ---
 title: Flows in Kestra – Define Orchestration Units
-h1: "Understand Flows: The Core Units of Kestra Orchestration"
+h1: Understand Flows: The Core Units of Kestra Orchestration
 description: Understand Kestra Flows, the fundamental units of orchestration. Learn to define tasks, inputs, outputs, and logic to automate your business processes.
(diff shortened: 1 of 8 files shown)


src/contents/docs/05.workflow-components/01.flow/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Flows in Kestra – Define Orchestration Units
+h1: "Understand Flows: The Core Units of Kestra Orchestration"
```
aj-emerich:
```suggestion
h1: Understand Flows: The Core Units of Kestra Orchestration
```

aj-emerich (approved):

src/contents/docs/05.workflow-components/01.flow/index.md
wrussell1999: Does this need the quotes so the YAML doesn't get confused? I know we had to add them for blueprints extend block

wrussell1999: Because I don't think you can have a colon inside a string in YAML without wrapping in quotes

aj-emerich: Ah maybe you're right. I can add them back

aj-emerich: Oh nevermind this is resolved

wrussell1999: Ah good

Issue #4688, opened 2026-04-28
[WEBSITE] Add H1 to plugin pages


aj-emerich:
### Describe the issue

We want to start adding `.md` files to plugin parent pages to help improve search. These will include the basics of using the plugin and some FAQs. Currently, some plugins, such as [AWS](https://kestra.io/plugins/plugin-aws), have a "How to use this plugin" section. That header is currently hardcoded as an H2 and does not include the plugin name. A couple of things:

1. Can we pipe in the plugin name into the header, similar to what is done just below with the blueprints?
2. Can we make this an H1 controllable within the `markdown-container` rather than inserted onto the page without control from the plugin's repo (If so, we can forget point 1)? It'll be easier to modify if everything is controlled from the plugin's resources.
3. Is it just parent plugin pages that have the capability for these markdown containers?

<img width="1920" height="856" alt="Image" src="https://github.com/user-attachments/assets/b5c9fbae-ffca-4da1-a50b-600123af7d79" />

</br>

@vfanucci, is there anything else to add from the SEO perspective? Titles and descriptions we can already control from the plugin's resources, and fallbacks are in place in case there is no metadata.

fdelbrayelle:
It's merged on docs @aj-emerich so I'll let you handle https://github.com/kestra-io/plugin-aws/pull/716 and other plugin repos too where you want to improve the docs. Should we keep open this issue? (I'm ok with it if you need a single source of truth).

aj-emerich:
@fdelbrayelle Yeah I will edit the issue title to now include adding markdown to plugins that benefit the most from this.

aj-emerich:
Follow-up to FE merge:

- [ ] Assess the highest value plugins to have "how to use" markdown files
- [ ] Create a prompt for Claude to automate this from the central repository directory
- [ ] Review PRs and close this issue :)

aj-emerich:
Nearly all plugins have `.md` files merged in, and I will steadily start releasing :)

PR #4789, opened 2026-05-13
feat(plugins): extract first heading from longDescription for TOC entry

fdelbrayelle:
## Summary

- Adds `extractFirstHeading` helper in `buildPluginPageProps.ts` that parses the first `#` heading from the plugin's markdown content
- Uses that heading text as the right-sidebar TOC label instead of the hardcoded `"How to use this plugin"` string
- Falls back to `"How to use this plugin"` when the markdown has no heading (backwards-compatible for plugins that haven't added an H1 yet)

## Test plan

- [ ] `/plugins/plugin-aws` right-sidebar TOC shows "How to Use the AWS Plugin" (from the plugin's H1)
- [ ] A plugin whose markdown has no heading still shows "How to use this plugin" in the TOC
- [ ] TOC anchor `#how-to-use-this-plugin` still scrolls to the correct section (anchor is on the wrapper div via the ui-libs change)

Part-of: https://github.com/kestra-io/docs/issues/4688

### DIFF
--- a/src/components/plugins/PluginsIndexWrapper.vue
+++ b/src/components/plugins/PluginsIndexWrapper.vue
@@ -2,8 +2,8 @@
     import { onMounted, onUnmounted, ref } from "vue"
     import { navigate } from "astro:transitions/client"
 
-    import { type Plugin, type PluginMetadata } from "@kestra-io/ui-libs"
-    import PluginIndex from "@kestra-io/ui-libs/src/components/plugins/PluginIndex.vue"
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
Superseded by #4792 which migrates the plugin components directly into docs (per Bart's feedback on kestra-io/kestra#16028 that these are website-only). The TOC heading extraction and H2 removal have been added to that PR instead.

PR #4824, opened 2026-05-19
seo(redirects): /sitemap_index.xml -> /sitemap/index.xml

vfanucci:
## Summary
- Add a 301 redirect from the legacy `/sitemap_index.xml` URL to the canonical `/sitemap/index.xml`.
- Avoids 404s for external links, crawlers and any cached references still pointing at the old path.

## Test plan
- [ ] After deploy, `curl -I https://kestra.io/sitemap_index.xml` returns a 301 to `/sitemap/index.xml`
- [ ] `https://kestra.io/sitemap/index.xml` keeps serving the sitemap index unchanged


vfanucci:
IA crawlers naturally fetch the sitemap here /sitemap_index.xml with the redirection we avoid 404 and better readility

PR #4873, opened 2026-05-27
fix(plugins): prevent 500 on /meta/plugins/group-undefined.svg

vfanucci:
## Summary
- Fallback to the default plugin OG image (`/meta/plugins/{pluginType ?? pluginName}.svg`) when the subgroup lookup returns no match, so `group-undefined.svg` is never emitted into pages.
- Harden the SSR route `src/pages/meta/plugins/group-[cls].svg.ts` to return `404` instead of crashing the worker when the `cls` param is missing/literal `"undefined"`, when the metadata API fails, or when no `title` is returned.

## Context
`https://kestra.io/meta/plugins/group-undefined.svg` was returning a 500. Root cause: in [buildPluginPageProps.ts](src/utils/plugins/buildPluginPageProps.ts), the OG image URL was built with optional chaining on the subgroup lookup, so when no match was found the URL became literally `group-undefined.svg`. The route handler then called the metadata API with `cls=undefined`, the response had no `title`, and the unhandled access crashed the Cloudflare worker.

## Test plan
- [ ] Visit a plugin page that previously generated the bad URL — confirm the `<meta property="og:image">` now points to the default plugin OG image
- [ ] Hit `/meta/plugins/group-undefined.svg` directly — expect `404`, not `500`
- [ ] Hit a valid `/meta/plugins/group-{validCls}.svg` — expect the SVG to render as before


vfanucci:
Hello @fdelbrayelle , I don't know if it'a frontor plugins issue, but the /meta/plugins/group-undefined.svg returns a 500. The crawler hit it a lot (339 hits over the last month)

 Looks like a fallback icon for plugin groups that aren't matched

fdelbrayelle:
Hi @vfanucci :wave: Let's add @Piyush-r-bhaskar and @RemiBarthe also as reviewers to be sure :)

fdelbrayelle (approved):

PR #4919, opened 2026-06-03
ci: index docs on push to releases/* branches instead of tags

brian-mulier-p:
Migrates release pinning from tags to `releases/vX.Y.x` branches (kestra-io/kestra-ee#7397).

- `index-doc.yml` now triggers on pushes to `main` + `releases/*` branches (tag trigger removed); webhook payload format unchanged.
- `versioning-tag.yml` → `versioning-branch.yml`: uses the new `branch-previous-version` composite (kestra-io/actions#112) with the new GH-app token inputs — the old caller was already broken since the actions token refactor (still passed `secrets: toJSON(secrets)`).

**Ordering requirements:**
1. kestra-io/actions#112 merged first (composite referenced `@main`).
2. `releases/*` branches created from existing tags before the flows-engineering indexing change deploys.
3. After merge, the new `index-doc.yml` must be backported to each release branch — push triggers evaluate the workflow file on the pushed ref, so branches created from old tags carry the old workflow and fire nothing until backported.

**Secret requirement:** `GH_BOT_APP_ID` / `GH_BOT_PRIVATE_KEY` must be visible to this repo (could not verify without org admin).


fhussonnois (approved):
LGTM!

PR #4928, opened 2026-06-04
docs: document namespace and subkey arguments for secret()

brian-mulier-p:
- Document the optional `namespace` argument on `secret()` — reads a secret from another namespace using that namespace's own secret backend (available as of 1.3).
- Document the `subkey` argument for extracting a field from a JSON secret.
- Add a cross-namespace example and security note to the Secrets concept page (same-tenant only; allowed by default, restrict via `allowedNamespaces`).

Companion PR: kestra-io/kestra (basic.md in-app reference).


aj-emerich (approved):
Thanks Brian, LGTM

PR #4929, opened 2026-06-04
feat(docs): serve versioned documentation via Astro middleware

brian-mulier-p:
Serve `kestra.io/docs/{MAJOR.MINOR}/{path}` (e.g. `/docs/1.2/...`) for past releases, styled to match the current `/docs` design (sidebar, header, TOC, breadcrumb, prev/next) by rendering through the real Astro layout and components instead of a hand-styled standalone page.

## What
- **Middleware** (`src/middleware.ts`) regex-gates `/docs/{major.minor}/...`, fetches the version's raw markdown from api.kestra.io, and rewrites in-chain to a new `prerender = false` page (`src/pages/docs-versioned.astro`) via Astro's `next("/docs-versioned")` — the URL in the browser stays `/docs/{version}/...`. A missing page or failed fetch redirects to the version home (or latest, avoiding a redirect loop). `.md` requests are passed through as raw markdown, mirroring the latest docs' "append `.md` to any `/docs/*` URL" convention.
- **`renderVersionedDocBody`** (`src/utils/renderVersionedDoc.ts`) parses the markdown with `@nuxtjs/mdc`'s `createMarkdownParser` and remaps the known MDC components (`::alert`, `::collapse`, `::badge`, child-card grids, CTA buttons, support links) to the *site's own* CSS classes, so the body HTML drops straight into the real `DocsLayout` and inherits current styling for free. Unknown/relic `::` blocks degrade to their inner content instead of leaking directive syntax.
- **Real layout reuse**: `docs-versioned.astro` renders through the actual `DocsLayout`/`NavSideBar`/`NavToc`/`Breadcrumb`/`PrevNext` components — same sidebar, header, right-rail TOC, and dark-mode handling as the latest docs, achieved by construction rather than duplicated CSS.
- **`versionedNavigation.ts`** adapts the flat children map the API returns into the `NavigationItem` tree these components expect, reusing the latest docs' curated section groupings (`getNavigationTree.ts`) when enough of a version's page titles match, falling back to a flat tree for older/renamed releases.
- **Version `<select>`** added to `NavSideBar.vue`, available on both latest and versioned pages, switching directly between them.
- **SEO**: every versioned response gets `<meta name="robots" content="noindex">` and an `X-Robots-Tag: noindex` header; versioned URLs are not in the sitemap.
- **vitest + 126 unit tests** across three suites (`versionedDocs.test.ts`, `renderVersionedDoc.test.ts`, `versionedNavigation.test.ts`). Verified end-to-end on both `astro dev` and a full `wrangler dev` (Cloudflare Workers) build.

## Limitations (accepted degradations)
- Search and Ask-AI query the **latest** docs corpus, not the version being viewed (labeled "Search (Latest)" in the sidebar).
- `:PluginCount` (a live-count JS island with no SSR equivalent) shows a static "hundreds of" instead of the real number.
- Older versions (pre-1.0-era children data) get a flat sidebar tree, not curated section groupings.

Closes https://github.com/kestra-io/kestra-ee/issues/7397


anna-geller:
a matter of adjusting the design (change between regular and versioned view) and should be handed over soon

aj-emerich:
@brian-mulier-p Do you think we need a stronger visual indicator that they are on an older version of the documentation like we do with the plugins?

Screenshot 2026-09-02 at 14 47 44

src/pages/docs-versioned.astro
```diff
@@ -0,0 +1,154 @@
...
+    // delegation survives both. `.astro-code` scopes it to this renderer's
+    // buttons (SchemaToCode's `.copy` lives under `.code-block`).
+    document.addEventListener("click", ({ target }) => {
+        const button = (target as HTMLElement).closest<HTMLButtonElement>(".astro-code .copy")
+        const code = button?.parentElement?.querySelector("code")
+        if (!button || !code || !navigator.clipboard) return
+
+        void navigator.clipboard.writeText(code.textContent?.trimEnd() ?? "")
+        button.classList.add("copied")
+        setTimeout(() => button.classList.remove("copied"), 2000)
```
elevatebart: Why not use a vue component here since we load the Vue runtime anyways?

src/components/docs/NavSideBar.vue
```diff
@@ -34,14 +34,31 @@
...
+                    @change="onVersionChange"
+                >
+                    <option
+                        v-for="option in versionOptions"
+                        :key="option.label"
+                        :value="option.version"
+                        :selected="option.selected"
+                    >
+                        {{ option.label }}
+                    </option>
```
elevatebart, line 59: The caret is too close to the edge.

Image

src/components/docs/NavSideBar.vue
```diff
@@ -84,8 +102,47 @@
...
+        },
+        // No `default: () => []` here: combining a factory default with an
+        // array PropType breaks this component's inferred type for TS
+        // consumers (astro check reports the whole module as having "no
+        // default export"). Default to [] at the read site instead.
+        knownVersions: {
+            type: Array as PropType<string[]>,
+        },
+        /** Version label of the page being viewed (e.g. "1.2"), unset on the latest docs. */
+        currentVersion: {
```
elevatebart, line 118: This could (should?) be a v-model since we pass down values and emit updates of the said value.

you can use `defineModel()` to create it.

elevatebart (approved):
It seems to work fine so my only concern is for performance.
Will it keep the fast, speedy SEOptimized toolset that we have.

Also is search adapted to this new version?

A few nits but all good in my book.

brian-mulier-p:
> It seems to work fine so my only concern is for performance. Will it keep the fast, speedy SEOptimized toolset that we have.

The versioned doc is not indexed + I saw @vfanucci already prepared a PR to exclude crawlers definitely on those pages (even if noindex should be enough already, more protection is always better): https://github.com/kestra-io/docs/pull/5308

> Also is search adapted to this new version?

I just introduced search for the currently browsed version instead of latest only so now it should be fine

> A few nits but all good in my book.

Will fix, ty for the review :rocket:

vfanucci:
Hey ! 

Thanks for checking the SEO :) 

SEO side looks right to me, nothing to change

- noindex, follow in the meta and X-Robots-Tag, on both the HTML and the .md passthrough 
- self-canonical on the versioned URL rather than a cross-canonical to latest — correct call, the content genuinely differs 
- out of the sitemap, and no crawlable link from latest → versioned (the <select> emits no href) ✅

That's everything Google needs.

One question,worker.ts adds VERSIONED_DOCS_PATH to isEdgeCacheablePage, but the middleware sets X-Robots-Tag after next() — could you confirm on the preview that a cache HIT still returns the header? The meta tag is the real safety net

I'm keeping #5308 as a draft on purpose I'll monitor AI-bot hits on /docs/1.* in the Cloudflare logs once this is live and merge it if it picks up and change the robots.txt if needed

PR #4932, opened 2026-06-05
fix(dev): use Node adapter for astro dev so images render locally

### LINKED ISSUE #4887: [Docs Locally]: Running the docs server locally lacks CSS/JS injection
### Describe the issue

Trying to run the docs locally results in a broken site - loads of 404s in the browser console for failed styling and components. Prohibits testing changes locally. I am using Brave browser, but it was also reported by a Windows user.

  Environment:
  - Astro v6.1.5 with @astrojs/cloudflare adapter
  - Node.js on macOS (Darwin 24.6.0)
  - Confirmed after fresh npm install (packages were up to date)

  Steps to reproduce: npm run dev, open localhost:4321

  Symptoms:
  1. All Vite virtual modules return 404 — /@vite/client, /@id/@astrojs/vue/client.js, /@id/astro:scripts/before-hydration.js all
   return 404, so Vue <astro-island> components never hydrate
  2. Page is 1.35MB — all CSS is inlined across 50+ <style> blocks rather than served as separate files; this makes the initial
  render extremely slow
  3. No external <script> or <link> tags in the <head> (only one inline theme-detection script), so Vite's module graph never
  loads
  4. Bootstrap CSS not linking — it's expected to be compiled and available but isn't being referenced via a link/module tag

  What works: HTTP 200 on the root URL, fonts (/_astro/fonts/) serve correctly, the HTML structure is fully rendered server-side

  Suspected cause: The @astrojs/cloudflare adapter may be interfering with Vite's dev-mode asset serving. The workers:dev script
  (wrangler dev -- npm run dev) is the typical Cloudflare dev path — astro dev alone may not be the right entrypoint for this
  adapter setup.

Image

https://github.com/user-attachments/assets/e815d74e-8d5b-4dbe-94a7-cb9a5bf2b858

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/docs/issues/4887

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -1,8 +1,9 @@
 // @ts-check
-import { defineConfig, envField, fontProviders } from "astro/config"
+import {defineConfig, envField, fontProviders, svgoOptimizer} from "astro/config"
 
 import * as path from "path"
 import cloudflare from "@astrojs/cloudflare"
+import node from "@astrojs/node"
 import vue from "@astrojs/vue"
 import mdx from "@astrojs/mdx"
 import icon from "astro-icon"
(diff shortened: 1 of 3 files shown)


elevatebart (approved):
Agreed! that's much better.

NOTE: I was hoping that this would be done directly by the Cloudflare adapter. Because this introduces a difference between built and dev environment.

PR #4969, opened 2026-06-15
seo: link plugin logos to group page + fix dead EE redirect rules

vfanucci:
## Context
Technical SEO audit (Screaming Frog, 2026-06-15): **~1,207 internal links** point to URLs that return a **301**, concentrated on `/use-cases/*` and `/features/*`. Source = the plugin logo carousels.

## Root cause
The three logo components build links as `/plugins/${artifactId}/${name}`. For **~82 plugins** the `/${name}` subgroup URL does not exist (404) and the middleware (`notFoundRedirect`) 301-redirects it to the group page `/plugins/${artifactId}`. Each affected logo is a wasted 404→301 hop (×105 for the 4 most frequent: airtable-record, argo-cd-apps, ansible-cli, airflow-dags).

Note: **385 `/plugins/plugin-X/name` subgroup pages do exist** (200). Since these are decorative logo links, all logos now point at the **group page** (always 200) — the semantically correct destination for a logo, with no redirect. *If the team prefers to keep deep-links for the 385 valid ones, the `name → slug` mapping should instead be made reliable in the plugin metadata.*

## Changes
- `PluginsBlock.astro`, `EveryPlugin.astro`, `PluginsBox.astro`: logo link → `/plugins/${artifactId}`.
- `src/contents/redirects/plugins.yml`: the middleware applies the **first** matching rule (`allEntries[0]`). The catch-all `/plugins/plugin-([^/]+).*` was placed **before** the specific EE rules (aws/azure/gcp/kubernetes runner), so those never fired. Moved the catch-all to the end, fixed a leading-space typo in the `plugin-azure` rule, and pointed the EE rules directly at their final 200 URL (adding the `aws-batch-task-runner` / `azure-runner` / `google-cloud-task-runner` subgroup) so reactivating them is a single 301, not a chain.

## Verified
- Group destinations `/plugins/plugin-{airtable,argocd,ansible,airflow,trello,jira,microsoft-fabric}` confirmed **200** in the crawl.
- All EE rule targets confirmed **200 in a single hop** live.
- `plugins.yml` validated (YAML parses, catch-all last, no leading-space rule).


iitzIrFan (approved):
Nice patch @vfanucci 🙌

PR #4986, opened 2026-06-18
update docs with cleaner taskrunner

MartinRst:


aj-emerich (approved):
lgtm

MartinRst:
problem solved don't touch on the doc, just add a warning that it does not work with cloud version to use the process task runner

PR #4988, opened 2026-06-18
Move sorting logic into getNavigationTree function to ensure all components receive consistently sorted results. 

### LINKED ISSUE #4987: The API Reference page shows inconsistent ordering between the sidebar navigation and the generated child navigation section.
### Describe the issue

In the sidebar:

- Cloud & Enterprise Edition API
- Open Source API
- SDKs

However, the generated child navigation on the page renders:

- Cloud & Enterprise Edition API
- SDKs
- Open Source API

---

Ref img: 

Image

---

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/4987

---

- Root Cause: The navigation tree sorting logic existed only in `buildDocsNavigation.ts` Components like `ChildTableOfContents.astro` called `getCollection("docs")` directly and `passed unsorted results to getNavigationTree,` which didn't apply any sorting internally.

- Solution: Moved the sorting logic into `getNavigationTree.ts function`. This ensures all components using `getNavigationTree `receive consistently sorted results, eliminating code duplication and guaranteeing alphabetical file path sorting respects numeric prefixes.

---

Ref img: 
Fix

---


aj-emerich:
Wow nice spot. LGTM

aj-emerich (approved):

aj-emerich:
@kestra-io/frontend anything else to consider with such a change?

src/utils/buildDocsNavigation.ts
```diff
@@ -3,12 +3,6 @@ import { getNavigationTree } from "./getNavigationTree"
 
 export async function buildDocsNavigation() {
     const docsPages = await getCollection("docs")
```
elevatebart, line 5: Should this be `getNavigationTree`

```suggestion
    const docsPages = getNavigationTree(await getCollection("docs"))
```

elevatebart (commented):
Maybe check that small issue.

src/utils/buildDocsNavigation.ts
iitzIrFan: One doubt that, after this change the `docsPage` will be the `navigation tree` instead of `raw docs pages` Also would need some change in `getNavigationTree` for it and don't think this would be optimal though?

elevatebart (approved):

PR #5004, opened 2026-06-22
docs: add dotnet script guide

aj-emerich:
Part of https://github.com/kestra-io/plugin-scripts/issues/375

All examples tested and working:

Screenshot 2026-06-22 at 14 00 38


src/contents/docs/15.how-to-guides/dotnet/index.md
```diff
@@ -0,0 +1,158 @@
+---
```
fdelbrayelle, line 1: I think it worth mentionning that `mcr.microsoft.com/dotnet/sdk:10.0` is the default image but it can be overriden with containerImage to use an older SDK version for instance.

fdelbrayelle (approved):

src/contents/docs/15.how-to-guides/dotnet/index.md
aj-emerich: 💯

PR #5115, opened 2026-07-09
fix(plugins): hide the `longDescription` on subgroups

### LINKED ISSUE #5113: [Plugin Pages] How to should only be on root page
### Describe the issue

The how-tos for the Plugins are meant to be high-level and sit on the root plugin page (https://kestra.io/plugins/plugin-redis), but right now they are also showing up on all the children task pages (https://kestra.io/plugins/plugin-redis/redis-vector).

Let's update so it only appears on the root. If we want to make dedicated guides for the task groups, then we can revisit. If possible, let's fix in a way that opens up that possibility.

Piyush-r-bhaskar:
closes https://github.com/kestra-io/docs/issues/5113

### DIFF
--- a/src/components/plugins/PluginIndex.vue
+++ b/src/components/plugins/PluginIndex.vue
@@ -52,7 +52,7 @@
                 </div>
             </div>
         </template>
-        <template v-if="description !== undefined && plugin?.longDescription">
+        <template v-if="showLongDescription && description !== undefined && plugin?.longDescription">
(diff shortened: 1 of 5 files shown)


Piyush-r-bhaskar:
AJ , could you test and if all good, feel free to merge.

fdelbrayelle (approved):

aj-emerich (approved):

PR #5129, opened 2026-07-10
docs: better clarify cloud with AWS batch

### LINKED ISSUE #5127: Docs: AWS Batch task runner page has no Kestra Cloud auth section (STS reads as keyless but needs static keys on Cloud)
## Summary

The [AWS Batch Task Runner](https://kestra.io/docs/task-runners/types/aws-batch-task-runner) page is tagged for both `EE` and `Cloud`, but its authentication guidance and every example assume a **self-hosted** control plane running inside AWS. It never covers how **Kestra Cloud** (whose control plane runs outside the customer's AWS account, so it has no ambient AWS identity) authenticates to a customer's AWS account. A Cloud customer running the Batch runner against their own AWS lost ~2 days to this, and it contributed to a sustained run of "failed" executions that were actually AWS IAM/role misconfiguration rather than Kestra problems. Three concrete gaps below.

## 1. The "STS role assumption" section reads as keyless, but is not on Cloud

The [STS role assumption](https://kestra.io/docs/task-runners/types/aws-batch-task-runner#sts-role-assumption) section opens with:

> Instead of static `accessKeyId` / `secretKeyId` credentials, you can authenticate via AWS STS `AssumeRole` for cross-account access or short-lived credentials

and the example shows only `stsRoleArn`, with no base credentials, which reads as a fully keyless path:

```yaml
taskRunner:
  type: io.kestra.plugin.ee.aws.runner.Batch
  region: eu-central-1
  stsRoleArn: "arn:aws:iam::123456789012:role/kestra-batch-role"
```

On Kestra Cloud there is no EC2 instance profile or EKS IRSA to bootstrap the `sts:AssumeRole` call, so the AWS SDK still needs base `accessKeyId` / `secretKeyId` credentials to assume the role. STS on Cloud is therefore **not** keyless, which breaks a "no long-lived static credentials" security policy.

**Requested change:** add a Cloud vs self-hosted note to that section. State that keyless STS works only when Kestra runs on AWS (an EC2 instance profile or EKS IRSA supplies the base identity), and that on **Kestra Cloud you must supply base static credentials to assume the role**. Say plainly that there is currently no fully keyless Cloud to customer-AWS path.

## 2. OIDC / AWS Web Identity Federation for Cloud to AWS is not documented

Nothing on the page (or in the Cloud edition docs) states whether Kestra Cloud can act as an OIDC identity provider so AWS can trust it via `AssumeRoleWithWebIdentity` (the way GitHub Actions does with `token.actions.githubusercontent.com`). Customers with a "no static keys" policy expect this and had to email support to learn it is not available.

**Requested change:** state explicitly whether OIDC / Web Identity Federation from Kestra Cloud to AWS is supported (today: no) and give the recommended alternative. If it ships later, document the provider URL and setup steps.

## 3. There is no "Running from Kestra Cloud" section

The "Minimum permissions required" and setup guides assume the IAM principal and the control plane live in the customer's AWS. The EKS section covers IRSA for the *pod*, but not how the *Cloud control plane* authenticates to the AWS Batch API, nor the network path from Cloud into a customer VPC.

**Requested change:** add a "Running from Kestra Cloud" subsection covering (a) how the Cloud control plane authenticates to a customer's AWS (cross-link to gaps 1 and 2 above), (b) the **Kestra Cloud egress IP ranges to allowlist** in the customer's security groups / API gateways, and (c) a pointer to cost considerations for cross-cloud runners.

## Suggested placement

Add the "Running from Kestra Cloud" section to the [AWS Batch Task Runner](https://kestra.io/docs/task-runners/types/aws-batch-task-runner) page, plus short callouts near the top of "Minimum permissions required" and "STS role assumption". Cross-link from the Cloud / EE edition docs.

## Already addressed (no action needed)

The related CloudWatch Live Tail cost and the STS-expiry stream failure are already documented via `streamLogs: false` in the [Log streaming](https://kestra.io/docs/task-runners/types/aws-batch-task-runner#log-streaming) section, so this issue is scoped to authentication only.

## Note for triage

Gap 2 (OIDC) and the egress-IP allowlist in gap 3 are partly product gaps, not only docs. Docs can describe today's reality (no keyless path); a follow-up product issue may be needed for the OIDC endpoint and published egress ranges.

aj-emerich:
Closes https://github.com/kestra-io/docs/issues/5127

### DIFF
--- a/src/contents/docs/task-runners/04.types/04.aws-batch-task-runner/index.md
+++ b/src/contents/docs/task-runners/04.types/04.aws-batch-task-runner/index.md
@@ -51,6 +51,10 @@ The task runner maps AWS Batch job statuses to exit codes as follows:
 
 ## Minimum permissions required
 
+:::alert{type="warning"}
+**Running on Kestra Cloud?** The Cloud control plane runs outside your AWS account and has no EC2 instance profile or EKS IRSA to supply a base AWS identity. Static `accessKeyId` / `secretKeyId` credentials are required. See [Running from Kestra Cloud](#running-from-kestra-cloud) for the full details before configuring authentication.
+:::
+
 To submit and monitor AWS Batch jobs, the IAM principal used by Kestra needs permission to create, tag, inspect, and clean up Batch job definitions and jobs. It also needs permission to pass the ECS roles used by the job and to read the AWS Batch log group.
 
 The following policy is the minimum set required by the task runner:
@@ -380,8 +384,8 @@ Once both resources are created, copy the compute environment ARN and job queue 
 
 To use the AWS Batch task runner, you must configure resources in your AWS account. You can set up the environment in two ways:
 
-1. Using Terraform to provision all necessary resources using a simple `terraform apply` command.
-2. Creating the resources step by step from the AWS Management Console.
+1. Use Terraform to provision all required resources with a single `terraform apply` command.
+2. Create the resources step by step from the AWS Management Console.
 
 
 <div class="video-container">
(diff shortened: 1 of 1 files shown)


soyasis (approved):
Thanks @aj-emerich, all covered now!

PR #5140, opened 2026-07-14
docs: improve task runner types list

aj-emerich:


wrussell1999 (approved):
Worth checking with @vfanucci if this is for SEO but otherwise we should remove the extra mentions of “Kestra”

PR #5151, opened 2026-07-15
blog: fix OSS/EE mismatch in closing CTAs

elliotgunn:
Both posts closed with an open-source-only CTA right after covering EE-only features (worker groups, tenants), reads as if the OSS edition gets you what the post just described.


xdatanomad (approved):
👍

PR #5169, opened 2026-07-17
fix(prevnext): improve button alignment

### LINKED ISSUE #4194: Next and Previous Buttons are misaligned (no text and too much text)
### Describe the issue

Image

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/4194

---

Preview img: 
Screenshot 2026-07-17 at 11 50 34 AM


elevatebart:
I don't mind the change, this was simply what was specified as design and I liked it.
It set the focus on next and not on previous. 

let's move "Forward" 😄

iitzIrFan:
Now it has a bit more symmetry though :) Let's see what Nico thinks from the design perspective.

Nico-Kestra:
Capture d’écran 2026-07-21 à 09 14 48

don't know, originally it was pretty simple, and 'previous' was way less emphasized. The flow is more like: 'I'm reading something—could the next thing interest me? Probably less so the previous one.' So yeah, right now it's too big, too cluttered, too much info for a module that's just supposed to do previous/next, don't you think?

iitzIrFan (commented):
Thanks for the explanation @Nico-Kestra  Should I go ahead and close the PR (and the issue as well), since the current design is intentional?

PR #5174, opened 2026-07-17
fix(ndjson): correct code block syntax from ndjson to json

### LINKED ISSUE #5173: Use json syntax highlighting for NDJSON examples
### Describe the issue

The NDJSON examples currently use the ndjson language identifier, which isn't supported by astro-expressive-code. This results in a build warning and falls back to plain text highlighting.

---

Preview img: 

Image

---

Tentative soln:
We can simply use json instead, which provides proper syntax highlighting without requiring any additional configuration.

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5173

---


vfanucci (approved):
LGTM

PR #5178, opened 2026-07-17
docs: 2.0 Blog Post

aj-emerich:
## Summary

Draft of the Kestra 2.0 release blog post at `src/contents/blogs/2026-09-01-release-2-0/index.md`.

Covers all major 2.0 features in the order of their relative weight:
- MCP Tool Trigger + MCP Server
- Worker Groups 2.0
- RBAC action-based permissions
- Loop task (ForEach/ForEachItem replacement)
- Trigger `when` expression
- kestractl IAM commands
- Blueprint version control
- AWS EC2 Task Runner
- PurgeStorage
- Additional improvements (Reusable Inputs, ION binary, TRACEPARENT, Syslog CEF, AI Agent metrics, etc.)
- Breaking changes summary table + migration links

## Before publishing

- [x] Add author frontmatter (name, image, linkedin)
- [x] Add `main.jpg` hero image
- [x] Add any available feature screenshots (MCP server Connect tab, Worker Groups UI, etc.)
- [x] Confirm release date is still 2026-09-01 (update frontmatter if not)
- [x] Verify all internal doc links resolve on the merged branch
- [x] Review the `toolDescription` example for accuracy against the shipped trigger
- [x] Confirm Worker Groups fallback default flip (WAIT to FAIL) with engineering before calling it out in the post

## Notes

PR targets `docs/2.0.0-release` so it flows through with the main docs release. When that branch merges to main, this blog goes live with it.


src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+authors:
+  - name: "AJ Emerich"
+    image: aemerich
+    linkedin: https://www.linkedin.com/in/alex-emerich/
+image: ./main.jpg
+---
+
+<!-- TODO: hero image (main.jpg) needed -->
+
+Kestra 2.0 ships with intentional breaking changes. Version 2.0 removes constructs that accumulated technical debt across three release cycles (ForEach, Java-class trigger conditions, CRUD-based RBAC permissions) and replaces them with designs that hold up under production load, multi-team governance, and the increasingly common requirement that your orchestration platform be reachable by AI agents.
```
Coding-with-Adam: Maybe it would behoove us to tackle the elephant in the room (few OSS features) from the very beginning and tweak the inro slightly:
> Alongside intentional breaking changes, Kestra 2.0 places a focus on Enterprise enhancements, though our open source users will find new additions to enjoy as well. Version 2.0 removes...

OR

> "While Kestra 2.0 introduces a strong suite of Enterprise features, we’ve also included a few open source enhancements alongside a few intentional breaking changes. Version 2.0 removes...

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+| Feature | What | Edition |
+|---|---|---|
+| MCP Tool Trigger + MCP Server | Expose flows as tools callable by AI agents | EE, Cloud |
+| AI Copilot agentic loop | Three-mode chat sidebar (Edit, Plan, Ask) with multi-turn memory and confirmation gate | EE, Cloud |
+| Worker Groups 2.0 | Tag-based routing, capacity reservation, JWT auth | EE |
+| RBAC action-based permissions | Resource + action model replaces CRUD | EE |
+| Policies | Namespace-scoped governance rules replace `pluginDefaults` | EE |
+| Loop task | Replaces ForEach and ForEachItem with isolated sub-executions | All |
+| Trigger `when` expression | Pebble replaces Java-class conditions on all trigger types | All |
```
Coding-with-Adam: Wherever it say All under the Edition column, can we change it to: `OSS, EE, Cloud`.
I think that would help highlight OSS a little more

src/contents/blogs/release-2-0/main.jpg
wrussell1999, line 1: This needs to be updated. Just a placeholder so Cloudflare builds.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+authors:
+  - name: "AJ Emerich"
+    image: aemerich
+    linkedin: https://www.linkedin.com/in/alex-emerich/
+image: ./main.jpg
+---
+
+<!-- TODO: hero image (main.jpg) needed -->
+
+Kestra 2.0 ships with intentional breaking changes. Version 2.0 removes constructs that accumulated technical debt across three release cycles (ForEach, Java-class trigger conditions, CRUD-based RBAC permissions) and replaces them with designs that hold up under production load, multi-team governance, and the increasingly common requirement that your orchestration platform be reachable by AI agents.
```
wrussell1999: I'm wondering if opening with the breaking changes is a bit much. Maybe more like "2.0 enables this, but in order to enable it there are breaking changes etc."

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+The headline capability is MCP: Kestra flows can now be exposed as typed tools on an MCP server and invoked directly by Claude, Cursor, Codex, or any MCP-compatible agent. A class of AI-driven automation that previously required bespoke integrations now works out of the box. The AI Copilot is rebuilt from the ground up as a persistent agentic loop with three modes, multi-turn memory, and a confirmation gate before any mutations run. Worker Groups are redesigned from scratch with tag-based routing, capacity reservation, and JWT-based worker authentication. RBAC moves from generic CRUD on resources to resource-plus-action pairs, so you can grant a CI service account the right to execute flows without being able to delete them.
```
wrussell1999: > The headline capability is MCP

Or is it the architecture rework? Not that it's a feature really. Just a thought.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+All executions created via MCP are tagged with `system.from: mcp`, `system.mcpServerId`, and `system.mcpSessionId`, so you can filter by agent origin in the execution list.
+
+For access control, the `MCP_SERVER` resource in the EE RBAC model governs who can create and manage MCP servers. A user needs `FLOW: EXECUTE` permission on at least one namespace with a registered `McpToolTrigger` to call the matching tool.
+
+See the [MCP server docs](/docs/ai-tools/mcp-server) and [McpToolTrigger reference](/docs/workflow-components/triggers/mcp-tool-trigger) for full setup.
+
+## AI Copilot
+
+The AI Copilot is rebuilt in 2.0. The old one-shot generation modal is replaced by a persistent right-sidebar chat panel, opened via the **AI** button in the top toolbar. Conversations are multi-turn and held in memory for the browser session. Click **New chat +** to start fresh; use **Recents** to return to a prior conversation.
```
wrussell1999: The right-sidebar chat panel might change. Maybe reword it as it has a dedicated page now.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
+---
+title: "Kestra 2.0: MCP Tools, Worker Groups, and a Leaner Core"
+description: "Kestra 2.0 makes flows callable by AI agents via MCP, redesigns Worker Groups with tag-based routing and JWT auth, brings action-based RBAC, and removes legacy constructs in favor of cleaner primitives."
+date: 2026-09-08T10:00:00
```
wrussell1999, line 4: PR description says the release date is 2026-09-01, the frontmatter says 2026-09-08. Pick one before merge.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+category: News & Product Updates
+authors:
+  - name: "AJ Emerich"
+    image: aemerich
+    linkedin: https://www.linkedin.com/in/alex-emerich/
+image: ./main.jpg
+---
+
+<!-- TODO: hero image (main.jpg) needed -->
```
wrussell1999: Update `main.jpg` to the new one and remove comment.

The other three (L39, L43, L379) still need resolving or removing before publish, otherwise they ship as HTML comments in the page source.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+Kestra 2.0 ships with intentional breaking changes. Version 2.0 removes constructs that accumulated technical debt across three release cycles (ForEach, Java-class trigger conditions, CRUD-based RBAC permissions) and replaces them with designs that hold up under production load, multi-team governance, and the increasingly common requirement that your orchestration platform be reachable by AI agents.
+
+The headline capability is MCP: Kestra flows can now be exposed as typed tools on an MCP server and invoked directly by Claude, Cursor, Codex, or any MCP-compatible agent. A class of AI-driven automation that previously required bespoke integrations now works out of the box. The AI Copilot is rebuilt from the ground up as a persistent agentic loop with three modes, multi-turn memory, and a confirmation gate before any mutations run. Worker Groups are redesigned from scratch with tag-based routing, capacity reservation, and JWT-based worker authentication. RBAC moves from generic CRUD on resources to resource-plus-action pairs, so you can grant a CI service account the right to execute flows without being able to delete them.
+
+This post covers what shipped.
```
wrussell1999: The opener stacks three CLAUDE.md problems.

1. "ships with intentional breaking changes" leads a release announcement with its own liability. "intentional" is also doing rhetorical work, nobody claims their breaking changes were accidental.
2. Back-to-back tricolons. L15: "production load, multi-team governance, and the increasingly common requirement...". L17: "three modes, multi-turn memory, and a confirmation gate", then "tag-based routing, capacity reservation, and JWT-based worker authentication". Three rule-of-three constructions in two paragraphs.
3. "This post covers what shipped." is a one-line fragment paragraph doing manufactured emphasis, and it restates the title. Cut it, the table underneath already announces the structure.

Separately, the title promises "a Leaner Core" and the post then documents fifteen additions. The removals are real but "leaner" isn't what the body argues.

src/contents/blogs/release-2-0/index.md
wrussell1999: "rebuilt from the ground up" and "redesigned from scratch" are the same cliche twice in adjacent sentences. It then recurs at L86 ("is rebuilt in 2.0") and L106 ("a ground-up redesign").

The paragraph is also anaphoric: three consecutive sentences open "The AI Copilot is rebuilt...", "Worker Groups are redesigned...", "RBAC moves from...".

Say what changed instead. "The Copilot is now a persistent chat panel with three modes" is shorter and more informative than "rebuilt from the ground up as a persistent agentic loop".

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+When you open the sidebar while viewing a resource, that resource attaches automatically as a context pill above the input. Pills are independently dismissible. Each add and remove is recorded in the transcript so you can always see what the agent is looking at. Attachable resources include flows, namespaces, executions, dashboards, apps, test suites, blueprints, and plugins. The Copilot also reads namespace metadata (Policies, Variables, Secrets, Key-Value pairs) to ground authoring suggestions against your actual configuration, so prompts like "create a task that reads from our MongoDB" can reuse configured credentials without extra hints.
+
+Actions that modify resources require explicit confirmation before the Copilot executes them. A prompt appears in the chat with an optional field to steer the next attempt. Approving applies the change; rejecting resumes the conversation in Edit mode, or cancels the current plan in Plan mode.
+
+The `COPILOT` resource in the RBAC model controls who can use the feature. Assign `VIEW` or `DENY` at tenant or namespace scope via the Roles UI.
```
wrussell1999: Wrong action name. `migration-guide/v2.0.0/rbac-action-model` maps old `AI_COPILOT: READ` to `COPILOT: USE`. Neither `VIEW` nor `DENY` appears in the action table for this resource, and `DENY` isn't an action at all, it's how a binding is applied.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+        - a100
+      match: ALL
+      fallback: FAIL
+```
+
+![Diagram showing how developers read task-to-queue top-down and operators read workers-to-group-to-queue bottom-up, meeting at the Worker Queue layer with a many-to-many relationship](./worker-groups-routing.png)
+
+A Worker Group subscribes to one or more queues. The platform routes a task to the first available group that covers all required tags (or any, with `match: ANY`). `fallback` controls what happens when no matching group is available: `FAIL` (the default in 2.0, changed from the 1.x default of `WAIT`), `WAIT`, `CANCEL`, or `IGNORE` (drop the requirement and route to the default queue).
+
+The fallback default flip is the sharpest gotcha for upgraders. Tasks that previously waited silently for a matching worker will now fail immediately. Set `fallback: WAIT` explicitly on tasks where the old behavior was intentional.
```
wrussell1999: Incomplete, and the omission is the one that bites upgraders. `enterprise/scalability/worker-group` adds a warning this paragraph drops:

> `fallback` only applies when a matching Worker Queue exists but has no live worker right now. If no queue matches the tags at all, the task fails immediately regardless of the `fallback` value.

So the advice "set `fallback: WAIT` explicitly" does not restore 1.x behaviour for a task whose tags match no queue. This section is the upgrade warning, it should carry the caveat.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+## RBAC: Action-Based Permissions
+
+The CRUD permission model (READ, CREATE, UPDATE, DELETE on generic resources) is replaced by a resource-plus-action model. Each resource now exposes only the actions that make sense for it.
+
+A few examples of what this makes possible:
+
+| Goal | Old model (minimum viable) | New model |
+|---|---|---|
+| CI/CD service account deploys and runs flows | FLOW: CREATE + EXECUTION: CREATE | `FLOW: EXECUTE` |
```
wrussell1999: This row contradicts itself. The goal is "deploys and runs flows", the new-model column grants `FLOW: EXECUTE` alone, which cannot deploy anything.

Per the migration mapping, old `FLOW: CREATE` becomes `CREATE` + `IMPORT`, and `EXECUTE` comes out of old `FLOW: UPDATE`. The minimum for this goal is closer to `FLOW: CREATE, UPDATE, EXECUTE`.

It's the most persuasive example in the section, which is why it needs to be right.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+        value: io.kestra.plugin.scripts
+    property: containerImage
+    regex: "^registry.internal/.*"
+    action: block
+    errorMessage: "Container images must be pulled from registry.internal."
+```
+
+`Add` rules inject values at resolution time. With `override: false` (the default), the author's explicit value wins and the policy fills in only what's absent. With `override: true`, the policy value always wins. Either way, every injection is annotated in the flow editor's merged preview, so forced values are never invisible to authors.
+
+Before enabling enforcement, set `enforcement: EVALUATE`. The policy checks every flow in scope and surfaces violations in the Governance UI, but nothing is blocked and no values are injected. When the violation report looks right, flip to `ACTIVE`.
```
wrussell1999: Overstates what `EVALUATE` does. The Policies reference is explicit that mutate rules are suppressed in this mode:

> Mutate rules (`Add`, `Delete`) are suppressed, no values are injected or removed, and the Governance UI shows no violations for them. [...] it is not useful for auditing mutate rule coverage.

As written ("checks every flow in scope and surfaces violations", "when the violation report looks right, flip to `ACTIVE`") a reader will take an empty report as confirmation their `Add` rules are safe to enable. Add the carve-out.

Same shape of gap at L206: validate rules (`Deny`, `Restrict`, `Require`) are not enforced in `REFERENCE` mode, which is worth one clause since the paragraph is selling reference bundles.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+Five managed roles ship with 2.0: Viewer, Launcher, Editor, Developer, and Admin. Existing custom roles and bindings are migrated automatically on upgrade.
+
+See the [RBAC reference](/docs/enterprise/auth/rbac) and the [migration guide for the RBAC action model](/docs/migration-guide/v2.0.0/rbac-action-model).
+
+## Policies
+
+Without enforcement tooling, keeping flows compliant across many namespaces is a manual coordination problem: authors must set values correctly on every task, and administrators have no way to verify or block non-compliant flows. Policies solves this at the platform layer.
+
+Policies is the EE replacement for `pluginDefaults`. It gives platform administrators governance rules that inject configuration, validate compliance, and block non-conforming flows across namespaces, tenants, and flow-level properties that `pluginDefaults` could never reach, like `retry`, `concurrency`, and `labels`.
```
wrussell1999: Two consecutive paragraphs both introduce Policies. L168 sets up the problem and lands on "Policies solves this at the platform layer", then L170 restarts from the top with "Policies is the EE replacement for `pluginDefaults`". Merge them.

Broader structural note: the problem-then-solution opener repeats at L168, L214, L337, L365, L420 and L430. Six sections with the identical rhetorical shape reads as a filled-in template rather than writing. Some of these features can just be described.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+`pluginDefaults` is removed in 2.0 for both OSS and EE. The [migration guide](/docs/migration-guide/v2.0.0/plugin-defaults-removed) covers all three scopes (flow-level, namespace-level, and global server config) with before-and-after examples. See the [Policies reference](/docs/enterprise/governance/policies) for the full rule DSL.
+
+## Loop Task
+
+ForEach and ForEachItem are removed in 2.0. The `Loop` task replaces both.
+
+The removal was driven by a real stability problem. A `ForEach` task with a large input list could generate thousands of child task runs within a single flow execution, exhausting executor memory and affecting every other flow running on the instance at the same time. `Loop` runs each iteration as an isolated sub-execution. A runaway loop cannot destabilize the instance.
+
+The expression syntax is also cleaner. Where ForEach used `{{ taskrun.value }}` and ForEachItem used `{{ taskrun.value }}` with a `parentOutput()` helper for nested access, Loop uses:
```
wrussell1999: Two problems in one sentence.

It names the same expression for both tasks (`{{ taskrun.value }}` twice), so the contrast is empty.

And `parentOutput()` doesn't exist. It isn't in `foreach-loop`, or anywhere else in the 2.0 docs. The actual old mechanism for nested access was `parent.taskrun.value` / `parents[0].taskrun.value`. Drop it.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+| Old expression | New expression |
+|---|---|
+| `{{ taskrun.value }}` | `{{ item.value }}` |
+| `{{ taskrun.iteration }}` | `{{ item.index }}` |
+| `{{ parent.taskrun.value }}` | `{{ item.parent.value }}` |
```
wrussell1999: Wrong mapping. `foreach-loop` gives `parent.taskrun.value` to `item.value`, no prefix, because `item` is bound to the sub-execution and reachable from any depth inside a Loop.

`item.parent.value` is the mapping for `parents[0].taskrun.value`, and only inside the inner Loop of two nested Loops.

Anyone copying this row into a single Loop with a nested `If` gets a null.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+triggers:
+  - id: weekly
+    type: io.kestra.plugin.core.trigger.Schedule
+    cron: "0 9 * * 1"
+    when: "{{ isWeekend(trigger.date, 'Europe/Paris') == false and isPublicHoliday(trigger.date, 'FR') == false }}"
+```
+
+New date and calendar helper functions ship alongside the redesign: `isWeekend()`, `isPublicHoliday()` (with country and subdivision), `isDayWeekInMonth()`, `isLastWorkingDay()`, `dayOfWeek()`, `dayOfMonth()`, `monthOfYear()`, and `hourOfDay()`.
+
+Flow triggers also change significantly. The old `conditions` and `preconditions` system is replaced by a `dependsOn` list of upstream flow entries, each specifying `flowId`, `namespace`, `states` (default: SUCCESS and WARNING), `labels`, and an optional per-entry `when`. A top-level `when` on the Flow trigger evaluates before the `dependsOn` check and can use `trigger.namespace`, `trigger.flowId`, `trigger.state`, and `trigger.outputs`.
```
wrussell1999: `states` default is wrong, and wrong in the direction that causes incidents. From `trigger-conditions-redesign`:

> When no `states` are specified on a `dependsOn` entry, the trigger evaluates against all terminal states (`SUCCESS`, `WARNING`, `FAILED`, `KILLED`, `CANCELLED`, `RETRIED`, `SKIPPED`, `RESUBMITTED`) and `PAUSED`.

"default: SUCCESS and WARNING" tells readers a failed upstream run won't fire the downstream trigger. It will.

Also, `window` moved to the trigger level in 2.0 and is missing from this paragraph, even though the breaking-changes table at L481 references it.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+kestractl roles create \
+  --name ci-role \
+  --permission FLOW:READ,CREATE,UPDATE \
+  --permission EXECUTION:CREATE,READ
+
+# Bind the role to the service account
+kestractl roles bind \
+  --role ci-role \
+  --service-account ci-deployer
+```
```
wrussell1999: Fix this one first. The example uses old CRUD action names in a post whose thesis is that CRUD is gone, and the migration guide carries a warning about exactly this case:

> `READ` is still accepted as a legacy alias and maps to `VIEW` only, it does **not** expand to `VIEW` + `LIST` + `ACCESS_LOGS` [...] Using old CRUD values in new role definitions will silently under-grant permissions.

So a reader who copies this block gets a CI service account that can't list flows or read logs, with no error to explain why. `--permission FLOW:READ,CREATE,UPDATE` also mixes both models in one line, since `CREATE` and `UPDATE` are real 2.0 actions and `READ` is a legacy alias.

The migration guide's own example is the shape to copy: `--permission FLOW:VIEW,LIST,EXECUTE,DISABLE,ENABLE`.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+The pattern mirrors the `PushFlows`/`SyncFlows` GitOps model. Platform teams can manage the approved blueprint library centrally, review changes via pull requests, and deploy consistently across multiple Kestra instances or environments.
+
+This complements Templated Blueprints (also new in 2.0), which let platform teams author blueprints with Pebble-based form templates. Users fill a form; Kestra generates the flow YAML. Combined with version control, the governance loop closes: templates are authored in code, reviewed in Git, and distributed as a self-service library.
+
+See the [Custom Blueprints reference](/docs/enterprise/governance/custom-blueprints).
+
+## Cases
+
+Failed executions are incidents, and most teams track them in a separate tool. Cases brings incident management into Kestra so you can create, assign, and resolve incidents next to the executions that caused them.
```
wrussell1999: "most teams track them in a separate tool" is an unsourced quantity claim. Either name them (PagerDuty, Opsgenie, Jira) or drop the quantifier. "Failed executions are incidents, and they usually get tracked outside Kestra" makes the same point without inventing a statistic.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+The `CreateCase` task opens a case from any block in a flow: `errors`, `finally`, `afterExecution`, or a regular task combined with `runIf`. It calls the Kestra API, so it needs an endpoint and credentials — `kestraUrl` defaults to the current instance, and `auth` accepts an API token or username/password. Namespace or tenant-level default credentials work as a fallback so you don't have to repeat auth config on every task.
```
wrussell1999: Em dash. Comma or parentheses instead: "...so it needs an endpoint and credentials. `kestraUrl` defaults to the current instance, and `auth` accepts..."

The other one is at L470.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+      resolution: PT8H
+```
+
+The `linkMatchingExecutions` property is the most useful option for high-frequency flows. A single external API going down can generate dozens of failed executions per hour. With `linkMatchingExecutions: true`, each subsequent failure attaches to the already-open case rather than creating a new one. The same behavior is available from the UI on any existing case via auto-attach, which generates a Flow trigger behind the scenes and removes it when the case resolves.
+
+Each case tracks severity, status, assignees, watchers, SLA targets (acknowledgement and resolution), linked executions, linked assets, and a full activity timeline with comments and file attachments. Cases can also have case actions: flows attached as one-click remediation buttons on the case detail page.
+
+The Cases board view and list view sit in the left menu. The board groups cards by status, severity, or assignee with a live SLA countdown per card. Dragging a card to Resolved opens the resolve modal, where a resolution reason is required.
+
+A new `CASE` RBAC resource controls access with actions for `VIEW`, `LIST`, `CREATE`, `UPDATE`, `DELETE`, `FOLLOW`, and `TEMPLATE`.
```
wrussell1999: Check `TEMPLATE` against the RBAC guide, which states that `TEMPLATE` permissions are silently dropped on upgrade because there's no equivalent resource in 2.0. If `CASE: TEMPLATE` is a genuinely distinct action for case templates then this is fine, but the two statements contradict each other on the page. The Cases reference only exercises `CASE: UPDATE` and `FLOW: EXECUTE`.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+![Deploy tab showing a source-to-target diff with the target selector and Promote button](./promote-deploy-tab.png)
+
+See the [Promote reference](/docs/enterprise/governance/promote).
+
+## AWS EC2 Task Runner
+
+<!-- TODO: optional screenshot or flow snippet for GPU Spot example -->
+
+The EC2 task runner is a new EE runner type that executes task commands directly on an EC2 instance via [AWS Systems Manager Run Command](https://docs.aws.amazon.com/systems-manager/latest/userguide/execute-remote-commands.html). No SSH. No container runtime.
```
wrussell1999: "No SSH. No container runtime." is the negated-countdown pattern. Fold it into the sentence above: "...via AWS Systems Manager Run Command, so no SSH access or container runtime is required."

Same fragment habit at L369 ("No opening each instance separately to check.").

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+See the [AWS EC2 Task Runner reference](/docs/task-runners/04.types/05.aws-ec2-task-runner) for IAM permissions, AMI requirements, and setup.
+
+## PurgeStorage
+
+The existing `PurgeExecutions` task deletes execution records and their associated files, but it is database-driven: it cannot clean files whose execution records are already gone. This becomes a problem in deployments where worker groups use isolated internal storage and files accumulate without any corresponding execution record to trigger a cleanup.
+
+`io.kestra.plugin.core.storage.PurgeStorage` takes the storage-driven approach instead. It walks the storage tree directly and deletes files based on last-modified date, regardless of whether a matching execution record exists. The task defaults to `dryRun: true`, so the first run only reports what would be deleted.
+
+The workerGroup property (inherited from all Kestra tasks) means PurgeStorage can target a specific worker's isolated storage by running on that worker group.
```
wrussell1999: Internal contradiction. L482 lists `workerGroup.key` as removed in 2.0 and directs readers to `workerSelector.tags`, but this sentence depends on "the workerGroup property (inherited from all Kestra tasks)".

The purge doc on the base branch still shows `workerGroup:` / `key: my-worker-group`, so either the removal at L482 is broader than reality or that doc is stale. Worth settling with engineering, then fixing whichever side is wrong. If the property is gone, this should say `workerSelector.tags`.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+
+AI Agent observability: AI Agents emit Prometheus metrics for tool calls, provider calls, and embedding store calls (`ai.agent.tool.calls`, `ai.provider.calls`, `ai.embedding.store.calls`). New MCP client tasks (`SseMcpClient`, `StdioMcpClient`, `DockerMcpClient`, `StreamableHttpMcpClient`) let Agent tasks call external MCP servers as tools.
+
+mTLS on the worker channel: Worker-to-Executor communication can be secured with mutual TLS. Configure a certificate authority, a server certificate for the Kestra server, and a client certificate for each worker. Workers that cannot present a valid client certificate are rejected at the TLS handshake before reaching the application layer. See the [gRPC TLS/mTLS configuration reference](/docs/configuration/enterprise-and-advanced#grpc-tlsmtls-ee-only).
+
+Draft revisions: save any flow change as a draft from the flow editor without affecting live executions. A draft revision is never executed — any trigger or manual run falls back to the last published revision. A warning banner in the run panel shows a Publish button when the latest revision is a draft. See [Revisions](/docs/concepts/revision).
```
wrussell1999: Nine consecutive paragraphs shaped "Label: sentence. Sentence." This is a bullet list in prose clothing, the same tell as bold-first bullets.

Either make it a real list, or rewrite each item so it doesn't open by naming itself. Nine is also a lot for a catch-all, and mTLS on the worker channel plus Draft revisions are substantial enough to earn their own H2 rather than being buried at the bottom.

Em dash on this line as well.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,496 @@
...
+triggers:
+  - id: mcp
+    type: io.kestra.plugin.core.trigger.McpToolTrigger
+    toolName: run-dbt-model
+    title: Run a dbt model
+    toolDescription: >
+      Runs a specific dbt model in the production warehouse. Use when a user asks
+      to refresh a specific model or rebuild a table. Input: the model name as it
+      appears in the dbt project (e.g. "stg_orders", "fct_revenue").
+```
```
wrussell1999, line 133: Does this need a flow level outputs block?

I have an non working example below but I am not sure.

```suggestion
```yaml
id: run_dbt_model
namespace: company.data

tasks:
  - id: dbt
    type: io.kestra.plugin.dbt.cli.DbtCLI
    commands:
      - dbt run --select {{ inputs.model_name }}

outputs:
  - id: output
    type: STRING
    value: "{{ outputs.dbt.vars..... }}"

triggers:
  - id: mcp
    type: io.kestra.plugin.core.trigger.McpToolTrigger
    toolName: run-dbt-model
    title: Run a dbt model
    toolDescription: >
      Runs a specific dbt model in the production warehouse. Use when a user asks
      to refresh a specific model or rebuild a table. Input: the model name as it
      appears in the dbt project (e.g. "stg_orders", "fct_revenue").
```
```

**Review by wrussell1999 — COMMENTED** [review]
Mixture of my own and the CLAUDE.md

File: src/contents/blogs/release-2-0/index.md (hunk 3 above)
Comment (aj-emerich) [reply src/contents/blogs/release-2-0/index.md]: You can safely ignore these lead paragraphs. I have something different locally.

File: src/contents/blogs/release-2-0/index.md (hunk 4 above)
Comment (aj-emerich) [reply src/contents/blogs/release-2-0/index.md]: Similar to above.

File: src/contents/blogs/release-2-0/index.md [hunk 25]
```diff
@@ -0,0 +1,469 @@
...
+
+The `toolDescription` field is the most important property to get right. Agents use it to decide when and how to invoke the tool, so a vague description produces poor routing. Write it from the agent's perspective: what situation should trigger this call, and what does the input represent.
+
+Flow inputs map automatically to the tool's JSON schema parameter spec. Outputs become the tool's response payload. If a flow has a `JSON` input with a `jsonSchema` property, the schema propagates to the tool spec.
+
+A `default` MCP server is provisioned for every tenant on startup. Additional servers (separate servers per team, or one per environment) can be created from the UI. Each server generates ready-to-paste connection configuration for Claude Desktop, Claude Code, Cursor, and Codex on its Connect tab.
+
+The MCP server also works in the other direction. Connect Claude, Cursor, or any MCP-compatible client to the Kestra MCP server and you can create, search, and manage flows directly from your editor or AI assistant, no UI required.
```
tchiotludo: > and you can create, search, and manage flows directly 

This is wrong, you can interact with tools, but not create flows

aj-emerich:
```suggestion
The MCP server also works in the other direction. Connect Claude, Cursor, or any MCP-compatible client to the Kestra MCP server and you can interact with flows as tools directly from your editor or AI assistant, no UI required.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
+---
+title: "Kestra 2.0: MCP Tools, Enterprise Governance, and Worker Groups"
```
anna-geller:
```suggestion
title: "Kestra 2.0: A New Engine, Workers Anywhere, and Flows as Agent Tools"
```

Reasons:
1. MCP is just a protocol and not everybody is fond of it, putting it as nr 1 2.0 feature seems like too much
2. Enterprise governance and worker groups are features we already have for a long time. We have audit logs, RBAC, fine-grained user permissions, all of that. Have enterprise governance features and worker groups existed since release 0.11 or so, I think?

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+description: "Kestra 2.0 ships enterprise governance for the first time (Policies, Cases, Promote), makes flows callable by AI agents via MCP, rebuilds Worker Groups with tag-based routing and JWT auth, and replaces accumulated legacy constructs with cleaner primitives."
```
anna-geller:
```suggestion
description: "Flow configuration, incidents and promotion between environments can now be governed with Policies, Cases and Promote. Flows can be saved as Drafts and published when ready, and the AI Copilot is now a persistent chat panel with memory."
```

We don't introduce enterprise governance for the first time. Kestra had enterprise governance since day one, we just make it stronger now with new features. Subjective but IMO we also shouldn't repeat what is already stated in the title in the subtitle, it makes things feel repetitive and boring.

"multi-turn" memory seems like a technical implementation detail - agree?

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+That foundation is what makes the rest of this release possible. AI agents can now invoke flows directly as MCP tools — Kestra orchestrates your AI stack, not the other way around. Enterprise governance ships for the first time in Policies, Cases, and Promote. Worker Groups are rebuilt with tag-based routing, capacity reservation, and JWT authentication. The permission model moves from CRUD to a resource-plus-action model. The editor gains a No Code canvas and a persistent Copilot sidebar with multi-turn memory. Loop replaces ForEach with isolated sub-executions, and the trigger `when` expression replaces Java class conditions across all trigger types. The feature table covers edition availability for everything.
```
anna-geller:
```suggestion
Kestra 2.0 is available today. Until now, network access to the Kestra database was required for every worker, so workers could only be deployed on the same network as the control plane. The queue was coupled to the database as well, so every feature had to be built twice, once for JDBC and once for Kafka.

In 2.0, each worker is connected to the control plane over a single outbound gRPC stream, so it can be deployed in another region, in a different cloud, or in a network that only allows outbound connections. The queue is independent of the database now, so there is a single implementation for both backends.

Everything else in 2.0 is built on those two changes. In this post, we cover what's new and what has to be changed when upgrading. The table below lists features we introduced in each edition.
```

Big claim sentence followed by an explanation of the technical implementation and listing of features. I think we can make the intro a more high-level value explanation/story - agree?

TBD if the first sentence is not too dry. We could make it warmer like "We're excited to announce that Kestra 2.0 is available today", but it's the most templated opening in software release posts so IMO worth cutting to the chase instead

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+| Feature | What | Edition |
+|---|---|---|
+| MCP Tool Trigger + MCP Server | Expose flows as tools callable by AI agents | EE, Cloud |
```
anna-geller:
```suggestion
| MCP Tool Trigger + MCP Server | Flows exposed as tools callable by AI agents | EE, Cloud |
```

We should make this table consistent. Either we start each row with a verb/action or none of them.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| AI Copilot agentic loop | Three-mode chat sidebar (Edit, Plan, Ask) with multi-turn memory and confirmation gate | EE, Cloud |
```
anna-geller:
```suggestion
| AI Copilot agentic loop | Three-mode chat sidebar (Edit, Plan, Ask) with multi-turn memory and a confirmation step | EE, Cloud |
```

gate is such a strange-sounding Claude-gen word, I'd ban it unless talking about feature gating :)

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| No Code canvas editor | Visual canvas-based flow editor with guided forms, synced with YAML and Copilot | OSS, EE, Cloud |
```
anna-geller:
```suggestion
| No Code canvas editor | Visual canvas-based flow editor with guided forms, in sync with YAML and AI Copilot | OSS, EE, Cloud |
```
Never say just Copilot. It's AI Copilot. These days, even Copilot is a bit controversial, as the feature is more than Copilot. It's more like an AI chat experience. But don't worry, we keep the name AI Copilot. We will not rename it :)

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| RBAC action-based permissions | Resource + action model replaces CRUD | EE |
```
anna-geller:
```suggestion
| RBAC action-based permissions | Resource plus action model in place of CRUD | EE |
```
The same as the first row: either we make it verb/action-oriented in all rows or in none of them.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Policies | Namespace-scoped governance rules replace `pluginDefaults` | EE |
```
anna-geller:
```suggestion
| Policies | Namespace-scoped governance rules in place of `pluginDefaults` | EE |
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Cases | Incident management for executions: create, deduplicate, and track to resolution without leaving Kestra | EE, Cloud |
+| Promote | Move flows across environments from the UI, with drift detection and a review gate | EE |
```
anna-geller:
```suggestion
| Promote | Move flows across environments from the UI, with drift detection and a review step | EE |
```
gate :)

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Blueprint version control | PushBlueprints and SyncBlueprints tasks for Git-based governance | EE |
+| kestractl IAM commands | Full IAM management (users, groups, roles, service accounts) from CLI | EE |
+| Worker Groups 2.0 | Tag-based routing, capacity reservation, JWT auth | EE |
+| New task runners | AWS EC2, Azure VM, Google Compute Engine, Huawei CCI | EE, Cloud |
+| Loop task | Replaces ForEach and ForEachItem with isolated sub-executions | OSS, EE, Cloud |
```
anna-geller:
```suggestion
| Loop task | In place of ForEach and ForEachItem, with isolated sub-executions | OSS, EE, Cloud |
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Trigger `when` expression | Pebble replaces Java-class conditions on all trigger types | OSS, EE, Cloud |
```
anna-geller:
```suggestion
| Trigger `when` expression | Pebble expression in place of Java-class conditions on all trigger types | OSS, EE, Cloud |
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| PurgeStorage | Storage-driven cleanup for orphaned execution files | OSS, EE, Cloud |
+| External Log Data Store | Route execution logs to a separate JDBC database or Elasticsearch | OSS (JDBC), EE (Elasticsearch) |
+| Reusable Inputs | Shared input groups defined once at namespace level | EE, Cloud |
+| Slim image + plugin auto-install | `kestra/kestra:*-slim` ships without bundled plugins; set `KESTRA_PLUGINS_AUTO_INSTALL_ENABLED=true` to install what's needed on first use | OSS |
```
anna-geller:
```suggestion
| Slim image + plugin auto-install | `kestra/kestra:*-slim` has no bundled plugins; with `KESTRA_PLUGINS_AUTO_INSTALL_ENABLED=true`, whatever a flow references is installed on first use | OSS |
```
More natural, I think + no action in all rows consistently

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Plugin artifacts | Plugins can ship Vue.js UI components that load into the Kestra execution topology at runtime | OSS, EE, Cloud |
```
anna-geller:
```suggestion
| Plugin artifacts | Vue.js UI components bundled in a plugin and loaded into the execution topology at runtime | OSS, EE, Cloud |
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+:::alert{type="info"}
+**Upgrading from 1.x?** You must be on Kestra 1.3.x before upgrading. Several constructs are removed in 2.0 (ForEach, trigger conditions, `workerGroup.key`, `pluginDefaults`), but `kestra-migrate` handles most flow rewrites automatically. The [Upgrade and Migration](#upgrade-and-migration) section at the bottom covers the full checklist, and every breaking change has a dedicated guide in the [v2.0.0 migration hub](/docs/migration-guide/v2.0.0).
```
anna-geller:
```suggestion
**Upgrading from 1.x?** You must be on Kestra 1.3.x before upgrading. Several constructs are removed in 2.0 (ForEach, trigger conditions, `workerGroup.key`, `pluginDefaults`), but most flow rewrites are handled automatically by `kestra-migrate`. The full checklist is in [Upgrade and Migration](#upgrade-and-migration), and every breaking change has a dedicated guide in the [v2.0.0 migration hub](/docs/migration-guide/v2.0.0).
```
Subjective

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+:::
+
+## MCP Tool Trigger and MCP Server
+
+Any Kestra flow can now register as a named tool on an MCP server. An AI agent sends a tool call; Kestra creates an execution with the matched inputs, runs the flow, and returns the outputs. No custom API wrapper or polling loop required.
+
+The `McpToolTrigger` handles registration. Add it to any flow alongside your task list:
+
+```yaml
+id: run_dbt_model
```
anna-geller: As discussed via Slack, let's replace with a more generic example.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Four core tasks removed | `io.kestra.plugin.core.execution.Count`, `Resume`, `trigger.Toggle`, and `log.Fetch` are removed. Replace with their equivalents in the `plugin-kestra` SDK. |
+| `CANCELED` enum alias removed | Replace the single-L spelling with `CANCELLED` in flow expressions, API consumers, and any tooling that checks execution state. |
+| SDK auth required for internal tasks | Tasks that call the Kestra API internally (git sync tasks and others) now require explicit credentials. See the [migration guide](/docs/migration-guide/v2.0.0/sdk-authentication). |
+| Super Admin renamed to Instance Owner | The Super Admin role is renamed to Instance Owner across the UI, CLI, config, and API. HTTP API responses emit `instanceOwner` instead of `superAdmin`; update any consumers that read this field. See the [migration guide](/docs/migration-guide/v2.0.0/superadmin-renamed-instance-owner). |
+
+## Get Started
+
+2.0 is the foundation the next phase of Kestra is built on — and it's the first LTS release, with bug and security fixes backported for one year. Teams upgrading now won't face another major migration in that window.
```
anna-geller:
```suggestion
2.0 is the foundation the next phase of Kestra is built on, and it's an LTS release, with bug and security fixes backported for one year. Anyone upgrading now won't face another major migration in that window.
```
emdash controversies haha

+ kestra is for individuals too

aj-emerich: Sad to never be able to use an em dash again 🥲 - good point!

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Plan | Proposes a numbered sequence of steps for a complex task and executes each one after you confirm. Rejecting any step cancels the rest. |
+| Ask | Answers questions about Kestra grounded in the official documentation via an internal Kestra MCP client. Can also read execution logs directly to help diagnose a failed run. |
+
+![AI Copilot Edit mode sidebar showing the Copilot searching plugins and proposing a validated S3-to-Postgres flow with an Apply button](./ai-copilot-edit-mode.png)
+
+When you open the sidebar while viewing a resource, that resource attaches automatically as a context tag above the input. Context tags are independently dismissible. Every addition and removal is recorded in the transcript so you can always see what the agent is looking at. Attachable resources include flows, namespaces, executions, dashboards, apps, test suites, blueprints, and plugins. The Copilot also reads namespace metadata (Policies, Variables, Secrets, Key-Value pairs) to ground authoring suggestions against your actual configuration, so prompts like "create a task that reads from our MongoDB" can reuse configured credentials without extra hints.
```
anna-geller:
```suggestion
When you open the sidebar while viewing a resource, that resource attaches automatically as a context tag above the input. Context tags are independently dismissible. Every addition and removal is recorded in the transcript so you can always see what the agent is looking at. Attachable resources include flows, namespaces, executions, dashboards, apps, test suites, blueprints, and plugins. The AI Copilot also reads namespace metadata (Policies, Variables, Secrets, Key-Value pairs) to ground authoring suggestions against your actual configuration, so prompts like "create a task that reads from our MongoDB" can reuse configured credentials without extra hints.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+Actions that modify resources require explicit confirmation before the Copilot executes them. A prompt appears in the chat with an optional field to steer the next attempt. Approving applies the change; rejecting resumes the conversation in Edit mode, or cancels the current plan in Plan mode.
```
anna-geller:
```suggestion
Actions that modify resources require explicit confirmation before the AI Copilot executes them. A prompt appears in the chat with an optional field to steer the next attempt. Approving applies the change; rejecting resumes the conversation in Edit mode, or cancels the current plan in Plan mode.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+Without enforcement tooling, keeping flows compliant across many namespaces is a manual coordination problem: authors must set values correctly on every task, and administrators have no way to verify or block non-compliant flows. Policies address this at the platform layer, replacing `pluginDefaults` in EE with governance rules that inject configuration, validate compliance, and block non-conforming flows, including flow-level properties that `pluginDefaults` could never reach, like `retry`, `concurrency`, and `labels`.
+
+A Policy is a named set of rules scoped to a namespace or a tenant. Rules from a parent namespace cascade to all child namespaces automatically, so a company-wide constraint placed at the root namespace reaches every team without per-namespace configuration.
+
+Five rule types ship in 2.0: `Add` and `Delete` mutate configuration before execution without altering stored flow YAML; `Deny`, `Restrict`, and `Require` validate it and can block or warn when a flow violates a constraint. Rules target either the flow (`on: FLOW`) or any plugin instance in it (tasks, triggers, task runners) (`on: PLUGIN`), narrowed by a `where` clause that matches on the plugin type.
```
anna-geller:
```suggestion
Five rule types are available in 2.0: `Add` and `Delete` mutate configuration before execution without altering stored flow YAML; `Deny`, `Restrict`, and `Require` validate it and can block or warn when a flow violates a constraint. Rules target either the flow (`on: FLOW`) or any plugin instance in it (tasks, triggers, task runners) (`on: PLUGIN`), narrowed by a `where` clause that matches on the plugin type.
```
ship sounds unnatural

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+A practical example: require that every flow carries a team label, and restrict all script tasks to an approved container registry.
```
anna-geller:
```suggestion
A practical example: require that every flow declares a team label, and restrict all script tasks to an approved container registry.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+    severity: HIGH
+    linkMatchingExecutions: true
+    sla:
+      acknowledgement: PT1H
+      resolution: PT8H
+```
+
+The `linkMatchingExecutions` property is the most useful option for high-frequency flows. A single external API going down can generate dozens of failed executions per hour. With `linkMatchingExecutions: true`, each subsequent failure attaches to the already-open case rather than creating a new one. The same behavior is available from the UI on any existing case via auto-attach, which generates a Flow trigger behind the scenes and removes it when the case resolves.
+
+Each case tracks status, severity, assignees, SLA timers, and linked executions, with a full activity timeline. Cases can also carry case actions: flows attached as one-click remediation buttons on the case detail page.
```
anna-geller:
```suggestion
Each case tracks status, severity, assignees, SLA timers, and linked executions, with a full activity timeline. Cases can also have case actions: flows attached as one-click remediation buttons on the case detail page.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+The Cases board view and list view sit in the left menu. The board groups cards by status, severity, or assignee with a live SLA countdown per card. Dragging a card to Resolved opens the resolve modal, where a resolution reason is required.
+
+See the [Cases reference](/docs/enterprise/governance/cases).
+
+## Promote
+
+Moving a flow from dev to prod has never been a first-class action in Kestra. The usual workarounds are fragile (copy-paste YAML between instances, which silently drifts) or require building a Git pipeline yourself, which blocks anyone who isn't fluent in CI/CD setup. Promote adds a dedicated path for moving flows across environments directly from the UI, with a review step before anything lands in production.
```
anna-geller:
```suggestion
Moving a flow from dev to prod has never been a first-class action in Kestra. The usual workarounds are fragile (copy-paste YAML between instances, which silently drifts) or require building a Git pipeline yourself, which blocks anyone who isn't fluent in CI/CD setup. Promote adds a dedicated path for moving flows across environments directly from the UI, with a review step before anything is deployed to production.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+Each flow gains a Deploy tab alongside the editor. From there, select a target environment, review a diff of exactly what changes in that revision, and confirm. Gated targets (typically production) require explicit confirmation before the promotion runs. Every promotion is recorded in full: what moved, which revision, where it went, who confirmed it, and when.
```
anna-geller:
```suggestion
Each flow gains a Deploy tab alongside the editor. From there, select a target environment, review a diff of exactly what changes in that revision, and confirm. Protected targets (typically production) require explicit confirmation before the promotion runs. Every promotion is recorded in full: what moved, which revision, where it went, who confirmed it, and when.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Declarative topology bootstrap
+
+A `kestra.ee.setup` configuration block lets you declare the full worker topology (queues, groups, subscriptions, and registration tokens) in `application.yml`. Kestra provisions everything at startup, so there is no need for a second deployment pass or an init Job against a running webserver. Workers retry registration until their token is known to the controller, so all services can start concurrently. The semantics are create-if-not-exists: an existing entity is skipped on restart, so the database stays the source of truth once an entity exists.
+
+See the [Worker Groups reference](/docs/enterprise/scalability/worker-group) for migration steps, the `workerGroup.key` to `workerSelector.tags` mapping, and full [declarative configuration](/docs/enterprise/scalability/worker-group#declarative-configuration) details.
+
+## New Task Runners
+
+Four new EE task runners ship in 2.0, each targeting workloads that cannot or should not run in a container: GPU training tied to a custom AMI, licensed software bound to a specific machine image, or workloads where direct VM control matters.
```
anna-geller:
```suggestion
Four new EE task runners arrive in 2.0, each targeting workloads that cannot or should not run in a container: GPU training tied to a custom AMI, licensed software bound to a specific machine image, or workloads where direct VM control matters.
```

Too much "ship"

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+- [Azure Virtual Machine Task Runner](/docs/task-runners/types/azure-virtualmachine-task-runner): runs commands on Azure VMs via the Azure Run Command API, with no SSH and no public IP required.
+- [Google Compute Engine Task Runner](/docs/task-runners/types/google-computeengine-task-runner): runs commands directly on a Compute Engine VM as a startup script, with no SSH or IAP tunnel.
+- [Huawei Cloud CCI Task Runner](/docs/task-runners/types/huawei-cci-task-runner): runs tasks as bare Pods on Huawei Cloud CCI for serverless container execution.
+
+## Loop Task
+
+ForEach and ForEachItem are removed in 2.0. The `Loop` task replaces both.
+
+The removal was driven by a real stability problem. A `ForEach` task with a large input list could generate thousands of child task runs within a single flow execution, exhausting executor memory and affecting every other flow running on the instance at the same time. `Loop` runs each iteration as an isolated sub-execution. A runaway loop cannot destabilize the instance.
```
anna-geller:
```suggestion
The removal was driven by a stability problem. A `ForEach` task with a large input list could generate thousands of child task runs within a single flow execution, exhausting executor memory and affecting every other flow running on the instance at the same time. `Loop` runs each iteration as an isolated sub-execution. A runaway loop cannot destabilize the instance.
```

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+## No Code Editor
+
+The flow editor gains a No Code canvas view alongside the existing YAML editor. The canvas displays each flow section (Triggers, Tasks, Errors, Finally, After Execution) as a group of visual blocks. Clicking a block opens its configuration form in a side panel with two tabs: **Form** (guided fields with inline documentation) and **Source** (raw YAML for that block).
+
+The form's left panel lists every upstream task output and execution context variable available at that point in the flow, so you can reference them in Pebble expressions without leaving the form or switching to the YAML editor to look them up.
+
+Add blocks with **+ Add task** or **+ Add trigger**, or press `/` on the canvas to search and insert a plugin at the cursor position. All three views (YAML editor, No Code canvas, and AI Copilot) stay in sync. Start in any mode and switch freely; every change reflects instantly across all three.
+
+A new `FORM` input type groups related inputs as a multi-step wizard in the Execute modal. Each step is a labeled section in the inputs list, making complex trigger forms easier to fill out without presenting every field at once.
```
anna-geller: why describe all of it in text?

I recommend cutting almost all of this text -- I'd only introduce what has been changed in the no-code editor in a sentence or two, and show the rest in an Arcade or a GIF or a YouTube video

This is purely a visual feature, so describing it in a wall of text without actually seeing it is not helpful.

anna-geller: I'd also caution against using the word "canvas" because people may think of a canvas with a true drag-and-drop UX

anna-geller: why is FORM-type input included in a No-Code editor section? it's unrelated

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+
+When you open the sidebar while viewing a resource, that resource attaches automatically as a context tag above the input. Context tags are independently dismissible. Every addition and removal is recorded in the transcript so you can always see what the agent is looking at. Attachable resources include flows, namespaces, executions, dashboards, apps, test suites, blueprints, and plugins. The Copilot also reads namespace metadata (Policies, Variables, Secrets, Key-Value pairs) to ground authoring suggestions against your actual configuration, so prompts like "create a task that reads from our MongoDB" can reuse configured credentials without extra hints.
+
+Actions that modify resources require explicit confirmation before the Copilot executes them. A prompt appears in the chat with an optional field to steer the next attempt. Approving applies the change; rejecting resumes the conversation in Edit mode, or cancels the current plan in Plan mode.
+
+The `COPILOT` resource in the RBAC model controls who can use the feature. Assign `USE` at tenant or namespace scope via the Roles UI.
+
+See the [AI Copilot reference](/docs/ai-tools/ai-copilot).
+
+## No Code Editor
```
anna-geller:
```suggestion
## No-code Editor
```
The tab itself in the UI is called "No-code", we should use it consistently everywhere.

src/contents/blogs/release-2-0/index.md
aj-emerich: Good point - I can quickly make arcades to capture this.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,469 @@
...
+| Goal | Old model (minimum viable) | New model |
+|---|---|---|
+| CI/CD service account deploys and runs flows | FLOW: CREATE + EXECUTION: CREATE | `FLOW: CREATE, UPDATE, EXECUTE` |
+| Support engineer reads logs | EXECUTION: READ | `EXECUTION: ACCESS_LOGS` |
+| Scheduler triggers backfills | EXECUTION: CREATE | `TRIGGER: BACKFILL` |
+| Analyst follows a live execution | EXECUTION: READ | `EXECUTION: FOLLOW` |
+
+New resources in 2.0 include `TRIGGER` (previously part of `FLOW`), `SYSTEM_SETTINGS`, `TENANT_SETTINGS`, `COPILOT`, and `MCP_SERVER`.
+
+Five managed roles ship with 2.0: Viewer, Launcher, Editor, Developer, and Admin. Existing custom roles and bindings are migrated automatically on upgrade.
```
anna-geller:
```suggestion
In 2.0, RBAC permissions have been redesigned to give fine-grained control over what each user is allowed to do. Every resource used to have the same four permissions (CREATE, READ, UPDATE, DELETE) regardless of what could actually be done with them. There are now 159 permissions across the product, one per action.

Executions are the clearest example. Under CRUD, restarting an execution, killing a running one, replaying it and deleting it all had to be expressed as CREATE, UPDATE or DELETE, and it wasn't obvious which permission allowed which action. Killing an execution changes its state, so UPDATE seems right, but it also terminates the run, so DELETE seems just as plausible. Each of those actions is a separate permission now: `RESTART`, `REPLAY`, `KILL`, `PAUSE`, `RESUME`, `UNQUEUE`, `FORCE_RUN`, `CHANGE_LABELS`, `UPDATE`, `DELETE` and `EXPORT`, alongside read-only ones like `VIEW`, `LIST`, `FOLLOW`, `ACCESS_LOGS`, `ACCESS_OUTPUTS` and `ACCESS_FILES`. Whoever is on call can be allowed to kill a runaway execution without being able to delete it, while a data engineer can restart or replay a failed run without being able to stop a running one.

Every other resource works the same way, with its own action list. Starting a flow, for instance, is `FLOW: EXECUTE`, kept separate from anything that happens to the execution afterwards.

New resources in 2.0 include `TRIGGER` (previously part of `FLOW`), `SYSTEM_SETTINGS`, `TENANT_SETTINGS`, `COPILOT`, and `MCP_SERVER`.

Setting every permission by hand isn't the expected path. Five managed roles come with 2.0, available as presets when a role is created: Viewer, Launcher, Editor, Developer, and Admin. Existing custom roles and bindings are migrated automatically on upgrade.
```
This whole section was way too mechanic and technical, explaining what has been changed on a technical level without explaining why -- I feel like explaining this on one specific example like Executions makes this much clearer

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,519 @@
...
+    type: io.kestra.plugin.core.flow.Loop
+    values: "{{ inputs.file_list }}"
+    tasks:
+      - id: transform
+        type: io.kestra.plugin.scripts.python.Commands
+        commands:
+          - python transform.py --input "{{ item.value }}"
+        outputs:
+          - id: result_path
+            type: STRING
```
wrussell1999: This doesn't work on RC13. Just throws 38 errors.

aj-emerich: I know the issue and will fix.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,514 @@
...
+
+## No-code Editor
+
+For teams who prefer building flows through forms rather than YAML, the No-code editor has been polished significantly in 2.0. A contextual data panel now sits alongside every configuration form, listing every input, upstream task output, and execution context variable available at that point in the flow, organized by category and filterable. A demo below says more than a thousand words, and all three views (YAML editor, No-code editor, and AI Copilot) stay in sync throughout the editing process.
+
+<div style="position: relative; padding-bottom: calc(48.8542% + 41px); height: 0px; width: 100%;"><iframe src="https://demo.arcade.software/GTFXessaiDkaI6hw0mof?embed&embed_mobile=tab&embed_desktop=inline&show_copy_link=true" title="2.0 No-code Editor - Kestra" frameborder="0" loading="lazy" webkitallowfullscreen mozallowfullscreen allowfullscreen allow="clipboard-write; autoplay" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; color-scheme: light;" ></iframe></div>
+
+The [flow editor docs](/docs/ui/flows) cover both No-code and YAML editing in full.
+
```
MartinRst, line 184: Suggest a short section between No-code Editor:

- The rebuilt navigation and execution views: less visual noise between authoring, monitoring and troubleshooting
- The redesigned topology, with animation driven by live execution state
- Inline plugin details in the topology (which is the natural lead-in to Plugin Artifacts right after)
- Per-phase timing in the execution view, so a slow run has an identifiable owner
- Purged output files now handled properly in the outputs panel

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,515 @@
...
+category: News & Product Updates
+authors:
+  - name: "AJ Emerich"
+    linkedin: https://www.linkedin.com/in/alex-emerich/
+    image: aemerich
+    role: Technical Writer
+image: ./main.jpg
+---
+
+Kestra 2.0 is out today. This release ships a new gRPC-based engine, resolves architectural constraints around worker deployments and backend coupling, and introduces new features across AI tooling, governance, infrastructure, and developer experience. If you haven't already, [Ludo's post on the 2.0 architecture](/blogs/kestra-2-0-engineering) goes deep on the engineering decisions that made this possible. We've put a lot into this one and can't wait to see what the community builds on it. In this post, we cover everything that's new and make sure you have what you need to migrate confidently. Here's what's new, broken down by edition.
```
MartinRst: Kestra 2.0 is out today. The engine is rewritten: workers no longer hold a database connection, the queue and the repository are independent choices, and there is one execution path instead of two. On that foundation, this release adds AI tooling, governance, new task runners, and a rebuilt authoring experience.

2.0 is also our first LTS release. Bug and security fixes are backported for a year, with no forced major upgrade in that window, so upgrading now means one migration instead of two.

[Ludo's post on the 2.0 architecture](https://github.com/blogs/kestra-2-0-engineering) goes deep on the engineering decisions that made this possible. This post covers what shipped, what each edition gets, and what you need to change to migrate.

MartinRst (changes requested):

src/contents/blogs/release-2-0/index.md
aj-emerich: new example

src/contents/blogs/release-2-0/index.md
aj-emerich: Not the first LTS release though since 1.0 was I believe the first.

anna-geller: "The engine is rewritten" - This is already overstated everywhere, including the landing page. let's keep AJ's current intro Martin 🙏

anna-geller:
```suggestion
Kestra 2.0 is available today. This release resolves architectural constraints around worker deployments and backend coupling, and introduces new features across AI tooling and governance. If you haven't already, check [Ludo's post on the 2.0 architecture](/blogs/kestra-2-0-engineering), which goes deep on the engineering decisions that made this possible. We've put a lot into this release and can't wait to see what the community builds on it. In this post, we cover everything that's new and make sure you have what you need to migrate confidently. Here's what's new, broken down by edition.
```
Only removing duplication:
- The new gRPC-based engine is the same as the removal of architectural constraints around worker deployments. We should cut one or the other. I think this way it makes sense. 
- Improving 'developer experience' is overly formulaic and can be said about every release of every devtool in the world. I think it's fair to cut it. Agree?
- For Ludo's post some verb is missing, like check, read. 
- Unclear antecedent in "we've put a lot into this".

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,515 @@
...
+| Slim image + plugin auto-install | `kestra/kestra:*-slim` has no bundled plugins; with `KESTRA_PLUGINS_AUTO_INSTALL_ENABLED=true`, whatever a flow references is installed on first use | OSS |
+| Plugin artifacts | Vue.js UI components bundled in a plugin and loaded into the execution topology at runtime | OSS, EE, Cloud |
+
+:::alert{type="info"}
+**Upgrading from 1.x?** You must be on Kestra 1.3.x before upgrading. Several constructs are removed in 2.0 (ForEach, trigger conditions, `workerGroup.key`, `pluginDefaults`), but most flow rewrites are handled automatically by `kestra-migrate`. The full checklist is in [Upgrade and Migration](#upgrade-and-migration), and every breaking change has a dedicated guide in the [v2.0.0 migration hub](/docs/migration-guide/v2.0.0).
+:::
+
+## MCP Tool Trigger and MCP Server
+
+Connecting an AI agent to real infrastructure usually means writing a custom integration layer. The MCP Tool Trigger skips it: any flow you've already built (a data pipeline, a provisioning sequence, an incident response) is callable by an AI agent as a named tool. No custom API, no polling loop.
```
anna-geller:
```suggestion
Connecting an AI agent to real infrastructure usually means writing a custom integration layer. The MCP Tool Trigger skips it: any flow you've already built (a data pipeline, a provisioning sequence, an incident response) is callable by an AI agent as a named tool.
```
Claude loves that: no X, no Y. I think it sounds too market-y. I'd cut it.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,515 @@
...
+
+The [Cases docs](/docs/enterprise/governance/cases) cover SLA configuration, case actions, and the auto-attach trigger setup.
+
+## Promote
+
+Before Promote, moving a flow between environments meant a CI/CD pipeline outside Kestra, a manual copy-paste of YAML, or both, with no drift visibility and no audit trail inside the platform. The goal, same as Cases, is to keep you in Kestra rather than context-switching to a separate CI/CD tool.
+
+Promote gives you a built-in path for moving flows between environments. Each flow gains a Promote tab alongside the editor: select a target, review a diff of exactly what changes in that revision, and confirm. Protected targets require explicit confirmation before anything lands in production. Every promotion is recorded in full: what moved, which revision, where it went, who confirmed it, and when. No Git pipeline required.
+
+The flows table gains a column showing drift at a glance. If production is running an older revision, the column shows out of sync. If a flow has never been promoted to that environment, it shows not promoted, so you never need to open each instance separately to check.
```
anna-geller:
```suggestion
The flows table includes a column showing drift at a glance. If production is running an older revision, the column shows out of sync. If a flow has never been promoted to that environment, it shows not promoted, so you never need to open each instance separately to check.
```
Otherwise, we have: 
- each flow gains
- the flows table gains

next to each other

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,515 @@
...
+
+![Promote tab showing a source-to-target diff with the target selector and Promote button](./promote-deploy-tab.png)
+
+The [Promote docs](/docs/enterprise/governance/promote) cover the confirmation process, drift detection, and promotion history.
+
+<div style="position: relative; padding-bottom: calc(49.0084% + 41px); height: 0px; width: 100%;"><iframe src="https://demo.arcade.software/5bsRGVNXSLuSoWETF8vM?embed&embed_mobile=tab&embed_desktop=inline&show_copy_link=true" title="Promote a Flow to a Different Environment - Kestra" frameborder="0" loading="lazy" webkitallowfullscreen mozallowfullscreen allowfullscreen allow="clipboard-write; autoplay" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; color-scheme: light;" ></iframe></div>
+
+## Blueprint Version Control
+
+Custom Blueprints can now be version-controlled with Git using two new EE tasks: `PushBlueprints` commits and pushes blueprints from Kestra to a Git repository, and `SyncBlueprints` pulls blueprints from Git into Kestra, treating Git as the single source of truth.
```
anna-geller:
```suggestion
Custom Blueprints can now be version-controlled with Git using two new EE tasks: 
1. `PushBlueprints` commits and pushes blueprints from Kestra to a Git repository, 
2. `SyncBlueprints` pulls blueprints from Git into Kestra, treating Git as the single source of truth.
```
Since you explicitly mention two tasks, I think enumerating them like this in a numbered list makes it clearer and gives a better flow.

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,515 @@
...
+kestractl, introduced in Kestra 1.3, gains full EE IAM management in 2.0. If you're provisioning Kestra programmatically (onboarding teams, managing service accounts in CI/CD, or scripting role assignments) you no longer need the UI. Every IAM entity is now reachable from the CLI:
+
+- `kestractl users`: create, list, get, update, delete users; set passwords
+- `kestractl groups`: manage groups and memberships (tenant-scoped)
+- `kestractl roles`: create and bind roles; assign permissions via `--permission TYPE:ACTION[,ACTION]` or `--permissions-file`
+- `kestractl service-accounts`: create and manage service accounts (instance-level)
+
+The `--output json` flag applies across all commands, so kestractl output can pipe directly into `jq` or other tooling in CI scripts.
+
+kestractl gained more commands in the 2.0 cycle than fit in a release post; the [kestractl reference](/docs/kestra-cli/kestractl) has every command with flags and authentication configuration.
```
anna-geller:
```suggestion
kestractl contains more commands in the 2.0 cycle than fit in a release post; the [kestractl reference](/docs/kestra-cli/kestractl) has every command with flags and authentication configuration.
```
I have a feeling that the word "gain" is overused a bit in the post, no?

src/contents/blogs/release-2-0/index.md
aj-emerich: @MartinRst I just added a section for this - can you have a look? Starts at line 171

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,527 @@
...
+
+## External Log Data Store
+
+Execution logs are the highest-volume data Kestra writes. In most production installations they dwarf flows and executions combined, yet they share the same database. Schema migrations pay a per-row cost across every table including logs. Backup and retention schedules apply uniformly when the operational reality is that logs and executions have different lifecycles.
+
+In 2.0 (EE), logs can be routed to a separate store using `kestra.logs.type`. Both JDBC (H2, PostgreSQL, MySQL) and Elasticsearch backends are supported. When configured, the main database handles only flows, executions, and state. Existing installations see no change on upgrade; historical logs written before the switch remain in the main database.
+
+The [External Log Data Store docs](/docs/administrator-guide/log-data-store) cover Elasticsearch config, aggregation and pagination behavior, and the plugin developer guide for custom backends.
+
+## Slim Image and Plugin Auto-Install
```
aj-emerich, line 457: I am going to add an arcade to show this in the product once the release finishes and there is an image to use.

src/contents/blogs/release-2-0/index.md
MartinRst: lgtm

Nico-Kestra:
Could you update the main visual please > https://www.figma.com/design/7YIFRlwzARHPMOAZTkRF7w/WEBSITE?node-id=29122-75939&t=r7Y2mHKafrANRYrR-1

MartinRst (approved):

wrussell1999 (approved):

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,543 @@
...
+
+## Plugin Artifacts
+
+Plugins can now ship Vue.js frontend components that load into the Kestra UI at runtime, without any changes to the core application. Three named slots let components render in the execution topology view, the task side drawer, or the task detail modal; when a plugin's task types appear in an execution, the matching component renders in place.
+
+Each component is compiled as a Module Federation micro-frontend using `@kestra-io/artifact-sdk` and bundled into the plugin JAR. At startup, Kestra discovers these bundles and makes them available to the host UI without static linking.
+
+Components call the Kestra API through `@kestra-io/kestra-sdk`, which provides typed methods for executions, flows, metrics, logs, and live progress events.
```
fdelbrayelle, line 477: @aj-emerich Should be removed since we removed the kestra-sdk dep

src/contents/docs/kestra-cli/kestractl/index.md
```diff
@@ -16,9 +16,13 @@ For server components and system maintenance commands (starting standalone serve
 Source code and releases are available at [github.com/kestra-io/kestractl](https://github.com/kestra-io/kestractl).
 
 ```bash
-curl -fsSL https://raw.githubusercontent.com/kestra-io/kestractl/main/install-scripts/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/kestra-io/kestractl/main/install-scripts/install.sh | VERSION=2 bash
```
fdelbrayelle, line 19: @aj-emerich @AcevedoR changed it and it's `curl -fsSL https://raw.githubusercontent.com/kestra-io/kestractl/main/install-scripts/install.sh | bash` 

For Kestra 1.x it's `curl -fsSL https://raw.githubusercontent.com/kestra-io/kestractl/main/install-scripts/install.sh | VERSION=1 bash`

src/contents/blogs/release-2-0/index.md
```diff
@@ -0,0 +1,543 @@
...
+
+A concrete example: a task runner that breaks execution into distinct infrastructure phases (scheduling, image pull, file transfer, task code) can render a `topology-details` component showing per-phase timing directly in the execution view, so slow executions have an identifiable owner. Without a plugin artifact, that timing data lives in raw log output.
+
+Plugin artifacts are available to all plugin authors in 2.0. The [plugin artifact developer guide](/docs/plugin-developer-guide/pluigin-ui) covers the SDK, slot registration, and bundling setup.
```
fdelbrayelle, line 481:
```suggestion
Plugin artifacts are available to all plugin authors in 2.0. The [plugin artifact developer guide](/docs/plugin-developer-guide/plugin-ui) covers the SDK, slot registration, and bundling setup.
```

PR #5185, opened 2026-07-20
refactor: cookie consent to support Google Consent Mode v2

elevatebart:
- Set Consent Mode v2 defaults (`denied` in EU) before GTM loads, updating signals on accept/decline/revoke.
- Detect visitor region at the Cloudflare edge into a `kestra_region` cookie, instead of a spoofable client-only Intl check.
- Code-split the EU-only consent banner so non-EU visitors never fetch it.
- Add `cookieconsent.test.ts` coverage and a CI fix to generate Cloudflare Worker types before `astro check`.
- Fix an unrelated pre-existing type-check failure blocking CI (`use-cases/index.astro`).

https://claude.ai/code/session_01BQC1wQwL7MTxy2gbSagqaq


vfanucci (commented):
Solid PR — clean architecture and genuinely good test coverage (consent `default` → `gtm.js` ordering, revoke via `onChange`, cookie-over-Intl precedence, idempotence across soft navigations). Nothing mechanically broken. Before merging I'd like to settle two non-code decisions, plus a couple of minor notes.

## 🔴 1. Unconditional `Set-Cookie` on every response × edge caching

`consentRegion.ts` appends `Set-Cookie` on **every** HTML response via a global middleware. Cloudflare's CDN does not cache responses that carry a `Set-Cookie` header by default, so this can go one of two ways:

- it **silently disables edge caching site-wide** — which would regress the TTFB→LCP edge-cache work in #5197; or
- if some cache layer keeps the response anyway, we hit exactly the leak the code comments warn about (*"one visitor's region cookie could leak to another"*).

The code flags the risk but adds no guard. Could we confirm whether the edge-cached routes (plugins/blueprints) pass through this middleware? If so, options would be: only set the cookie when it's absent, add `Cache-Control: private`, or scope the middleware away from cacheable routes.

## 🟠 2. No banner at all outside the EU (behavior change)

Previously the banner rendered for everyone (`opt-out` outside the EU). Now non-EU visitors get **no banner and `granted` by default** (`bannerModule = null`). That's intentional and good for performance, but it removes cookie controls for regimes like **CCPA/CPRA (California "Do Not Sell/Share")** and LGPD. This is a compliance call rather than a technical one — can someone who owns privacy sign off before we merge?

## 🟡 Minor (non-blocking)

- **`gtag` pushes an array, not the `arguments` object** (`dataLayer.push(_args)`). GTM reads Consent Mode positionally so this works in practice, but it diverges from Google's canonical snippet, and the tests assert the dataLayer *shape* rather than that GTM actually registers the consent. Worth a quick sanity check in GTM Preview.
- **`use-cases/index.astro`** uses `as any[]` / `as { results: any[] }` to unblock CI — fine as an out-of-scope fix, just noting it's an `any` bandaid.
- **Revoke doesn't tear down PostHog** — `onChange` relies on Consent Mode signals (correct for GTM), but an already-initialized PostHog keeps running after a revoke. Pre-existing behavior, not a regression here.

## ✅ What's great

- `default denied` in the EU **before** GTM loads, with `wait_for_update: 500` — textbook Consent Mode v2, keeps cookieless pings for conversion modeling.
- Edge-side, non-spoofable region detection with an EU fail-safe (the right GDPR default).
- `onChange`/revoke handling is new — a real compliance improvement.
- Code-splitting the banner is a real perf win for the non-EU majority.
- The `cf-typegen`-before-`astro check` CI fix is well explained.

Happy to approve once (1) the `Set-Cookie`/edge-cache interaction is confirmed and (2) the non-EU banner removal is signed off. Everything else is just commentary.

vfanucci:
Thanks @elevatebart — both follow-ups look right, and one of my two blockers turns out to have been my own misread. Details below, plus a fix I've pushed.

## ✅ Blocker 1 (`Set-Cookie`) — resolved, but it collided with `main`

Dropping the cookie for an `HTMLRewriter` attribute is the right call. The catch is that #5197 (edge caching for plugin/blueprint pages) merged on Jul 21 — one day after this branch's base — so the two changes never met until now. `injectConsentRegion` ran as a middleware, and in `worker.ts` on main:

- a cache **hit** returns before the middleware loop → no attribute at all, silent fallback to the client-side `Intl` check;
- a cache **miss** runs the chain and then stores *its output* → the first visitor's region is baked into the cached HTML for the whole 1h TTL and replayed to everyone else hitting that PoP.

The second one is the real problem: the cache key is the URL only, while `resolveRegion` reads `request.cf.timezone`, which is per-visitor and not per-PoP. If a `row` copy lands in the cache, EU visitors on that PoP get `granted` consent defaults and no banner until the TTL lapses — and since the attribute is read once at module top-level, the wrong state sticks for the whole SPA session. `/plugins/*` is one of our top organic entry points, so this isn't a rare path.

I've merged `main` and pushed the fix in d0813bf3: `injectConsentRegion` becomes `withConsentRegion(response, request)`, applied *outside* the cache boundary on both paths. The copy stored in the Workers cache is cloned before the rewrite, so it stays region-agnostic and keeps its `s-maxage`; every visitor gets their own region injected, hit or miss.

One change worth your eyes because it touches #5197's shipped behaviour: the copy returned to the visitor is now `Cache-Control: private, max-age=0, must-revalidate` instead of `public, s-maxage=…`. Once a per-visitor attribute is in the body, a shared proxy downstream honouring `s-maxage` would reproduce the same leak one hop further out. The stored copy keeps `s-maxage`, so the Workers cache TTL is untouched — only what we advertise downstream changes, and Cloudflare's own CDN is bypassed by `run_worker_first` anyway.

Alternatives I considered and rejected, in case you'd rather go another way:

- **region in the cache key** (`__region=eu|row`): smaller diff, but ×2 entries per URL → up to two renders per page per TTL, which eats into exactly what #5197 bought on the long tail of plugin pages;
- **skip injection on edge-cacheable routes** and fall back to `Intl` there: three lines, no cache impact, no regression vs main — but it gives up edge geo precisely where the traffic is, and the real value of edge geo isn't anti-spoofing, it's visitors whose browser reports UTC (Firefox `resistFingerprinting` & co.) being misclassified as non-EU;
- **client-side `/api/region` fetch**: adds a round-trip before `default denied` can be pushed, which defeats the point of Consent Mode v2.

`npm run test:unit` (60 passing), `wrangler types` + `astro check` (0 errors) and `oxlint` are all green locally; CI is re-running.

## ❌ Blocker 2 (no banner outside the EU) — retracted, my misread

I based it on `mode: isEurope ? "opt-in" : "opt-out"` and missed the early return above it:

```js
if (!isEurope) {
    enabledAnalytics()
    enabledMarketing()
    return          // runCookieConsent() is never reached
}
```

The `"opt-out"` branch is dead code on main, so non-EU visitors already get no banner today. This PR doesn't change that — it improves it: GPC is honoured, and the banner bundle is no longer downloaded for the non-EU majority. The CCPA/CPRA gap is real but pre-existing and not yours to close here; I'll open a separate issue for a "Do Not Sell or Share" entry point plus `showPreferences` reachable outside the EU. Worth noting GPC's practical reach is limited — Firefox (private windows, or the pref), Brave, DuckDuckGo; neither Chrome nor Safari sends it natively.

The GPC implementation itself reads well: treating it as a ceiling the banner can't relax (`marketingGranted`) is the right shape, and keeping `analytics_storage: granted` while denying the three ad signals matches CPRA's sale/share scope.

## Still open from my first pass (all non-blocking)

- `gtag` pushes an array rather than the `arguments` object — works positionally, but a GTM Preview sanity check would be reassuring, since the tests assert the dataLayer *shape* rather than that GTM actually registered the consent.
- `use-cases/index.astro`'s `as any[]` bandaid — fine as an out-of-scope CI unblock, just noting it.
- Revoke doesn't tear down an already-initialized PostHog (pre-existing, not a regression here).

Nothing blocking left from my side. Since I now have a commit in here, a second pair of eyes on d0813bf3 wouldn't hurt — if you're happy with it, say so and I'll approve.

vfanucci:
Rebased on `main` and verified end-to-end in GTM Preview. The Consent Mode signals are confirmed registered by GTM itself, which is the check that had been open since my first pass. Details, plus two findings that aren't this PR's to fix.

## Why this couldn't be verified before

`layout.astro` only injects `analytics.js`, `outbound.js` and `cookieconsent.ts` when `!import.meta.env.DEV && !PREVIEW`, and `deploy.yml` sets `PREVIEW=true` for every branch that isn't `main`. So **no preview deploy has ever contained the consent script** — on the preview URL `window.dataLayer` is `undefined`, GTM never loads, no banner. `astro dev` excludes the same scripts.

The verification below therefore comes from a local production build (`astro build` + `wrangler dev`, `PREVIEW` defaults to `false`), which ships the real container `GTM-T4F85WRF`. Worth knowing for anyone reviewing tracking changes here: enabling the scripts on previews isn't the fix — it would feed prod GTM/PostHog/HubSpot and fire real Ads conversions from every PR, and it still wouldn't reach the edge-cache path, which is gated on `url.host === "kestra.io"`, not on `PREVIEW`.

## Consent Mode v2, confirmed by GTM

Pre-consent, first container load — Tag Assistant **Consent** tab:

| Type | On-page Default | Current State |
|---|---|---|
| `ad_storage` | Denied | Denied |
| `analytics_storage` | Denied | Denied |
| `ad_user_data` | Denied | Denied |
| `ad_personalization` | Denied | Denied |

Event order in the timeline: `1 Consent Default` → `2 Consent Initialization` → `3 Initialization` → `4 Container Loaded`. The unit tests assert that ordering inside the `dataLayer` array; this is GTM confirming it in its own pipeline.

Transitions:
- **Accept all** → `Consent Update`, all four `Granted`, and `enable_marketing` fires *after* the update — so no marketing tag runs on a stale state.
- **Revoke via the preferences modal** → `Consent Update`, all four back to `Denied`. This is the `onChange` path the PR adds; previously a withdrawal left the signals `granted` until the next full page load.

## The `gtag`/`arguments` point from my first review

Fixed in ac4f81db1. It was a real defect, not style: GTM only treats a dataLayer entry as a gtag command when it is an `Arguments` object. An `Array` indexes identically — which is why the tests, all `def[0]`/`def[1]`/`def[2]`, passed either way — but fails that check, so `consent`/`default` never registers.

The failure mode isn't a missing conversion, it's a compliance regression: with no consent state recorded, GTM reverts to unrestricted mode and ad tags run with storage for visitors who accepted nothing. Today those visitors get no GTM at all, so we'd have gone from compliant to not.

Verified at runtime on the minified bundle: `Object.prototype.toString.call(dataLayer[0])` returns `[object Arguments]`, and `dataLayer.push(arguments)` survives minification intact. Added the type assertion the tests were missing so a regression to `push([...])` fails instead of passing.

## Worker side

- On the preview deploy (real Cloudflare edge): `<html data-kestra-region="eu">`, resolved from `request.cf.timezone`.
- `Cache-Control: private, max-age=0, must-revalidate` on the document.

The cache **hit** path still can't be exercised anywhere but production, since `canEdgeCache` requires `url.host === "kestra.io"`. Post-merge check is two `curl`s on `/plugins/core`: the attribute must be present on both calls, including the one with `cf-cache-status: HIT`.

## Conflict resolution

- `worker.ts`: `main` added `setupHtmlCacheControl` (#5196) to the chain and `VERSIONED_DOCS_PATH` to the imports. Kept both; `withConsentRegion` stays out of the chain. **One thing worth a reviewer's eyes:** it now re-sets the `Cache-Control` #5196 just wrote — `max-age=0, must-revalidate` survives (that's what #5196 needs for asset-hash freshness), `public` becomes `private`, because the body carries a per-visitor region attribute. Versioned docs joined `isEdgeCacheablePage()` meanwhile and get the same hit/miss handling for free.
- `astro-check.yml`: both sides inserted a step at the same spot (`cf-typegen` here, `test:unit` on `main`). Kept both, typegen first.

`npm run test:unit` 310 passing / 18 files, `wrangler types` + `astro check` 0 errors on 484 files, `oxlint` clean.

## CSP addition

cfabc49df adds `https://*.googletagmanager.com` to `connect-src`. Tag Assistant attaches its debug session with a `fetch` to `googletagmanager.com/gtag/js?id=DUMMY`, which was blocked — the container was not inspectable on any environment serving this CSP, production included. The domain is already trusted in `script-src`, `img-src` and `frame-src`; it may already execute arbitrary script on the page, so allowing it to be fetched grants strictly less than it holds today. Happy to split this out if you'd rather keep the PR narrow.

## Two findings, both out of scope

**No entry point to reopen the preferences modal.** The "Settings" button lives inside the consent banner, which is removed once a choice is made, and nothing in `src/` carries `data-cc="show-preferencesModal"`. So `onChange` — the compliance improvement this PR adds — is unreachable in production: a visitor can never withdraw consent. GDPR wants withdrawal as easy as consent, so this is a gap in itself, pre-existing and wider than the non-EU CCPA point I raised earlier. Fix is one line in the footer; folding it into the separate issue I mentioned.

**`gtm?.trackEvent()` is a permanent no-op.** The vue-gtm plugin is initialized with `enabled: false` in `vue-setup.ts` and nothing ever calls `gtm.enable(true)`. So `contact_form_submission`, `cloud_form`, `enterprise_trial_form`, `early_adopter_form`, `partner_form_submission`, `newsletter_form`, `ResourceCtaPair` and `identify` never reach the dataLayer, consent or not. Only `bookdemo_form` escapes it, because `Meeting.vue` pushes to `window.dataLayer` directly. Not a regression here, but it caps what any GTM-side conversion work can achieve until it's addressed.

Nothing blocking from my side. @elevatebart if you're happy with d0813bf3 and the conflict resolution, this is ready.

elevatebart:
We should merge this and if perf drops, revert it.

iitzIrFan (commented):
Let's ship it and verify consent signals in `Tag Assistant` then!

elevatebart:
... Still no review, ready whne @vfanucci approves. 
Do not merge yourself though. I would like to be there to monitor perf and revert if necessary.

vfanucci (approved):
Reviewed the Consent Mode v2 work — the core objective lands well.

**Why this is worth merging:** EU visitors who decline currently produce *zero* data — `gtm.js` was injected inside `enabledAnalytics()`, so it never loaded without analytics consent. Moving GTM out and setting Consent Mode v2 defaults before it loads turns that blind spot into cookieless pings, and gets the four EEA ad signals registered properly. The Arguments-vs-Array test is a good catch — that's the failure mode that silently disables Consent Mode.

**One thing to fix before merge (compliance, not measurement):** revoking consent via the Settings modal flips the gtag signals to `denied`, but PostHog keeps running. `analyticsEnabled` is a one-way latch and there is no `opt_out_capturing` anywhere in `src/`. Since `analytics.js` fires `posthog.capture("$pageview")` on every `astro:page-load` independently of this module, and `disable_session_recording: false`, a visitor who withdraws consent keeps sending identified pageviews and session recordings for the rest of the visit — with soft navigation there may never be a hard reload. Suggest exporting a small helper from `cookieconsent.ts` and calling it from `onChange`:

```ts
export const disableAnalytics = () => {
    if (!analyticsEnabled) return
    posthog.stopSessionRecording()
    posthog.opt_out_capturing()   // persisted, so it also covers later visits
}
```

**One thing to verify on the preview deploy (perf, unrelated to tracking):** `withConsentRegion` rewrites `Cache-Control` to `private, max-age=0, must-revalidate` on *every* HTML response, not just the region-varying edge-cached routes. Prod currently serves `cf-cache-status: HIT` with `public, max-age=0, must-revalidate` on `/docs`. `private` forbids shared caches from storing, so please confirm `cf-cache-status` doesn't regress HIT → MISS on a prerendered page — that would cost TTFB/LCP against the edge-cache work in #5197. `CDN-Cache-Control` can split edge vs. browser behaviour if needed.

**Minor:** the PR description says the region lands in a `kestra_region` cookie, but the implementation deliberately avoids a cookie (that's the whole `Set-Cookie` breaks `cache.put()` rationale) — worth updating, it's what future readers hit first. Also, nothing covers `consentRegion.ts` itself, and `cf.isEUCountry` would be more precise than the `Europe/*` timezone prefix (which over-includes Moscow/Istanbul/Kyiv — fails safe, so just an optimisation).

vfanucci:
I am less technical than you on this topic, let me know !

vfanucci:
@elevatebart  Approved and ready on my side one new conflict on use-cases/index.astro from #5637 (take main, the type fix is obsolete now), then it's yours to merge when you can watch perf

Thanks !

vfanucci:
Rebased on main and pushed — `MERGEABLE` again. There were two conflicts, not one: #5672 landed on `cookieconsent.ts` after my last message.

- `use-cases/index.astro` — took main, the type fix here is obsolete since #5637.
- `cookieconsent.ts` — kept this branch's structure and moved #5672's on-demand consent stylesheet into `cookieBanner.ts`. That chunk is already EU-only, so the consent CSS is now not even referenced from the bundle the rest of the world downloads, which is a bit better than what #5672 got on its own. `initBanner` is `async` and awaits it, so the modal never paints unstyled.

I also closed the one review point I'd left open: revoking via the Settings modal flipped the gtag signals to `denied` but PostHog kept running — `analytics.js` captures a `$pageview` on every `astro:page-load` independently of this module, and with soft navigation there may never be a hard reload. `disableAnalytics()` now stops recording and opts out; since `opt_out_capturing()` persists, re-granting opts back in explicitly. Tests cover revoke, re-grant, the never-granted case, and that the stylesheet is fetched once in the EU and never outside it.

Local: 442 tests pass, `astro check` 0 errors.

My `Cache-Control: private` concern is withdrawn — the comment in `consentRegion.ts` is right, the copy stored in the Workers cache is cloned before the rewrite, so `s-maxage` survives. Still worth a quick `cf-cache-status` check on the preview before you merge.

Yours to merge whenever you can watch perf. Only leftover: the description still says the region lands in a `kestra_region` cookie, which the implementation deliberately avoids.

PR #5195, opened 2026-07-21
fix(seo): stop faceted /blueprints navigation from being a crawler trap

vfanucci:
## Problem

Faceted `/blueprints` URLs (`?tags=`, `?tools=`, `?page=`, `?sort=`, `?q=`) generate a **near-infinite combinatorial URL space**. Because the tag chips toggle values into a comma-joined set (`?tags=Policy,Compliance,AI,Cloud,Infrastructure…`), every filter combination is a distinct crawlable URL — and they were crawlable `<a href>` links.

### Cloudflare evidence (2026-07-18)

- **GPTBot made 79,645 requests in a single day** to `/blueprints?tags=...` URLs — **99.6% of an 80k crawl-spike**.
- These are combinatorial tag-filter permutations of the `/blueprints` listing (e.g. `/blueprints?tags=Policy%2CCompliance%2CAI%2CCloud%2CInfrastructure...`).
- This wastes crawl budget for **GPTBot, Googlebot and Bingbot** and risks **index bloat**. Only the canonical `/blueprints` listing and the real `/blueprints/<slug>` detail pages have value.

## Fix

Four layers, scoped to blueprints faceted navigation + `robots.txt`:

1. **Canonical** — `src/pages/blueprints/index.astro` now passes `canonical="/blueprints"` to the layout whenever a query string is present, so every faceted variant points at the clean, param-free URL.
2. **`noindex, follow`** — the same condition sets `noindex`, emitting `<meta name="robots" content="noindex, follow">` on any faceted listing URL. Links are still *followed* through to the real pages.
3. **`robots.txt`** (`src/pages/robots.txt.ts`) — replaced the narrow `?clid=` / `?size=` rules with:
   ```
   Disallow: /blueprints?*
   Disallow: /*?tags=
   ```
   These block **all** faceted variants. Verified with a Google-style matcher that the clean `/blueprints` listing and `/blueprints/<slug>` (and `/blueprints/patterns/<slug>`) detail pages are **not** matched and stay crawlable/indexable.
4. **Non-crawlable filter links** — the combinatorial anchors now carry `rel="nofollow"` so bots can't discover the permutations:
   - quick-filter chips and the "See all … " link on the listing (`index.astro`);
   - the "See more … " chip on the blueprint **detail** page (`[id].astro`), which was a live discovery vector from indexable pages;
   - blueprint **pagination** links (they inherit the active `?tags=` filter → `?tags=X&page=2`). Threaded via an opt-in `nofollow` prop on `PaginationContainer` → `Pagination` (default `false`, so no other listing is affected).

   Clean links — category tiles (`/blueprints/<slug>`), pattern tiles (`/blueprints/patterns/<slug>`), the single-tag "solo slug" chips, and "Clear all filters" (`/blueprints`) — **remain followable**.

## Safety / scope

- Clean and detail blueprint pages remain fully indexable and crawlable (verified robots matching + canonical logic).
- On-page filtering UX is unchanged — filters run client-side via `navigate()`/`@click.prevent`; `rel="nofollow"` and `noindex` don't affect user interaction.
- The `nofollow` prop defaults to `false`, so pagination on every other listing is untouched.
- No unrelated pages modified. (`OrchPattern.astro` emits a single clean `?q=` link that is already covered by the existing `Disallow: /*?q=` rule plus the new `noindex`, so it was intentionally left alone.)


iitzIrFan (commented):
One thing I am unsure about that, we are using both Disallow: `/blueprints?*` in robots.txt and `noindex` on the faceted pages. If bots are blocked by `robots.txt`, they `won't fetch the page to see the noindex directive`. Should we instead pick one strategy here?

- If the goal is crawl budget, Disallow + canonical seems sufficient.
- If the goal is removing already-indexed faceted URLs, we'd need the pages to remain crawlable so noindex can actually be processed.

Just wanted to check the intended SEO strategy before we approval!

vfanucci:
Great catch @iitzIrFan — you're right that `Disallow` + `noindex` are self-defeating together( Claude was too fast) : once a URL is blocked in robots.txt the crawler never fetches it, so it never sees the `noindex`. They can't both do work at once.

We dug into it and this isn't an indexation problem — **GSC is clean**, no faceted `?tags=` URLs are indexed. It's purely a **crawl-budget** problem from GPTBot. Cloudflare logs for 2026-07-18 are unambiguous:

| Path type | GPTBot hits |
|-----------|-------------|
| `/blueprints?…` (faceted) | **79,628** |
| `/blueprints/<slug>` (real pages) | 17 |

GPTBot is stuck cycling tag permutations, not doing a legitimate crawl. Since there's nothing to *de-index*, `noindex`/`canonical` are inert here (the bot can't fetch the page to read them anyway) and only muddy the intent — so I've dropped them ([3a89dfa](https://github.com/kestra-io/docs/commit/3a89dfa72)). The fix is now pure crawl-control:

- **`robots.txt Disallow`** → blocks the *fetch* (kills the 79.6k). GPTBot honours robots.txt.
- **`rel="nofollow"`** → blocks *discovery* so permutations don't re-enter the crawl graph.

**On your core worry — does this block the real blueprint pages?** No, and I verified it rather than assuming. I ran all **482 real blueprint URLs** from the live sitemap through Google's robots.txt matching algorithm against the new rules: **0 are blocked.** The reason is that `?` is a literal character in robots.txt (only `*` and `$` are wildcards), so `Disallow: /blueprints?*` only matches URLs where a `?` immediately follows `blueprints`. Real pages use a `/` (`/blueprints/<slug>`), so they can't match — they stay fully crawlable and indexable. Kept deliberately scoped so Googlebot/GPTBot keep crawling and citing the real pages.

Hopefully, it's only GPTBot crawling bad URLs, the Google bot is smarter :)

iitzIrFan (approved):

PR #5196, opened 2026-07-21
fix(cache): stabilize asset & icon caching to save crawl budget and kill stale-asset 404s

vfanucci:
## Problem

Googlebot is spending the overwhelming majority of its crawl budget re-fetching **assets**, not content, and users + bots are hitting recurring 404s on stale JS/CSS bundles.

**GSC crawl stats (last 90 days) — share of crawled `200`s:**

| Path | Share of crawled 200s | What it is |
|------|----------------------|------------|
| `/icons/*` | **40.8%** | plugin SVGs |
| `/_astro/*` | **28.6%** | content-hashed JS/CSS bundles |
| **Assets total** | **~70%** | — |
| real content pages | small minority | — |

**Stale-asset 404s (Cloudflare logs): ~5k/week** on `/_astro/*.js` and `*.css`.

**Root cause of both symptoms:**
- `/icons/*` were served with only `Cache-Control: public, max-age=86400` (24h), so bots re-crawled every plugin SVG roughly daily even though the icons never change.
- Every deploy re-hashes the `/_astro/*` filenames. Edge/browser-cached HTML keeps referencing the *previous* deploy's hashes — which no longer exist — so those requests `404`. That both wastes crawl budget (re-download churn) and breaks JS for anyone on a cached page.

## Fix (repo side)

**1. Cache content-hashed assets "forever" + `immutable`.**
- `/_astro/*` — already set to `public, max-age=31536000, immutable` in [`public/_headers`](public/_headers) ✅ (confirmed, no change needed; the ASSETS binding serves these).
- `/icons/*` — bumped from `max-age=86400` → `public, max-age=31536000, immutable` in [`src/pages/icons/[cls].svg.ts`](src/pages/icons/%5Bcls%5D.svg.ts). This is a dynamic SSR endpoint (`prerender = false`), so `_headers` does **not** apply to it — the header has to be set on the `Response` in the route. The URL is keyed by plugin class (`/icons/{cls}.svg`), so a given icon's URL is stable across deploys; a new plugin gets a *new* URL rather than mutating an existing one — which is exactly what makes a year-long `immutable` cache safe.

**2. Keep HTML on a revalidated cache.**
Added a small `setupHtmlCacheControl` middleware in [`src/middlewares/worker.ts`](src/middlewares/worker.ts) that sets `Cache-Control: public, max-age=0, must-revalidate` on **`text/html` responses only**. This is the actual fix for the stale-`/_astro/*.js` 404s: caches may store the page but must revalidate every request (cheap `304`s when unchanged), so a served page always references the *current* asset hashes. JSON / feed / SVG / sitemap endpoints keep their own `Cache-Control` (the guard only touches `text/html`, and extensioned assets bypass the middleware entirely).

**4. Icons already served from a stable URL** (per-plugin-class, no per-deploy hash) — verified, no recrawl churn from URL changes.

## Expected impact
- **Crawl budget:** once bots see `immutable`, they stop re-fetching `/icons/*` and `/_astro/*` — freeing the ~70% currently burned on assets so Googlebot spends it on actual content pages.
- **404s:** revalidated HTML always points at live asset hashes, eliminating the ~5k/week stale `/_astro/*` 404s and the associated broken-JS reports.

## Cloudflare-side follow-ups (not in this repo)
These belong in the Cloudflare dashboard / deploy pipeline, not the repo — listing them so they aren't lost:

1. **(Fix #3) Retain the previous build's `/_astro/*` for a short grace period.** Cloudflare Workers static assets replace `dist/` wholesale on each deploy, so old hashed bundles disappear immediately. The HTML `must-revalidate` change above is what actually closes the 404 window; a grace period is belt-and-suspenders for in-flight cached HTML mid-deploy. Options: keep the prior deploy's `/_astro/` objects for a few minutes, or serve a stale-asset fallback. Needs a deploy-pipeline change.
2. **Verify no Cloudflare Cache Rule / Transform Rule overrides these headers** for `/icons/*`, `/_astro/*`, or HTML documents. Origin `Cache-Control` should be respected; if an Edge Cache TTL rule is forcing HTML caching, it would defeat the `must-revalidate` and should be scoped to assets only.

## Scope
Caching headers / asset hygiene only. No page content, routing, or redirect changes.


iitzIrFan (approved):

vfanucci:
Reviewed against the live prod headers. The `/icons/*` change is the real win here, but two findings — one that makes fix #2 a no-op, one that I think should change the icon TTL before merge.

### 1. Fix #2 (`setupHtmlCacheControl`) doesn't reach the pages that generate the 404s

Live headers today, with this PR **not** merged:

```
/docs                →  cache-control: public, max-age=0, must-revalidate
/plugins/plugin-aws  →  cache-control: public, s-maxage=3600, stale-while-revalidate=86400
```

Two consequences:

- **Prerendered HTML already has `must-revalidate`.** `origin/main` contains neither `setupHtmlCacheControl` nor `must-revalidate` (and `public/_headers` only covers `/_astro/*`), yet `/docs` already returns it — it's the Cloudflare Workers Assets default for HTML documents. So on these routes the middleware rewrites a header that is already correct.
- **On `/plugins/*` and `/blueprints/*` the new header is overwritten.** The edge-cache branch in `src/middlewares/worker.ts` (the `if (cacheKey && finalResponse.status === 200)` block) runs *after* the middleware loop and does `headers.set("Cache-Control", "public, s-maxage=…, stale-while-revalidate=…")`. With `EDGE_CACHE_TTL = 1h` and `EDGE_CACHE_SWR = 24h`, that leaves a ~25h window where a served page can still reference a previous deploy's `/_astro/*` hashes.

Since those SSR routes are the largest URL set on the site and ~37% of bot crawl, the stale-asset 404s mostly originate there — i.e. the segment the fix doesn't cover. Worth verifying against the Cloudflare logs whether the ~5k/week 404s are concentrated on `/plugins/*` referrers before claiming the 404 outcome.

If we want to close it, the options are: add `must-revalidate` to the header set by the edge-cache branch (keeping `s-maxage` for the edge), lower `EDGE_CACHE_SWR`, or the `/_astro/` grace period already listed as follow-up #1. Happy to do the first in a separate PR.

### 2. `immutable` + 1 year on `/icons/*` conflicts with the endpoint's own contract

The rationale states the URL is stable across deploys, which is true — but `immutable` is a claim about the **response body**, not the URL. And the first line of `src/pages/icons/[cls].svg.ts` says:

```ts
// do not prerender icons there are too many of them and are updated frequently
```

The endpoint proxies the live API (`/plugins/icons/{cls}`) and transforms the result (`optimizeSvgIcon`, plus `currentColor` → modifier substitution). So the same URL legitimately changes content: a plugin updating its logo, or any change to `optimizeSvgIcon`. With `max-age=31536000, immutable`, any browser that cached an icon serves the stale one for a year with no revalidation path — a Cloudflare purge fixes the edge, not end-user browsers.

Also worth noting: the response carries **no `ETag` and no `Last-Modified`** (confirmed on prod), so Googlebot can't issue conditional requests — every recrawl is a full `200` with body. That's a large part of why this line item is so expensive.

Suggestion: `public, max-age=2592000, stale-while-revalidate=86400` (30d, no `immutable`) **plus an `ETag`** computed on the final SVG (after the modifier substitution, so `-white` variants get distinct validators). That captures most of the crawl saving, caps staleness at 30 days instead of 365, and enables cheap `304`s.

One caveat for whoever tracks the outcome: GSC crawl stats count **requests**, not bytes, so `200`s becoming `304`s will move them from `OK (200)` to `Not modified (304)` rather than making them disappear — the `By file type → Image` tile will drop less than the raw byte saving suggests. And since that tile is a *share*, `/icons/*` collapsing raises the HTML share mechanically without a single extra HTML fetch; the meaningful signals are the absolute request counts and the `Discovery` share (currently 14%).

Unrelated but visible in the same crawl-stats report: `301` is 8% of crawl requests versus 2% for `404`, so internal links pointing at redirected URLs may be a bigger crawl-budget item than the icons.

PR #5197, opened 2026-07-21
Add edge caching for plugin and blueprint pages

### LINKED ISSUE #5158: Core Web Vitals: origin fails field CWV — SSR plugin/blueprint pages have no edge caching (TTFB 1,136 ms)
## Summary

Field Core Web Vitals (CrUX, real users) **fail at the origin level**, so the negative page-experience signal applies to **the whole of kestra.io**, not a few pages. A full Screaming Frog crawl (16 Jul 2026, 3,997 HTML pages) points to a clear architectural root cause: **the plugin and blueprint pages (~72% of the site) are rendered on-demand (SSR) on every request with no edge caching.**

## Field data — CrUX, origin level (real users)

| Metric | Field | Threshold | Status |
|---|---|---|---|
| **TTFB** | **1,136 ms** | < 800 ms | 🔴 Poor — root cause |
| LCP | 3,005 ms | < 2,500 ms | 🟠 Needs improvement |
| FCP | Needs improvement | — | 🟠 inherits TTFB |
| **INP** | **287 ms** | < 200 ms | 🟠 Needs improvement (JS) |
| CLS | 0.00 | < 0.1 | ✅ Good |

**Origin CWV assessment = Fail** across all 3,784 pages that have field data. LCP, FCP and TTFB are all dragged by the slow server response; INP is dragged by JS. CLS is already excellent — **no work needed there.**

### Lab confirmation (Lighthouse via Screaming Frog, medians)
- Performance score: **67/100** — only **19 of 3,997 pages score ≥ 90**
- TTI: **~14 s** · TBT: **709 ms** · JS execution: **~2 s** · DOM size: **1,646 nodes**

Related PageSpeed issues by share of pages crawled:
- Use Efficient Cache Lifetimes: **94.7%**
- Render-Blocking Requests: **94.7%**
- Legacy JavaScript: **94.7%**
- Reduce Unused JavaScript: **91.4%** · Reduce Unused CSS: **92.6%**
- Reduce JavaScript Execution Time: **79.3%** · Minimize Main-Thread Work: **86.1%**
- LCP Request Discovery not optimised: **92.8%** · Optimize DOM Size: **85.9%**

## Root cause analysis (grounded in this repo)

**1. Plugin & blueprint pages are SSR on every request.**
`src/pages/plugins/[...slug].astro`, `src/pages/plugins/index.astro`, `src/pages/blueprints/[id].astro` and `src/pages/blueprints/index.astro` all declare `export const prerender = false`. These templates back **~2,425 plugin pages + ~472 blueprint pages ≈ 72% of the crawled site**. Every hit runs a full Astro render inside the Cloudflare Worker instead of serving prebuilt HTML from the edge → high TTFB.

**2. No edge caching on HTML responses.**
There is no `Cache-Control` header set for pages in `src/middlewares/worker.ts` (only `src/utils/sitemap.ts` sets `public, max-age=3600`). With no cache directive, Cloudflare re-renders SSR pages on every request. This is the exact mirror of the "Use Efficient Cache Lifetimes" issue affecting 94.7% of pages.

**3. Worker runs first on every request.**
`wrangler.jsonc` sets `"run_worker_first": true`, so the Worker executes before static assets are served, adding latency on every request. The CSP string in `setupContentSecurityPolicyHeaders` is also rebuilt per request.

**4. Three plugin pages exceed Google's 2 MB HTML indexing limit** (content beyond 2 MB is not indexed):
- `/plugins/plugin-ai/rag/io.kestra.plugin.ai.rag.chatcompletion` — **3.3 MB**
- `/plugins/plugin-ai/agent/io.kestra.plugin.ai.agent.aiagent` — **2.9 MB**
- `/plugins/plugin-ai/completion/io.kestra.plugin.ai.completion.chatcompletion` — **2.1 MB**

**5. INP 287 ms** is driven by heavy client JS (Vue island hydration + legacy/unused JS bundles across ~94% of pages).

### P0 — TTFB / server response (biggest lever, whole origin)
- [ ] **Prerender the plugin & blueprint pages** to static HTML at build time (`prerender = true` / `getStaticPaths`) instead of SSR. If the data volume makes full prebuild impractical, use Astro on-demand rendering **with edge caching** (below) so each page is rendered at most once per TTL.
- [ ] **Set `Cache-Control` on HTML responses** in `src/middlewares/worker.ts` for cacheable routes, e.g. `public, s-maxage=3600, stale-while-revalidate=86400`, so Cloudflare serves cached HTML from the edge. Use Cloudflare Cache API / `cf: { cacheEverything: true }` for SSR routes that stay dynamic.
- [ ] **Re-evaluate `run_worker_first: true`** in `wrangler.jsonc` — let static/prerendered assets be served directly by Cloudflare Assets where the Worker isn't required.

### P1 — HTML size / INP / render path
- [ ] Reduce the HTML weight of plugin pages, especially the 3 pages > 2 MB (inline schema/examples appear to be the bulk). Target < 2 MB, ideally well under.
- [ ] Address render-blocking resources and legacy/unused JS (94% of pages): audit Vue island hydration strategy (`client:visible`/`client:idle` where possible), ship modern bundles (module/nomodule), split/trim unused JS & CSS.
- [ ] Set `fetchpriority=high` and avoid lazy-loading the LCP image; make it discoverable in the initial HTML (LCP Request Discovery, 92.8%).

### P2 — nice to have
- [ ] Add `font-display` and preconnect hints for third-party origins (Network Dependency Tree, 98%).

## Acceptance criteria
- Origin CrUX assessment moves from **Fail** to **Pass** (target: TTFB < 800 ms, LCP < 2.5 s, INP < 200 ms — CLS already good).
- Lab TTFB on a representative plugin page (e.g. `/plugins/plugin-ai/rag/...chatcompletion`) drops materially; Cloudflare cache HIT ratio on HTML rises.
- No indexable HTML page > 2 MB.

## Appendix — crawl context
Screaming Frog full crawl, 16 Jul 2026: 4,010 HTML URLs (3,997 × 200 OK). Site is Astro 7 + Vue islands on Cloudflare Workers. CWV numbers above are CrUX field data (origin); lab numbers are Lighthouse via Screaming Frog.

---
*Filed from a technical SEO audit (Screaming Frog export). Root-cause hypotheses reference the current repo config; final implementation approach is the dev team's call.*

elevatebart:
## Summary
Implements edge caching for plugin and blueprint template pages in the Cloudflare Workers middleware. These pages are rendered on every request due to `prerender = false`, causing high time-to-first-byte (TTFB). Since their HTML is public and request-independent, they can now be safely cached at the edge.

## Key Changes
- **Edge cache configuration**: Added `EDGE_CACHE_TTL` (1 hour) and `EDGE_CACHE_SWR` (24 hours) constants synchronized with upstream API cache TTL
- **Page eligibility detection**: Implemented `isEdgeCacheablePage()` to identify cacheable routes (`/plugins`, `/plugins/*`, `/blueprints`, `/blueprints/*`)
- **Cache key normalization**: Added `edgeCacheKey()` function that strips analytics/campaign tracking parameters (utm_*, gclid, fbclid, etc.) and sorts remaining params to prevent cache fragmentation from campaign traffic
- **Cache API integration**: Added `edgeCache()` helper to access Cloudflare's `caches.default` with proper TypeScript typing
- **Request/response handling**: 
  - Check cache before rendering and return cached response if available
  - After rendering, store successful (200 status) responses in edge cache with appropriate `Cache-Control` headers
  - Use `ctx.waitUntil()` to ensure cache write completes without blocking response

## Implementation Details
- Cache key is based on normalized URL (tracking params removed, remaining params sorted) to ensure consistent cache hits across different campaign sources
- Only GET requests to kestra.io are cached to avoid caching redirects or other non-idempotent responses
- Cache headers include `stale-while-revalidate` for graceful degradation if cache expires
- Response body is cloned before caching to avoid consuming the stream

https://claude.ai/code/session_01A1JMAsuvdD5HXBFaURVteN

### DIFF
--- a/src/middlewares/worker.ts
+++ b/src/middlewares/worker.ts
@@ -88,6 +88,80 @@ const noIndex = defineCFMiddleware(async (url, next) => {
 
 const middlewares: CFMiddleware[] = [setupContentSecurityPolicyHeaders, noIndex]
 
+// TTL (seconds) for edge-cached SSR HTML. Kept in sync with the 1h upstream
+// API cache in src/utils/fetch.ts so a page's HTML and its data expire
+// together. Once s-maxage lapses the next request is a cache miss and re-renders.
+// `stale-while-revalidate` is advisory for browsers and any downstream shared
+// cache (the Workers Cache API does not revalidate in the background itself).
+const EDGE_CACHE_TTL = 60 * 60
+const EDGE_CACHE_SWR = 60 * 60 * 24
+
+// The plugin & blueprint templates declare `prerender = false`, so every hit
+// re-runs a full Astro render inside the Worker (high TTFB — see issue #5158).
+// Their HTML is public and request-independent (any content that varies —
+// `?version=` on plugins, `?q=` on pattern pages — is part of the URL and
+// therefore part of the cache key), so it is safe to serve from Cloudflare's
+// edge cache. Other `prerender = false` routes (/api/*, /t/*, careers apply,
+// cloud-early-access, hubspot-submit, …) are intentionally excluded.
+function isEdgeCacheablePage(url: URL): boolean {
+    const path = url.pathname
+    return (
+        path === "/plugins" ||
+        path.startsWith("/plugins/") ||
+        path === "/blueprints" ||
+        path.startsWith("/blueprints/")
+    )
+}
+
+// Analytics/campaign query params never change the rendered HTML. Stripping
+// them (and sorting the rest) from the cache key keeps campaign traffic from
+// fragmenting the cache into per-visitor misses.
+const TRACKING_PARAMS = new Set([
+    "utm_source",
+    "utm_medium",
+    "utm_campaign",
+    "utm_term",
+    "utm_content",
+    "utm_id",
+    "gclid",
+    "gbraid",
+    "wbraid",
+    "fbclid",
+    "msclkid",
+    "mc_cid",
+    "mc_eid",
+    "ref",
+    "ref_src",
+    "yclid",
+    "twclid",
+    "igshid",
+    "_hsenc",
+    "_hsmi",
+])
+
+// `caches.default` is a Cloudflare Workers runtime extension that is absent
+// from the ambient DOM `CacheStorage` type, so we narrow it explicitly (the
+// runtime values here are DOM `Request`/`Response`).
+interface EdgeCache {
+    match(request: Request): Promise<Response | undefined>
+    put(request: Request, response: Response): Promise<void>
+}
+
+function edgeCache(): EdgeCache {
+    return (caches as unknown as { default: EdgeCache }).default
+}
+
+function edgeCacheKey(url: URL): Request {
+    const keyUrl = new URL(url.toString())
+    for (const param of [...keyUrl.searchParams.keys()]) {
+        if (TRACKING_PARAMS.has(param.toLowerCase())) {
+            keyUrl.searchParams.delete(param)
+        }
+    }
+    keyUrl.searchParams.sort()
+    return new Request(keyUrl.toString(), { method: "GET" })
+}
+
 export default {
     async fetch(
         request: Parameters<typeof handle>[0],
@@ -125,6 +199,24 @@ export default {
             return handle(request, env, ctx)
         }
 
+        // Edge-cache the rendered HTML of the SSR plugin & blueprint pages.
+        // Because `run_worker_first: true` routes every request through this
+        // Worker, Cloudflare's CDN never caches the generated HTML on its own;
+        // we do it explicitly via the Cache API so each page is rendered at
+        // most once per TTL (and the CSP is rebuilt at most once per TTL too).
+        const canEdgeCache =
+            request.method === "GET" &&
+            url.host === "kestra.io" &&
+            isEdgeCacheablePage(url)
+        const cacheKey = canEdgeCache ? edgeCacheKey(url) : undefined
+
+        if (cacheKey) {
+            const hit = await edgeCache().match(cacheKey)
+            if (hit) {
+                return hit
+            }
+        }
+
         let response: Response | undefined = undefined
         function next() {
             if (response) {
@@ -136,6 +228,24 @@ export default {
         for (const middleware of middlewares) {
             response = await middleware(new URL(request.url), next, request)
         }
-        return await next()
+
+        const finalResponse = await next()
+
+        if (cacheKey && finalResponse.status === 200) {
+            const cachedResponse = new Response(
+                finalResponse.body,
+                finalResponse,
+            )
+            cachedResponse.headers.set(
+                "Cache-Control",
+                `public, s-maxage=${EDGE_CACHE_TTL}, stale-while-revalidate=${EDGE_CACHE_SWR}`,
+            )
+            ctx.waitUntil(
+                edgeCache().put(cacheKey, cachedResponse.clone()),
+            )
+            return cachedResponse
+        }
+
+        return finalResponse
     },
 }
--- a/src/middlewares/worker.ts
+++ b/src/middlewares/worker.ts
@@ -232,18 +232,35 @@ export default {
         const finalResponse = await next()
 
         if (cacheKey && finalResponse.status === 200) {
-            const cachedResponse = new Response(
-                finalResponse.body,
-                finalResponse,
-            )
-            cachedResponse.headers.set(
+            const response = new Response(finalResponse.body, finalResponse)
+            response.headers.set(
                 "Cache-Control",
                 `public, s-maxage=${EDGE_CACHE_TTL}, stale-while-revalidate=${EDGE_CACHE_SWR}`,
             )
+
+            // Cloudflare's Cache API rejects `put()` on any response carrying a
+            // `Set-Cookie` header. Because the write runs inside
+            // `ctx.waitUntil`, that rejection would be swallowed — silently
+            // disabling the cache and re-rendering on every request. These
+            // routes are public and request-independent (no server-set cookies
+            // today), but strip `Set-Cookie` from the stored copy so a future
+            // cookie can never quietly break caching, and guard the write so a
+            // failed `put()` degrades to "not cached" instead of an unhandled
+            // rejection. The response returned to this (cache-miss) visitor is
+            // left untouched.
+            const toStore = response.clone()
+            toStore.headers.delete("Set-Cookie")
             ctx.waitUntil(
-                edgeCache().put(cacheKey, cachedResponse.clone()),
+                edgeCache()
+                    .put(cacheKey, toStore)
+                    .catch((error) => {
+                        console.error(
+                            `Edge cache put failed for ${cacheKey.url}: ${error}`,
+                        )
+                    }),
             )
-            return cachedResponse
+
+            return response
         }
 
         return finalResponse


iitzIrFan (commented):
It's a solid Edge caching but, need to think about SSR on every req though!

elevatebart:
Thanks @iitzIrFan! Good thing to pin down — worth clarifying what the edge cache actually changes for SSR here.

With `caches.default` in front of these routes, SSR no longer runs on every request. It runs **at most once per URL per TTL** (1h), plus on genuine cache misses (first hit, post-deploy, or eviction). Every cache HIT skips the Astro render entirely — and also skips the per-request CSP rebuild — so for the ~72% of the site these routes cover, the common path is now edge-served HTML with no render.

What's left is the cold-cache path: the first visitor per URL per hour still pays full SSR TTFB. Eliminating that entirely means prerendering these ~2,900 plugin/blueprint pages at build time (`getStaticPaths`) — the issue's alternative P0. That's a much bigger change with real tradeoffs: all plugin/blueprint data has to be fetchable at build, builds get substantially longer, and data updates require a rebuild (likely why SSR was chosen in the first place).

So this PR intentionally keeps the pages dynamic while removing the per-request render for the vast majority of traffic. Happy to spike full prerender as a separate follow-up if we want to compare the two — just say the word.

---

iitzIrFan:
That should give us happy `TTFB` I guess now?

elevatebart:
Yeah — for cache HITs, which will be the vast majority of traffic. On a HIT the response is served straight from Cloudflare's edge with no Astro render, so TTFB drops to roughly edge-latency (tens of ms) instead of the ~1.1s SSR path from the issue. The first visitor per URL each hour still pays the SSR cost (cache miss), so the metric to watch on prod is the **cache HIT ratio** climbing and origin TTFB falling as the field data (CrUX) refreshes.

One heads-up so the numbers aren't misread: the CI Lighthouse benchmark won't show this improvement — it runs against `http://localhost:8787`, and the cache is gated to `url.host === "kestra.io"`, so it's bypassed entirely in that lab run. The gain will show up on the deployed site with a warm cache, not in the PR's benchmark table.

---

vfanucci:
Set-Cookie will make cache.put() throw. Cloudflare's Cache API rejects any put() on a response carrying a Set-Cookie header. If the SSR response sets any cookie (consent, HubSpot, analytics), the put() fails — and since it's inside ctx.waitUntil(), it becomes a silent unhandled rejection: the cache is never populated and we re-render on every request while believing we're cached. 
Can we confirm these responses never set cookies, or strip Set-Cookie before caching ?

elevatebart:
Great catch @vfanucci — you're right on both counts: `cache.put()` rejects on `Set-Cookie`, and inside `ctx.waitUntil()` that rejection is swallowed, so we'd silently never populate the cache and re-render every request while thinking we're cached.

Confirmed **and** hardened in `1860c18`:

1. **No cookies today** — there's no `Astro.cookies.set` or `Set-Cookie` anywhere in `src`, and the plugin/blueprint page frontmatter sets none. So these responses don't set cookies as things stand.
2. **Future-proofed anyway** — I strip `Set-Cookie` from the *stored* copy before `put()`, so if a consent/HubSpot/analytics cookie is ever added server-side it can't quietly disable caching. The response returned to the cache-miss visitor is left untouched (keeps any cookie it carries).
3. **No silent rejection** — the `put()` now has a `.catch()` that logs, so any write failure (Set-Cookie or otherwise) degrades gracefully to "not cached for this entry" instead of an unhandled rejection.

```js
const toStore = response.clone()
toStore.headers.delete("Set-Cookie")
ctx.waitUntil(
    edgeCache().put(cacheKey, toStore).catch((error) => {
        console.error(`Edge cache put failed for ${cacheKey.url}: ${error}`)
    }),
)
return response
```

Stripping the cookie from the shared cached copy is also the correct semantics here: these pages are public and request-independent, so per-response cookies shouldn't be in a copy served to everyone.

---

iitzIrFan:
Cool! This addresses the server-side rendering overhead for the common path. The remaining work is confirming the improvement with production field data I guess?

elevatebart:
Exactly. Once this is deployed, the things to watch — which map to the acceptance criteria in #5158 — are:

- **Cloudflare cache HIT ratio** on the `/plugins` and `/blueprints` HTML responses climbing (Analytics → Caching, or the `cf-cache-status` header on a HIT).
- **Origin TTFB** dropping, ideally under the 800 ms threshold.
- **CrUX field data** flipping the origin CWV assessment from Fail toward Pass as it refreshes (it's a rolling 28-day window, so give it a few weeks post-deploy).

Worth noting the lab/Lighthouse numbers in this PR won't show it (they run against localhost, which the cache bypasses) — the signal is field data + the edge cache-hit metric on prod. And this only tackles the P0/TTFB lever; the P1/P2 items in the issue (the 3 plugin pages >2 MB, JS/INP, fonts) are still separate follow-ups.

---

iitzIrFan:
Looking forward to seeing the production metrics once this is deployed. Hopefully we will see the cache HIT ratio increase and TTFB trend in the right direction 👍

vfanucci:
Ok, so I can update the base branch and merge this PR ? 
Thanks !

vfanucci (approved):
OK

elevatebart:
Yes — good to go from the code side. It's approved, `mergeable_state` is `clean`, and all 7 checks are green on the latest commit (`1860c18`). Updating the base branch is safe: it just re-triggers CI, and since the only change is scoped to the Cloudflare worker's plugin/blueprint edge-cache path, there's nothing that should conflict with `main`. Merge whenever you're ready 👍

---

vfanucci:
KPI (origin, p75) | Before | Expected after | Flip likely?
-- | -- | -- | --
TTFB | 1,137 ms 🟠 | < 800 ms 🟢 | ✅ direct target
LCP | 3,111 ms 🟠 | ~2,200 ms → < 2,500 🟢 | ✅ the real prize
FCP | 2,498 ms 🟠 | < 1,800 ms 🟢 likely | ✅ follows TTFB
INP | 284 ms 🟠 | 284 ms — unchanged | ❌ not this PR
CLS | 0.00 🟢 | 0.00 🟢 | already green

iitzIrFan:
> KPI (origin, p75)	Before	Expected after	Flip likely?
> TTFB	1,137 ms 🟠	< 800 ms 🟢	✅ direct target
> LCP	3,111 ms 🟠	~2,200 ms → < 2,500 🟢	✅ the real prize
> FCP	2,498 ms 🟠	< 1,800 ms 🟢 likely	✅ follows TTFB
> INP	284 ms 🟠	284 ms — unchanged	❌ not this PR
> CLS	0.00 🟢	0.00 🟢	already green

This looks promising🤞

vfanucci:
@iitzIrFan for the follow-up, we will know excatly the figures in 2 weeks (for 28days range) 

⚠️ the window is still ~50 % pre-deploy — 14 days before, 13 days after.


KPI (origin, p75) | Before | Expected after | Actual, 5 Aug | Called it?
-- | -- | -- | -- | --
TTFB | 1,137 ms 🟠 | < 800 ms 🟢 | 1,094 ms 🟠 (−43 ms) | ❌ barely moved
LCP | 3,111 ms 🟠 | ~2,200 → < 2,500 🟢 | 2,776 ms 🟠 (−335 ms) | 🟡 right direction, not green
FCP | 2,498 ms 🟠 | < 1,800 ms 🟢 | 2,345 ms 🟠 (−153 ms) | ❌ didn't follow
INP | 284 ms 🟠 | 284 ms — unchanged | 278 ms 🟠 | ✅ correct
CLS | 0.00 🟢 | 0.00 🟢 | 0.00 🟢 | ✅ still green

Mobile verdict: still FAIL — LCP good% 69.6 and INP good% 67.2, both under the 75 % bar. 
The LCP win is mostly not the cache. LCP contains TTFB, so a 43 ms TTFB gain cannot mechanically produce a 335 ms LCP gain. Three other perf PRs merged within 72 hours — [#5213](https://github.com/kestra-io/docs/pull/5213) (homepage CWV), [#5200](https://github.com/kestra-io/docs/pull/5200) (plugins CLS + GTM main-thread + LCP), [#5215](https://github.com/kestra-io/docs/pull/5215) (lazy-load Arcade embeds)

Claude advices let me know what do you think about it @iitzIrFan 

1) Implement real SWR — return the stale entry immediately, re-render in ctx.waitUntil. This is the highest-leverage fix: it turns "one slow visitor per URL, per PoP, per hour" into "always fast."
2) Raise EDGE_CACHE_TTL — but it's deliberately aligned with the 1 h upstream API cache, so this needs purge-on-deploy or accepting HTML staler than its data.
3) INP is the remaining lock on a mobile PASS (278 ms vs 200 target) — JS weight, out of scope for this PR.

iitzIrFan:
Thanks for running the audit, @vfanucci! The mixed 28-day window makes sense, so I wil wait for the next report before drawing any conclusions.

From the numbers, it looks like the edge cache alone is not enough to move TTFB as much as we expected. I will also take a look at the SWR suggestion and our current cache strategy to see if there's anything worth improving, along with checking the cache hit ratio to understand where we are still missing.

Let's see what the metrics look like once the full 28-day window is post-deploy.

vfanucci:
@iitzIrFan 

CWV verdict — first fully post-deploy CrUX window (2026-07-28 → 08-24)

Merged 2026-07-21. This is the first 28-day window with zero pre-deploy data, so the numbers below are attributable to the shipped state of the site.

Metric | Baseline 07-21 | Now | Δ
-- | -- | -- | --
TTFB | 1140 ms | 1045 ms (65.6% good) | −95 ms
LCP | 3120 ms (64.7% good) | 2592 ms (73.0% good) | −528 ms, +8.3 pts
FCP | 2521 ms | 2235 ms | −286 ms
INP | 284 ms | 312 ms (64.1% good) | +28 ms ⚠️
CLS | 0.00 | 0.00 (96.2% good) | —

All form factors combined: PASS (LCP 2118 / INP 136 / CLS 0.00).
Mobile: still FAIL 

Recommendation (Claude TBD) : keep the cache (it's healthy — cf-cache-status: HIT, correct headers, 301s correctly excluded 5 weeks in), add real SWR as a follow-up, and open a separate JS-weight issue for INP.

iitzIrFan:
Thanks for the full post-deploy audit @vfanucci — agree with the read and the recommendation. My take:

1. **Keep the cache as-is.** It's healthy and it delivered the LCP/FCP gains; no reason to touch the TTL yet.
2. **SWR is the right next step for TTFB.** The −95 ms plateau confirms the cold-cache path is the bottleneck: at p75 across ~2,900 long-tail plugin/blueprint URLs, per-PoP per-hour misses are frequent enough to dominate. Serving the stale entry immediately and re-rendering in `ctx.waitUntil()` turns that into "always fast" without changing data freshness beyond what we already accept. I'll open a follow-up issue for it.
3. **INP deserves its own issue.** It regressed to 312 ms and is now the main blocker for a mobile pass — that's JS weight (island hydration, legacy/unused bundles), unrelated to this PR. I'll file it separately referencing the P1 items from #5158.

One thing worth checking before the SWR work: the actual cache HIT ratio on `/plugins` and `/blueprints` HTML in Cloudflare Analytics. If it's lower than expected, there may also be cache-key fragmentation or eviction to fix, which would compound the SWR win.

@elevatebart — given TTFB only moved −95 ms, curious whether you'd still start with SWR or whether this shifts you toward the full-prerender spike you offered earlier. The cold-miss rate looks higher than we assumed, which changes the math a bit.

PR #5213, opened 2026-07-22
perf(home): fix top 3 Core Web Vitals bottlenecks from PageSpeed report

elevatebart:
Fixes the three most impactful problems from the PageSpeed Insights report for https://kestra.io (mobile: performance **53**, lab LCP **7.1 s**, TBT **590 ms**; field LCP **3.2 s**, INP **462 ms** — both failing Core Web Vitals).

## Why these three

Of the performance score's metric weights, LCP contributed only **1 of 25 possible points** — by far the biggest lever — followed by TBT (15/30). CLS is already perfect. So the fixes target LCP first, then main-thread work (TBT/INP), then network payload contention.

## 1. LCP — the mobile hero background loaded late and weighed 544 KiB

The LCP element is `div.mobile-bg` in `Hero.astro`, whose image is a CSS `background-image`. Two compounding problems from the report:

- **830 ms resource load delay**: as a CSS background, the browser only discovers the image after downloading and parsing the stylesheets ("LCP request discovery" audit: not discoverable in initial document, no `fetchpriority=high`).
- **372 KiB of estimated waste**: it shipped as a 786×1348 PNG (544 KiB) — CSS `url()` assets bypass image optimization entirely.

**Fix:** converted to WebP at the same dimensions (557 → **117 KiB, −79 %**), added a `<link rel="preload" as="image" fetchpriority="high" media="(max-width: 767.98px)">` in the home page `<head>` so the download starts with the document, and scoped the `background-image` to the mobile breakpoint so desktop browsers never fetch it. Verified in the build output that the preload href and the CSS reference resolve to the same hashed asset.

## 2. TBT / INP — the hero animation burned 3.3 s of CPU on mobile while invisible

`hero-animation.astro` was the single biggest main-thread consumer in the trace (**3,266 ms** total CPU, several long tasks). The animation column is `display: none` below the `md` breakpoint — on phones, every `requestAnimationFrame` loop (canvas particles, badge `getBoundingClientRect` reflow, SVG border gradient), interval and DOM mutation ran for an element nobody can see. This directly feeds lab TBT (590 ms) and the failing field INP (462 ms), since long tasks compete with user input.

**Fix:** the animation script now waits for `matchMedia("(min-width: 768px)")` to match before booting (top-level await — it never runs on phones), and an `IntersectionObserver` skips all per-frame work while the hero is scrolled out of view, which also stops the burn on desktop after the user scrolls past it.

## 3. Network payload — a 396 KiB decorative background competing for mobile bandwidth

`bg_price.webp` (used by the `GetStarted` section on the home page via `WithLogo.astro`) was flagged in "enormous network payloads": 387.7 KiB for a below-the-fold decorative background. The source was **3658×1604** but renders at 65 % of section width — roughly 2× larger than ever displayed. On the throttled 1.64 Mb/s connection Lighthouse simulates, that's ~1.9 s of bandwidth taken from render-critical resources (it also loads on /pricing and /get-started).

**Fix:** resized to 1920 px wide and recompressed (396 → **74 KiB, −81 %**). Visually indistinguishable for a blurred purple glow background.

## Net effect

~870 KiB less transfer on a page with a 2,640 KiB total payload, the LCP image starts downloading immediately at high priority, and mobile main-thread time drops by roughly 3 s of animation work.

## Verification

- `astro build` (same flags as CI), `astro check` (0 errors/warnings/hints), `oxlint`, and `vitest` all pass.
- Confirmed in `dist/` output: preload URL == CSS asset URL (`/_astro/bg-mobile.fJt4TG7z.webp`), and the built hero-animation chunk starts with the media-query gate.
- The Lighthouse Benchmark workflow on this PR should show the lab-metric movement.

## Not addressed (follow-up candidates)

Render-blocking CSS (440 ms est., `layout.css` is ~90 % unused on this page — mostly the global Bootstrap bundle) and third-party tag weight (GTM loads 4 configs, 619 KiB) are the next levers, but both need decisions beyond a safe mechanical change.

https://claude.ai/code/session_01E12ENgX6XVFZTRqzBgVaXR

---
_Generated by [Claude Code](https://claude.ai/code/session_01E12ENgX6XVFZTRqzBgVaXR)_


elevatebart:
**A note on the Lighthouse Benchmark's "Home 🔻" flag — it's baseline noise, not a regression from this PR.** I dug into the workflow logs to verify rather than hand-wave:

| Run | Code | Home perf |
|---|---|---|
| main `48c5f08` @ 06:29 (this PR's base) | A | **78** |
| main `0f1247d0` @ 07:18 (CI-only change → used as baseline) | A | **87** |
| PR run 1 @ 07:54 | A + this PR | **76** |
| PR run 2 @ 08:07 (re-run, same commit) | A + this PR | **71** |

Two observations:

1. **Main swung 78 → 87 in 49 minutes on identical home-page code** — the single-run baseline just happens to be the lucky sample. Against the 06:29 sample this PR would read −2, not −11/−16. The same noise shows elsewhere: About Us moved 85 → 98 *between two runs of this same PR commit*, and Blog Index's "🟢 +10" is measured against a baseline sample (65) that main itself contradicted an hour earlier (75).
2. **This benchmark runs desktop (1350×940, no CPU throttle), which doesn't exercise anything this PR changes.** On desktop the animation media-query gate passes immediately (animation boots exactly as before), the preload is mobile-scoped (never fetched), and desktop now downloads strictly less than before (the mobile background is breakpoint-scoped, `bg_price.webp` is 81% smaller). The PR targets the failing *mobile* field metrics (LCP 3.2 s, INP 462 ms), which this desktop benchmark doesn't measure.

If useful as a follow-up, running Lighthouse 3× per page and reporting the median would remove this class of false signal from the PR comments.

---

iitzIrFan (approved):

PR #5215, opened 2026-07-22
perf: lazy-load Arcade demo embeds behind a click-to-play facade

elevatebart:
## Why

The mobile PageSpeed report on `/get-started` scores **performance 35** (LCP 9.2 s, TBT 3,460 ms, 6.8 MB total) — the worst page on the site and the #2 by traffic. The dominant cause is the **Arcade product-demo iframe**: ~3 MB from `demo.arcade.software` + `cdn.arcade.software` (nearly half the page) that boots a full Next.js app and fires the four longest tasks in the trace (528/368/337 ms). It already has `loading="lazy"`, but the section sits high enough that Chrome's near-viewport threshold loads it almost immediately on mobile.

The same embed appears on 5 pages, so the fix is a reusable component.

## What

New `src/components/common/ArcadeEmbed.astro` — a lightweight facade: a poster button (play icon + label) that injects the real Arcade iframe only on the first click. Nothing from arcade.software is fetched until the visitor opts in.

Wired into all 5 Arcade embeds:

| File | Page(s) |
|---|---|
| `get-started/DemoPreview.astro` | /get-started |
| `infrastructure/Arcade.astro` | /infra-automation |
| `common/VerticalsIntro.astro` | /infra-automation, /data, /ai-automation intros |

The facade's root fills its wrapper (`position:absolute; inset:0`), matching each existing aspect-ratio container, so it's a drop-in for the previous `<iframe>` in all three shapes. On click it builds the iframe with `allow="clipboard-write"`, `allowfullscreen`, and appends `autoplay=true`. Re-initializes on `astro:page-load` for client navigation.

## Effect

Removes ~3 MB of transfer and the Arcade long tasks from `/get-started`'s initial load (and the other four pages), directly targeting the failing mobile LCP/TBT. Desktop is unchanged for users who click to play.

## Verification

- `astro build` passes (exit 0); `oxlint` clean on all changed files.
- Facade markup verified to fill each of the three wrapper shapes; iframe injection preserves the original embed attributes.

## Follow-ups (from the same PageSpeed sweep, not in this PR)

- `/orchestration/*`: defer the 1.15 MB vue-flow `Topology.client.js` behind an IntersectionObserver (perf 48, TBT 2,120 ms).
- `/docs/quickstart`: a YouTube facade reusing this same pattern (perf 44).
- CSS `background-image` PNGs → WebP at build time (separate PR).

---
_Generated by [Claude Code](https://claude.ai/code/session_013wPnocqds1yTbJ5QrY49GV)_


elevatebart:
The 🔻 flags in the Lighthouse Benchmark above (Home −20, Docs −12, Blueprint −11) are **baseline noise, not regressions from this PR.**

This PR only touches four pages — `/get-started`, `/infra-automation`, `/data`, `/ai-automation` (the Arcade demo facade) — and **none of them are in the benchmark's page sample** (`tests/fixtures/page-sample.mjs`). No shared layout or global bundle changes, so it's not possible for this PR to have moved Home/Docs/Blueprints. The deltas are the known single-run, desktop (1350×940, no CPU throttle) benchmark variance already documented on #5213 (main swung 78 → 87 in 49 min on identical home-page code).

What this change actually targets is the **mobile field metrics** (LCP/INP/TBT) on the pages it touches: the Arcade iframe (~3 MB + several long tasks) no longer loads until the visitor clicks play. The desktop, unthrottled, single-run benchmark here doesn't exercise that. No action needed on these flags.

---

Piyush-r-bhaskar (approved):

iitzIrFan (commented):
LGTM from my side but claude has some comments on this which is as follows :)

1. Double click-listener binding in prod
The script runs initArcadeFacades() immediately and on astro:page-load. Since ClientRouter is enabled in PROD, astro:page-load also fires on the initial load, so each facade ends up with two click listeners (the second builds an orphan iframe that never attaches — harmless but sloppy). Our convention (Connect.astro) uses an astro:page-load listener only. Either drop the immediate call, or add an idempotency guard:

```
facades.forEach((facade) => {
    if (facade.dataset.arcadeInit) return
    facade.dataset.arcadeInit = "1"
    ...
})
```
2. Missing visual-regression baselines
Adding /get-started to page-sample.mjs makes visual-regression.spec.ts screenshot it at 3 viewports, but there are no Get Started-*-linux.png baselines committed — Playwright fails on missing baselines in CI. Either commit the baselines (npx playwright test --update-snapshots) or drop this fixture change (it's orthogonal to the perf fix).

Minor

The original .arcade-embed in VerticalsIntro used a -1px/+2px overlap to hide a hairline seam inside the gradient border. The injected iframe now uses inline inset:0, and the parent's scoped .arcade-embed rule won't match a JS-created element — worth a quick visual check on /infra-automation for a 1px line.
autoplay=true is now added to embeds that previously didn't have it (DemoPreview, VerticalsIntro). User-initiated so fine, just flagging the behavior change.
Facade uses hardcoded hex (#0e0e10, #718bfe) instead of --ks-* tokens — minor consistency nit.
Solid change overall 👍

elevatebart:
Addressed the two review comments from @iitzIrFan's review:

1. **Double click-listener binding** — fixed in 3d73065 with an idempotency guard (`data-arcade-init` marker on the facade), matching the suggested approach. Each facade now only ever gets one listener attached, regardless of whether `astro:page-load` also fires on initial load.

2. **Missing visual-regression baselines for `/get-started`** — agreed this is real (the fixture change does add it to `visual-regression.spec.ts`'s screenshot matrix), but since that check is non-blocking and manual, we'll generate the baseline artifacts (`npx playwright test --update-snapshots`) once this PR merges rather than block on it here.

Left the minor nits (hairline seam check on `/infra-automation`, hardcoded hex vs `--ks-*` tokens, `autoplay` behavior change) as-is since they're flagged as non-blocking — happy to follow up if you'd like those addressed too.

---

iitzIrFan (approved):
Rest is good, thanks 🙌

PR #5244, opened 2026-07-29
fix(sitemap): stop listing plugin URLs that redirect

vfanucci:
First of three PRs on the same root cause. **Stacked: #2 and #3 are based on this branch.**

## Root cause

`/plugins/[...slug].astro` strips the subgroup segment with a 301 when a plugin exposes a **single** subgroup (39 of the 96 plugins that have subgroups). Four places build plugin URLs and none of them applied that rule, so each emitted URLs that redirect.

This PR extracts the rule and fixes the first place: the sitemap.

## The sitemap was advertising 124 redirects

`plugins.xml` adds the subgroup segment whenever the API payload has a `subGroup` field. **124 of its 2387 plugin URLs answered 301** — 88 element URLs and 36 subgroup pages. Verified live, every URL, not a sample.

## Changes

`src/utils/plugins/canonicalUrl.ts` — `buildPluginUrlIndex()`, `canonicalPluginPath()`, `canonicalPluginUrl()`. The sitemap now generates from it.

Two things the old generator got wrong, both fixed by keying the lookup on the **element** rather than the iterated plugin:

- **`clsToSubgroup` is global across plugins**, so a subgroup slug leaked between plugins sharing a title. `plugin-git` has a single entry with no `subGroup`, yet inherited the `git` slug from `plugin-ee-git`'s subgroup and emitted `/plugins/plugin-git/git/<cls>`. Those answer **200 and are self-canonical**, so they were *duplicates* of `/plugins/plugin-git/<cls>`, not redirects. They are the 3 non-301 URLs this drops. They stay reachable — only the sitemap stops advertising them. Collapsing the duplicate itself is a separate concern (a canonical tag on the plugin page, not the sitemap).
- **A class exposed by both an OSS and an EE plugin** (`io.kestra.plugin.git.TenantSync` is in `plugin-git` and `plugin-ee-git`) now resolves to the OSS page, which is where the site's own redirects land.

## Net effect

| | |
|---|---|
| plugins.xml | 2387 → 2348 URLs |
| added | 88, **all 88 answer 200** |
| removed | 127, of which **124 answered 301** and 3 are the duplicates above |

## Validation

Beyond the unit tests, I built the index from the real `/plugins/subgroups` payload and compared it to the destination the crawler actually reached, for all **390** classes linked in flat form across the site:

- **383 exact matches, 0 mismatches**
- 7 legacy class names absent from the payload (kept in a plugin's `aliases`, or a removed task) → `canonicalPluginUrl` returns undefined and callers keep their fallback

The precedence rule that emerged from that comparison is encoded and commented: OSS outranks EE, then a subgroup-scoped entry outranks the plugin root entry (which repeats every element its subgroups declare without saying which subgroup they belong to).

`npm run test:unit` passes (23 tests).

## Note on impact

None of these URLs has Search Console impressions. This is crawl budget and index hygiene, not traffic recovery.


iitzIrFan (commented):
Hey @vfanucci, could you please resolve the merge conflicts before I take a look at it for review? 
Thanks!

vfanucci:
> Hey @vfanucci, could you please resolve the merge conflicts before I take a look at it for review? Thanks!

Done sorry @iitzIrFan

iitzIrFan:
No worries @vfanucci,  I will take a look in the morning 👍

---

Edit:
Meanwhile, could you check why the CI is red?

---

vfanucci:
> No worries @vfanucci, I will take a look in the morning 👍
> 
> Edit: Meanwhile, could you check why the CI is red?

Relaunched, all good

iitzIrFan (changes requested):
- Delete buildSubGroupSegmentPredicate, buildPluginMappings, PluginMappings, and the now-unused plugin.test.ts code.
- Add the canonicalUrl.test.ts fixture where the root entry repeats subgroup classes, and verify the result is the same regardless of array order.
- Also remove the now-unused slugify import from plugin.ts.

iitzIrFan (approved):
Approving to unblock the stack (#5245 and #5246 are already approved and based on this branch).

The core change is correct and verified: `buildPluginUrlIndex` / `canonicalPluginPath` / `canonicalPluginUrl` mirror the single-subgroup 301 rule from `[...slug].astro`, the OSS-over-EE and subgroup-over-root precedence matches the live crawl validation, and all unit tests pass on the branch head. CI is green.

**Follow-up expected once the stack lands** (waiving my earlier review items rather than blocking the base PR on them):
- Delete the now-dead `buildSubGroupSegmentPredicate`, `buildPluginMappings`, and `PluginMappings` from `src/utils/plugins/plugin.ts`, along with the `plugin.test.ts` cases covering them and the then-unused `slugify` import.
- Add the `canonicalUrl.test.ts` fixture where the plugin root entry repeats subgroup classes, asserting the index is the same regardless of payload array order (equal-rank ties are currently first-seen-wins).

@vfanucci please open that cleanup PR after #5246 merges so the dead code doesn't linger.

PR #5245, opened 2026-07-29
fix(blueprints): link the Tasks panel straight at the canonical plugin page

vfanucci:
Second of three on the same root cause. **Stacked on #5244** — review that one first; this diff is only the blueprint change. Retarget to `main` once #5244 merges.

## The single biggest chunk of the debt

The Tasks side panel on every blueprint page built its links as `/plugins/<cls>` from the raw Java class name ([BlueprintMarkdownBody.vue:18](https://github.com/kestra-io/docs/blob/main/src/components/blueprints/BlueprintMarkdownBody.vue#L18)). Every one took **two redirects** before landing:

1. the middleware lowercases the path → `/plugins/io.kestra.plugin.core.log.log`
2. `[...slug].astro` rewrites the flat class to its plugin/subgroup URL → `/plugins/core/log/io.kestra.plugin.core.log.log`

A crawl of kestra.io found **390 such URLs carrying ~2,076 internal inlinks** — 487 blueprint pages × 3-5 tasks each. All of it our own pages pointing at our own pages through a double redirect. Nothing in `src/contents` to grep: the links are generated at render time, which is why they survived every content-side link cleanup.

## Change

Resolve each class server-side with the index from #5244 and hand the island a map of **only the classes that blueprint uses**, so the hydration payload grows by 3-5 entries rather than the 3.2 MB plugin payload. `/plugins/subgroups` is already fetched on this page by `fetchCoreToolIndex`, so this adds no round trip.

Fallbacks, both deliberate:

- a class the payload does not know — a legacy name kept in a plugin's `aliases`, or a task whose plugin was removed — is left out of the map and falls back to the previous flat form, which still reaches the right page, just via the redirects
- a plugin API failure yields an empty map, not a broken page (same defensive shape as `fetchCoreToolIndex`)

## Verification

Ran a dev server against real API data and rendered five blueprints. The panel emits **18 distinct task URLs, and all 18 answer 200 with zero redirects** in production. Both shapes covered:

| | |
|---|---|
| multi-subgroup plugin, segment kept | `/plugins/core/flow/...`, `/plugins/plugin-dbt/dbt-cli/...`, `/plugins/plugin-serdes/csv/...` |
| single-subgroup plugin, segment stripped | `/plugins/plugin-git/...`, `/plugins/plugin-tencent/...`, `/plugins/plugin-email/...` |

For the same blueprint, production today emits `/plugins/io.kestra.plugin.aws.s3.Trigger` where this branch emits `/plugins/plugin-aws/aws-s3/io.kestra.plugin.aws.s3.trigger`.

`npm run test:unit` passes (28 tests), including `buildTaskUrls` against a mocked API for the resolve, omit, API-failure and empty-input paths.


iitzIrFan (approved):
LGTM!

PR #5246, opened 2026-07-29
fix(plugins): link index cards and the fallback redirect at canonical URLs

vfanucci:
Third of three. **Stacked on #5245** (which is stacked on #5244) — this diff is only the index and redirect changes. Retarget to `main` once the other two merge.

### Index cards → 15 URLs, 561 inlinks

`index.astro:21-25` deliberately keeps the **subgroup** entry when a plugin exposes exactly one (it has the richer title and icon), and `PluginCard` appends the subgroup slug whenever `subGroupTitle` is set. So those cards linked `/plugins/plugin-airtable/airtable-record` — a 301 to the plugin root.

A card sees one plugin and cannot know how many subgroups its plugin has, so the page resolves the href with the url index and `prunePluginsForCards` carries it. With no index supplied the field stays unset and `PluginCard` falls back to its existing logic, which leaves `SimilarPlugins` untouched.

Worth noting on the inlink count: those 561 are inflated by the 123 indexable `/plugins?page=` and `?category=` variants, each carrying the full card list. Capping that is a separate fix — this one makes the link itself correct.

### Fallback redirect → 16 URLs, 277 inlinks

For a URL with a single slug segment, `splitRouteSlug.slice(0, -1).join("/")` is empty, so an unknown plugin was redirected to `"/plugins/"` and the trailing-slash rule bounced it a second time. `/plugins/plugin-notifications` took two hops to reach `/plugins`.

## Verification

Dev server on real API data:

- the index serialises **563 card hrefs into the island props — all 563 answer 200** in production. The same page today emits subgroup links that redirect; `plugin-airtable/airtable-record`, `plugin-argocd/argo-cd-apps`, `plugin-ansible/ansible-cli` and `plugin-airflow/airflow-dags` are present in the production HTML and absent from this branch's.
- `/plugins/plugin-notifications` now returns `Location: /plugins` directly instead of `/plugins/`.

`npm run test:unit` passes (32 tests).

## Where this leaves the /plugins redirect debt

Across the three PRs, ~2,900 internal inlink hops removed and 124 sitemap URLs made canonical. What remains is upstream of this repo: plugin groups that no longer exist (`plugin-notifications`, `plugin-core`) are still linked from somewhere I could not pin down in the crawl data — they now cost one hop instead of two, and the catch-all in `plugins.yml` keeps them landing on `/plugins`.


iitzIrFan (commented):
- Astro check: remove parentSlug and inline the redirect path:
- SimilarPlugins: make urlIndex required, pass it from SimilarPlugins.astro and make PluginCard use the canonical href only. Remove the old subGroupTitle/slugify fallback.

iitzIrFan:
Hey @elevatebart could you verify these two above points? Thanks!

---

Edit: Some context of PR's,
https://github.com/kestra-io/docs/pull/5244 = build the canonical URL resolution logic
https://github.com/kestra-io/docs/pull/5245 = use it for Blueprint task links
https://github.com/kestra-io/docs/pull/5246 = use it for Plugin cards + fix a separate redirect

---

elevatebart (approved):

PR #5250, opened 2026-07-30
fix(docs): clarify behaviour of required input

yuri1969:
Also migrate `BOOLEAN` to post-v0.23.


aj-emerich (approved):
LGTM - thanks for catching. Something must've gotten mixed up with BOOL that this went back to BOOLEAN.

PR #5255, opened 2026-07-30
fix(seo): drop non-indexable URLs from the sitemaps

vfanucci:
Found while diffing a full Screaming Frog Spider crawl against a crawl of our own sitemaps (July 2026).

**Sitemap-only change** — the pages themselves are untouched, the diff is limited to `src/pages/sitemap/`.

## Problem

Eight URLs we submit to Google resolve to something Google is told not to index **at that address**. Each one spends crawl budget only to be discarded.

| URL | Issue |
|---|---|
| `/data-processing-agreement` | `noindex` |
| `/enterprise-license-agreement` | `noindex` |
| `/kestra-cloud-tos` | `noindex` |
| `/master-software-licence-agreement` | `noindex` |
| `/use-cases/stories` | 301 → `/customers` |
| `/videos` | canonical → `/videos/all` |
| `/blueprints/azure-blob-backup-retention-cutover` | 500 |
| `/blueprints/ldap-employee-offboarding-deprovision` | 500 |

## Changes

**`misc.xml.ts`** — the 4 legal pages already declare their own `noindex` with a `<meta>` tag in the markdown body. The route reads that directive back instead of keeping a second list, so a page and its sitemap entry can't disagree: adding or removing `noindex` on a legal page is enough.

**`default.xml.ts` / `use-cases.xml.ts`** — `/videos` (rewrites to `/videos/all` and canonicalises there) and `/use-cases/stories` (301s to `/customers`) are excluded from the `.astro` globs via a named, commented constant.

**`blueprints.xml.ts`** — the 2 blueprints are listed by the API and their `/blueprints/{id}/versions/latest` payload is fine, but `/blueprints/{id}/versions/latest/graph` answers `422` and [`[id].astro`](../blob/main/src/pages/blueprints/%5Bid%5D.astro#L137) fetches the graph outside the `try`/`catch` just above it, so the render throws. Nothing in the list payload distinguishes them from a healthy blueprint, so they are named explicitly with a note to remove them once the pages return 200.

## Verification

Ran each modified route on a dev server and diffed against the live sitemap. Exactly these 8 URLs disappear and nothing else changes:

| sitemap | prod | after | removed | added |
|---|---|---|---|---|
| `misc.xml` | 7 | 3 | 4 | 0 |
| `default.xml` | 38 | 37 | 1 | 0 |
| `use-cases.xml` | 49 | 48 | 1 | 0 |
| `blueprints.xml` | 482 | 480 | 2 | 0 |

`misc.xml` keeps `/cloud-license-agreement`, `/cookie-policy` and `/privacy-policy` — the legal pages without a `noindex`. All legal pages still render exactly as they do today. `astro check` clean.

## Follow-ups (not in this PR)

- **The blueprint 500 itself.** Wrapping the graph fetch would make both pages render without the widget instead of throwing, and any future blueprint with an ungraphable flow would degrade rather than break. Both pages are internally linked, so crawlers reach them regardless of the sitemap.
- **`Blueprint.id` is typed `number`** in `src/type.d.ts` but the API returns a slug string, hence the `String(r.id)` here.
- **`/videos` has no submitted URL left** after this PR: `/videos/all` and the 6 category pages come from the `[slug]` route, which is filtered out of the glob. Part of a separate "missing from sitemap" batch, along with the 38 `/vs/*` pages and 12 `/docs/changelog/*`.


iitzIrFan (commented):
Now it's good after, https://github.com/kestra-io/docs/pull/5255/changes/ed754fa8b7a989c617392f4df3b4ff80b670fdf1

iitzIrFan (approved):

PR #5256, opened 2026-07-30
fix(seo): stop listing redirecting plugin subgroup URLs in the sitemap

vfanucci:
Found while diffing a full Screaming Frog Spider crawl against a crawl of our own sitemaps (July 2026).

## Problem

`sitemap/plugins.xml` submits 2 387 URLs, and **124 of them return a 301** — every plugin subgroup URL of the form `/plugins/{plugin}/{subgroup}[/{task}]` for plugins that expose a single subgroup (airtable, dropbox, miro, temporal, trello, jira, terraform, …). Worse, the flat URLs they redirect *to* were themselves absent from the sitemap, so the canonical pages were never submitted at all.

The cause is a mismatch between two files:

- [`plugins.xml.ts`](../blob/main/src/pages/sitemap/plugins.xml.ts) emitted the subgroup segment whenever the plugin's subgroup slug appeared *anywhere* in the global `cls -> subgroup` mapping.
- [`plugins/[...slug].astro`](../blob/main/src/pages/plugins/%5B...slug%5D.astro#L110) 301s that segment away when the plugin exposes exactly one subgroup.

## Change

Gate the subgroup segment on the same condition the page route uses: keep it only for plugins exposing **more than one** subgroup (`plugin-aws`, `plugin-ee-dellemc`, …), drop it otherwise. The rule moves into `buildSubGroupSegmentPredicate` in `utils/plugins/plugin.ts` with a doc comment pointing at the page route, plus unit tests, so the two can't drift apart silently.

This also covers a case the page route does *not* redirect: a plugin with **no** subgroup of its own. `plugin-git` shares its classes with `plugin-ee-git`, whose `git` subgroup leaked in through the global mapping, producing `/plugins/plugin-git/git/…` — live, `200`, self-canonical duplicates of the flat pages. They are dropped here too.

## Verification

Ran the route on a dev server and diffed against the live `sitemap/plugins.xml`: **2 387 → 2 350 URLs**, matching a standalone simulation of the algorithm exactly. Every impacted URL was then checked against production:

| | count | live status |
|---|---|---|
| removed | 124 | `301` |
| removed | 3 | `200` — the `plugin-git/git` duplicates |
| **added** | **90** | `200` |

The 90 additions are the flat canonical URLs the redirects pointed at, which were missing from the sitemap. Multi-subgroup plugins are untouched — `plugin-ee-dellemc` and `plugin-aws` keep all 108 of their URLs, subgroup segment included.

`astro check` and `vitest` clean.

## Follow-up (not in this PR)

`plugins/[...slug].astro` redirects when a plugin has exactly one subgroup, but serves `200` when it has none — which is what makes `/plugins/plugin-git/git/io.kestra.plugin.git.namespacesync` a live self-canonical duplicate of `/plugins/plugin-git/io.kestra.plugin.git.namespacesync`. This PR removes them from the sitemap; making the route redirect in that case too would remove the duplicates outright.


iitzIrFan (approved):
LGTM!

PR #5257, opened 2026-07-30
feat(seo): submit the page types missing from the sitemaps

vfanucci:
## What

A crawl of kestra.io found **188 indexable, self-canonical URLs that return 200 but appear in no sitemap we submit**. This PR adds the **82** of them that are whole page types no sitemap route covers — each is built from a dynamic route, and `default.xml` only globs static `.astro` pages, so nothing was ever emitting them.

| Page type | URLs | Where it now comes from |
|---|---|---|
| `/vs/{competitor}` | 38 | new `sitemap/vs.xml` — `vs` collection |
| `/careers/{posting}` | 15 | new `sitemap/careers.xml` — Ashby job list |
| `/docs/changelog/{tag}` | 12 of 100 | `sitemap/docs.xml` — `fetchMajorReleases` |
| `/videos/{category}` | 7 | new `sitemap/videos.xml` |
| `/blueprints/{category}` | 6 | `sitemap/blueprints.xml` — `CATEGORY_TILE_META` |
| `/blogs/{category}` | 3 | `sitemap/blogs.xml` — `allBlogCategories` |
| the home page | 1 | `sitemap/default.xml` |

The **home page** was dropped by `default.xml`'s own `r !== ""` guard: with `trailingSlash: "never"` Astro reduces the route for `src/pages/index.astro` to an empty string, so the filter meant to skip non-routes skipped `/` as well. It now maps the empty route back to a slash.

The crawl found 12 changelog pages (the ones linked from `/docs/changelog`); the route emits all 100 that `getStaticPaths` builds, reusing the same `fetchMajorReleases` limit so the sitemap can't list a release page that wasn't built.

## Not in this PR: the 106 plugin URLs

The remaining URLs from the crawl are all plugin pages, and they belong on top of the plugin work already in flight rather than here:

- **82** are already added by #5256.
- **10** are subgroup-form duplicates of flat URLs the sitemap *already* lists (e.g. `/plugins/plugin-jdbc-postgres/postgresql/…query` vs. the listed `/plugins/plugin-jdbc-postgres/…query`). Both return 200 and both self-canonicalise, so these need a canonical — submitting the second form would just hand Google a duplicate.
- **14** need the subgroup slug itself fixed: the sitemap derives it from `slugify(title)` (`google-cloud-bigquery`, `aws-athena`) while the page serves the last segment of `subGroup` (`bigquery`, `s3`). That mismatch also means #5256 keeps emitting the `slugify(title)` shape for multi-subgroup plugins — worth a look on that PR.

## Keeping page and sitemap in sync

Every listing is enumerated from the same source its page route uses, and the two aliases that canonicalise elsewhere stay out:

- `/blogs/$all` canonicalises to `/blogs` → skipped (`/blogs` is already in `default.xml`).
- `/videos` rewrites to `/videos/all` and canonicalises there → the category slug is listed, `/videos` is not.
- `/blueprints/getting-started` is not a category page — `[id].astro` falls it through to the blueprint lookup → excluded.

The careers URL was built inline in `Positions.vue`. It moves into a shared `jobPath` helper used by both the listing links and the sitemap, so the sitemap cannot describe a URL the page doesn't link.

## Verification

Run against the dev sitemap routes and diffed with the live ones:

- all **67** non-careers URLs from the crawl now appear
- **no** URL the live sitemaps currently list disappears (0 regressions)
- 12 sampled changelog pages return **200**
- careers fall back to the Ashby mock without an API key, so that route was checked to emit exactly the 20 links the rendered page produces from the same fixture (shape verified against the 15 live links)
- `astro check` 0 errors / 0 warnings / 0 hints, `oxlint` clean, `vitest run` 13 passed, `check-astro-content-imports` OK

## Overlap with the other open sitemap PRs

Touches `default.xml.ts` and `blueprints.xml.ts`, which #5255 also touches — different hunks, small conflicts at most. #5255 removes `/videos` as non-canonical; this PR adds `/videos/all`, which is the same conclusion, so the two agree. If #5255 lands first its `NON_CANONICAL_ROUTES` entry for `/videos` simply stays as the thing doing that job.


iitzIrFan (commented):
I took a look at the CI failure and think I have narrowed it down, but I would appreciate a second opinion before making any changes.

It looks like the change in `Positions.vue` now imports `jobPath` from `careers.ts`. Since `careers.ts` imports `astro:env/server`, and `Positions.vue` is hydrated with `client:load`, it seems that server-only module is ending up in the client bundle, which matches the `astro:env/server` build error in CI.

Would moving `jobPath` into a separate client-safe utility (without the server env dependency) be the right approach here, or is there another pattern we would prefer?

@elevatebart, could you take a look when you have a chance? I would appreciate your thoughts before I make any changes.

iitzIrFan (commented):
@vfanucci  could you please resolve the merge conflicts when you get a chance? Thanks!

vfanucci:
@iitzIrFan Your diagnosis was exactly right — thanks for digging into it, and sorry for sitting on the conflicts. Both are handled in f8e5091.

### The build failure

`Positions.vue` is hydrated `client:load`, and this branch changed its import from `import type { AshbyJob }` (erased at build) to a **value** import of `jobPath`. That pulled `utils/careers.ts` into the client bundle, and that module reads `ASHBY_APIKEY` from `astro:env/server` at the top level for its `headers` const — hence `ServerOnlyModule`.

So yes, your suggestion is the right shape and that's what I did: `jobPath` now lives in `src/utils/careersPath.ts`, which depends only on `slugify` and the `AshbyJob` *type*. One thing worth recording for next time: moving the env read into a function would **not** have been enough — it's the `astro:env/server` *import* that the Vite plugin rejects on the client, not the access to the value.

The listing links and the sitemap still share the single helper, so they still can't describe different URLs.

While in there, `careers.xml.ts` now builds its list with `flatMap`: `.filter(p => p !== null)` doesn't narrow `string | null` in TypeScript, so the null case was still in the type.

### Conflicts

Merged `main` in:

- **`blueprints.xml.ts`** — kept both sides. `main`'s `BROKEN_BLUEPRINTS` filter and this branch's category listings are independent, so the categories keep their prefix position and the blueprint entries keep the filter.
- **`default.xml.ts`** — merged cleanly: `main`'s `NON_CANONICAL_ROUTES` set and this branch's home-page fix touch different steps of the same chain. #5255 dropping `/videos` and this PR adding `/videos/all` are the same conclusion, as expected.

### Verified locally on the merge result

- `npm run build` passes — no `ServerOnlyModule`
- `astro check` 0 errors / 0 warnings / 0 hints, `oxlint` clean, `vitest run` 43 passed
- `/sitemap/careers.xml` → the 20 links from the Ashby mock (no API key locally), i.e. exactly what the rendered listing produces from the same fixture
- `/sitemap/blueprints.xml` → 6 category URLs + 480 blueprints, `getting-started` excluded, both broken blueprints still filtered
- `/sitemap/default.xml` → contains `https://kestra.io/`, and no `/videos`
- `vs.xml` 38 · `videos.xml` 7 · `blogs.xml` 3 categories · `docs.xml` 100 changelog URLs

CI is running now. Happy to squash the two commits if you'd prefer a cleaner history.

elevatebart (approved):
Great job to you both !!!!

PR #5267, opened 2026-08-03
fix(seo): 301 subgroup-less plugin element URLs instead of serving duplicates

### LINKED ISSUE #5266: fix(plugins): non-canonical plugin element URLs serve duplicate content instead of 301
### Describe the issue

Every plugin element that lives under a subgroup is reachable at two indexable URLs, both returning 200 with a self-referencing rel="canonical". This is genuine duplicate content, not a redirect chain, Google indexes both and splits ranking signals between them.

The subgroup-less form is meant to 301 to the canonical form. The redirect logic exists but has never fired, for any plugin.

---

Reproduction:

Non-canonical form — currently 200, expected 301:
curl -sS -o /dev/null -w '%{http_code} %{redirect_url}\n' 'https://kestra.io/plugins/plugin-slack/io.kestra.plugin.slack.notifications.slackincomingwebhook'

Canonical form — correctly 200:
curl -sS -o /dev/null -w '%{http_code} %{redirect_url}\n' 'https://kestra.io/plugins/plugin-slack/slack-notifications/io.kestra.plugin.slack.notifications.slackincomingwebhook'

Both emit <link rel="canonical"> pointing at themselves, so neither defers to the other.

---

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5266

---

### DIFF
--- a/src/pages/plugins/[...slug].astro
+++ b/src/pages/plugins/[...slug].astro
@@ -164,8 +164,11 @@ if (!versionState.version) {
         )
         if (mappedUrl && mappedUrl !== pathname)
             return Astro.redirect(mappedUrl, 301)
-    } catch {
-        // ignore
+    } catch (e) {
+        // Fall through to the render path, which handles a genuinely unknown element. Logged
+        // because this catch silently hid a broken canonicalisation for every plugin: redirectAlias
+        // compared a lowercase page list against a real-cased FQCN, so it always threw here.
+        console.warn(`redirectAlias failed for ${pathname}:`, (e as Error)?.message ?? e)
     }
 }
 
--- a/src/utils/plugins/aliasMapping.test.ts
+++ b/src/utils/plugins/aliasMapping.test.ts
@@ -1,0 +1,129 @@
+import { describe, expect, it } from "vitest"
+import { getAliasMapping, redirectAlias } from "~/utils/plugins/aliasMapping"
+import { generateNavigationFromSubgroups } from "~/utils/plugins/generateNavigation"
+import { recursivePages } from "~/utils/navigation"
+import type { Plugin } from "~/utils/plugins/plugin"
+
+// Trimmed shape of GET /plugins/subgroups for plugin-slack: a root entry listing every element
+// plus one entry per subgroup. With more than one subgroup, the canonical element URL carries the
+// subgroup segment (/plugins/plugin-slack/slack-notifications/<fqcn>).
+const slackPlugins = [
+    {
+        name: "plugin-slack",
+        group: "io.kestra.plugin.slack",
+        title: "Slack",
+        aliases: ["io.kestra.plugin.slack.SlackIncomingWebhook"],
+        tasks: [
+            { cls: "io.kestra.plugin.slack.notifications.SlackIncomingWebhook" },
+            { cls: "io.kestra.plugin.slack.app.chats.Post" },
+        ],
+    },
+    {
+        name: "plugin-slack",
+        group: "io.kestra.plugin.slack",
+        subGroup: "io.kestra.plugin.slack.notifications",
+        title: "Slack Notifications",
+        tasks: [{ cls: "io.kestra.plugin.slack.notifications.SlackIncomingWebhook" }],
+    },
+    {
+        name: "plugin-slack",
+        group: "io.kestra.plugin.slack",
+        subGroup: "io.kestra.plugin.slack.app.chats",
+        title: "Slack Chats",
+        tasks: [{ cls: "io.kestra.plugin.slack.app.chats.Post" }],
+    },
+] as unknown as Plugin[]
+
+const pageList = recursivePages(generateNavigationFromSubgroups(slackPlugins)[0])
+const aliasMapping = getAliasMapping(slackPlugins)
+
+const CANONICAL =
+    "/plugins/plugin-slack/slack-notifications/io.kestra.plugin.slack.notifications.slackincomingwebhook"
+
+// Canonical slug is slugify(title) — "slack-chats" — while the subGroup's own last segment is
+// "chats". matchesSubGroup() accepts both, which is why the segment form renders instead of 404ing.
+const CHATS_CANONICAL =
+    "/plugins/plugin-slack/slack-chats/io.kestra.plugin.slack.app.chats.post"
+const CHATS_FQCN = "io.kestra.plugin.slack.app.chats.Post"
+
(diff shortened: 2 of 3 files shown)


vfanucci (approved):
Approving — the fix is correct, minimal, and catches more than it advertises. Verified locally on the branch (unit tests + dev server against the real `api.kestra.io`). Three non-blocking comments below, plus the non-regression evidence so it doesn't have to be re-derived.

Root cause confirmed exactly as described: `pageList` entries come from `slugify(item.cls)` (lowercase) while `pluginType` is rebuilt with the FQCN's real casing, so the case-sensitive `endsWith` never matched — for any plugin, ever. The alias branch right below it already lowercased both sides, which is why legacy alias URLs have always 301'd correctly while canonicalisation never did.

---

### 1. Please add a test for the segment form

This is the only comment I'd suggest addressing before merge.

The tests cover the flat form (`/plugins/plugin-slack/<fqcn>`). They don't cover the other non-canonical shape: `/plugins/plugin-aws/s3/<fqcn>`, where the URL carries the `subGroup`'s last segment instead of `slugify(title)`. `matchesSubGroup` explicitly accepts that segment, so it also returns 200 today.

That's the shape that is **measurably** splitting signals right now. Six elements currently rank on two URLs simultaneously, and on two of them the duplicate outranks the canonical:

| element | canonical | duplicate |
|---|---|---|
| `slack.app.chats.StopStream` | `slack-chats` — 386 clicks / 10,022 impr. | `chats` — 46 clicks / 1,532 impr. |
| `googleworkspace.drive.List` | `google-drive` — 1,138 impr. | `drive` — **1,636 impr.** |
| `shopify.orders.List` | `shopify-orders` — 191 impr. | `shopify-order` — **1,124 impr.** |
| `aws.s3.Upload` | `aws-s3` — 141 impr. | `s3` — 383 impr. |
| `aws.emr.SubmitSteps` | `aws-emr` — 51 impr. | `emr` — 99 impr. |
| `dbt.cli.DbtCLI` | `dbt-cli` — 363 impr. | `cli` — 2 impr. |

The fix does handle it — I verified `/plugins/plugin-aws/s3/io.kestra.plugin.aws.s3.upload` → 301 → `/plugins/plugin-aws/aws-s3/io.kestra.plugin.aws.s3.upload` on the dev server. But nothing pins it, so a later change to `matchesSubGroup` or to the subgroup slug source could silently regress the case that actually costs us traffic. Happy to hand over the fixture.

### 2. Worth noting the `.md` side effect in the description

Plugin element URLs with a `.md` suffix go from `200` to `301`. That's a **gain**, not a risk: `src/pages/plugins/` has no `.md` route (unlike `docs`, `blogs` and `resources`, which each have a dedicated `[...].md.ts`), so the suffix was silently absorbed by the catch-all and rendered the full HTML page — `content-type: text/html`, indexable, with a self-referencing canonical that *included* the `.md`. A fourth duplicate family, removed for free.

It's just outside the scope the title advertises, so a reviewer testing `.md` should know it's intentional.

### 3. The commit message understates the impact by ~5×

"1309 URLs across 58 plugins" matches the flat form alone. The actual total moving from 200 to 301:

| shape | URLs |
|---|---|
| flat — `/plugins/<plugin>/<fqcn>` | 1,306 |
| segment — `/plugins/<plugin>/<subgroup-segment>/<fqcn>` | 1,101 |
| `.md` on each of the above + on the canonical | 4,232 |
| **total** | **6,639** |

Canonical element pages staying at 200: 1,825. So on this surface we go from 8,464 indexable URLs to 1,825, and from 0 redirects to 6,639. Worth correcting because that number is the reference for interpreting the jump in "Page with redirect" in Search Console after deploy.

### Nit

The empty `catch {}` in `src/pages/plugins/[...slug].astro:167` stays, and one of the new tests now pins the throw as intentional (`"throws when no page matches, so the caller can fall through"`). Defensible — but that silence is exactly why this went unnoticed. A `console.warn` would cost nothing and would surface the next occurrence.

---

### Non-regression evidence

- **Tests**: 5/5 on the new file, 35/35 on the full suite. `tsc --noEmit` reports 2 errors in `src/utils/getNavigationTree.ts`, both pre-existing on `main` and unrelated.
- **Ranking URLs**: of 106 element URLs with Search Console data, 95 unchanged, 11 start redirecting — all 11 are genuine duplicates. The two pages with real traffic that had to stay at 200 do: `/plugins/core/flow/io.kestra.plugin.core.flow.foreach` (23 clicks) and `/plugins/plugin-telegram/io.kestra.plugin.telegram.telegramexecution` (90 clicks — single-subgroup plugin, where the flat form *is* canonical and the reverse redirect at `[...slug].astro:113` applies).
- **Sitemap**: 1,828 element URLs, emitted with `cls.toLocaleLowerCase()` while the nav uses `slugify(cls)`. Those agree on **1826/1826** classes, so no sitemap URL becomes a redirect. Spot-checked 8 sampled entries → all 200.
- **No chains, no ambiguous targets**: every 301 reaches a 200 in a single hop, including doubly non-canonical URLs (`/plugins/plugin-aws/io.kestra.plugin.aws.s3.upload.md` → one 301 straight to the canonical, not two). 0 subgroup slugs containing a dot, and 0 class reachable from more than one subgroup — so the `.find()` target is deterministic and can't flap with API ordering.
- **Internal links**: 0 of the 96 flat-form plugin links in `src/contents/` point at a non-canonical URL, so no content cleanup is needed.

### Post-merge

Expect "Page with redirect" to jump by ~6,600 on this section — that's the fix working, not an alert. The signal to actually watch is consolidation of the six pairs above, primarily `slack.app.chats.StopStream` (386 + 46 clicks to merge onto one URL).

### Out of scope, tracking separately

1. `/plugins/core/core-app-block/io.kestra.plugin.ee.apps.core.blocks.logs` — 240 impressions, class absent from `/plugins/subgroups`, absent from the sitemap, served as 200 before **and** after this PR. Orphan page, not a regression here.
2. No markdown endpoint for plugin pages. `docs`, `blogs` and `resources` each have one; `plugins` doesn't. After this PR an agent following the `<url>.md` convention gets a 301 to HTML. It's the largest section of the site with no markdown output, and element pages are precisely what LLMs go looking for.

Search Console figures above come from the Ahrefs GSC connector over the top 250 `/plugins/` URLs, May–Jul 2026 — reliable enough to detect the duplicate pairs, not to size the totals precisely.

vfanucci (approved):
LGTM

PR #5285, opened 2026-08-06
chore: Applied a backdrop blur effect to the top-navbar 

### LINKED ISSUE #5279: Apply a backdrop filter to the top-navbar for better contrast and legibility

m-t-a97:
This PR applies a nice backdrop blur effect to the top-navbar for better contrast and legibility.

Before:
image

After:
image

---

Closes #5279


m-t-a97:
Feel free to let me know if you want the blur to be stronger e.g. `blur(8px)` as it is currently at `4px`.

Some more examples with `4px`:

image

image

image

---

This is with `8px` blur applied:

image

image

image

iitzIrFan (commented):
Thanks for the contribution! I tested this locally and noticed that `&.navbar `applies in every navbar state, so it ends up overriding the existing `.scrolled/.open backdrop blur`. I think we should scope this to the unscrolled, closed state instead to preserve the current behavior while fixing the top-of-page gap!

m-t-a97:
> Thanks for the contribution! I tested this locally and noticed that `&.navbar `applies in every navbar state, so it ends up overriding the existing `.scrolled/.open backdrop blur`. I think we should scope this to the unscrolled, closed state instead to preserve the current behavior while fixing the top-of-page gap!

Yes you're right, with the current change I made, it was overriding the original `$menu-backdrop-filter` which is a blur of `44px`. However, you're saying to keep it scoped to the unscrolled state. That means you're not going to get the effect outlined in this issue because currently the blur is at `44px` when in an opened and scrolled state due to the `$menu-backdrop-filter`. The minute you start scrolling, the current `$menu-backdrop-filter` SASS variable is being applied.

So either we modify `$menu-backdrop-filter` to be `4px` to get this effect in an opened scrolled state or you have an alternative suggestion? Also not to mention that `$menu-backdrop-filter` is only being used in this scenario in the header within the opened scrolled state so I'm leaning towards modifying that SASS var to be 4px if we want this subtle blur effect on the top-navbar throughout the docs site.

Furthermore,  you mentioned "while fixing the top-of-page gap", can you provide a screenshot of that so I can see what you mean?  

Let me know if I'm misunderstanding anything here.

Piyush-r-bhaskar (commented):
Hey @m-t-a97 

It only broke in production because the CSS minifier merges the two `backdrop-filter `lines and keeps just the last one — in `Header.vue` the `-webkit-` version was last, so that's all production shipped, and browsers didn't apply the blur. [This PR](https://github.com/kestra-io/docs/pull/5287) swaps the order so the standard `backdrop-filter` line is kept, which will fix the blur in production.

Piyush-r-bhaskar:
Closing in favour of https://github.com/kestra-io/docs/pull/5287 , Thanks for your efforts 🍀

PR #5297, opened 2026-08-07
fix(plugins): use subgroup title when element name duplicates container title

iitzIrFan:
Ref: https://kestra-io.slack.com/archives/C052XEFFC90/p1786085765167359

---


Piyush-r-bhaskar (changes requested):
Hey @iitzIrFan 

I don't understand the changes here. why we need to construct the title and do surgical changes. why not just change the title at Plugin/API level ?

Piyush-r-bhaskar (commented):
This will create confusion where by Opening an element `Kubernetes` shows `Kubernetes Runner` title. instead I would prefer a change in title at element level.

Image

Image

iitzIrFan (commented):
Made the change after confirming with Alex that the card and page title should stay aligned.

Piyush-r-bhaskar (approved):

PR #5298, opened 2026-08-07
fix(ci): fail the contributor onboarding workflow when the welcome action errors

### LINKED ISSUE #17940: Contributor Onboarding workflow reports success while welcoming nobody (GH_READ_ORG_TOKEN returns 401)
The `Contributor Onboarding` workflow (`.github/workflows/welcome.yml`) has not welcomed a single first-time contributor for weeks, while reporting a green ✅ on every run.

## Symptom

Every run of the workflow ends with:

```
Checking if @<author> is an org member of `kestra-io`.
Org membership check returned status: 401
##[error]Bad credentials - https://docs.github.com/rest
```

…and the job still concludes `success`.

Example: https://github.com/kestra-io/kestra/actions/runs/31000349981/job/92287429768

## Root cause

Two things combine:

1. **The credential is dead.** `GH_PAT_READ_ORG` is fed from the org-level secret `GH_READ_ORG_TOKEN`. It resolves (it is masked as `***` in the logs, so it is set and non-empty) but GitHub rejects it with `401 Bad credentials`. The secret was last updated 3 months ago, which matches a classic PAT created on the default 90-day expiry window.

2. **The failure is swallowed.** The `fail-on-error` input of `plbstl/first-contribution` defaults to `false`. In that mode the action catches every error and calls `core.error()` instead of `core.setFailed()`, so the step still exits 0.

Because we pass `skip-internal-contributors: true`, the org-membership check is the *first* thing the action does — and `is_internal_contributor()` only swallows a `404`; any other status is re-thrown. So the expired token aborts the action before `add_reactions` and `create_comment` ever run, and the job goes green anyway.

## Impact

- No first-time contributor has received the welcome comment or the reactions.
- Confirmed on https://github.com/kestra-io/kestra/pull/17865 — there is no bot comment on it; the contributor was welcomed by hand.
- The oldest run still in the log-retention window (2026-07-21) already fails this way, so the outage is at least two and a half weeks old and probably longer.
- **`kestra-io/docs` is affected identically** — it runs a copy of the same workflow against the same secret.

## Fix

1. **Rotate the credential** (org admin) — regenerate a PAT with `read:org` scope at https://github.com/settings/tokens/new?scopes=read:org and update the `GH_READ_ORG_TOKEN` org secret. Setting it to **no expiration** avoids a silent repeat in another 90 days. Being an org-level secret, one rotation covers both repos.

2. **Make the failure loud** — set `fail-on-error: true` so a broken credential surfaces as a red check instead of a green one:
   - https://github.com/kestra-io/kestra/pull/17939
   - https://github.com/kestra-io/docs/pull/5298

## Verification

After the token is rotated and both PRs are merged, the next external PR is the test: a green `welcome` job **with** a bot comment means it genuinely works; a red job means the credential is still wrong — which is the point of the change.

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/17940.

The Contributor Onboarding workflow has been reporting ✅ while welcoming nobody.

## What

Set `fail-on-error: true` on the `plbstl/first-contribution` step so an error surfaces as a red check instead of a green one.

## Why

`fail-on-error` defaults to `false`. In that mode the action catches every error and calls `core.error()` rather than `core.setFailed()`, so the step still exits 0.

Because we pass `skip-internal-contributors: true`, the very first thing the action does is an org-membership check — and `is_internal_contributor()` only swallows a `404`; anything else is re-thrown. So an invalid `GH_READ_ORG_TOKEN` aborts the action *before* it can react or comment, and the job still goes green.

That is the current state. Recent runs fail identically:

```
Checking if @<author> is an org member of `kestra-io`.
Org membership check returned status: 401
##[error]Bad credentials - https://docs.github.com/rest
```

## What this does *not* fix

The credential itself — `GH_READ_ORG_TOKEN` is an org-level secret whose PAT has expired or been revoked, and an org admin still needs to rotate it. This PR only ensures we find out immediately next time, rather than weeks later.

## Companion PR

https://github.com/kestra-io/kestra/pull/17939 — same one-line change, same failure.

### DIFF
--- a/.github/workflows/welcome.yml
+++ b/.github/workflows/welcome.yml
@@ -21,6 +21,8 @@ jobs:
           GH_PAT_READ_ORG: ${{ secrets.GH_READ_ORG_TOKEN }}
         with:
           skip-internal-contributors: true
+          # The action defaults to false, which turns a broken GH_READ_ORG_TOKEN into a green job that silently welcomes nobody.
+          fail-on-error: true
           pr-opened-msg: |
             ### Hey @{fc-author} :wave: 
 


MilosPaunovic:
Hey @iitzIrFan, there is no real need to assign people to the PR, as the same person is usually the PR author.

iitzIrFan:
Noted @MilosPaunovic

MilosPaunovic:
So if you can only give it a ✅ I'd appreciate it 🙂

iitzIrFan (approved):
LGTM!

iitzIrFan:
Done @MilosPaunovic 🙌

PR #5334, opened 2026-08-18
Build-time live counters: plugins + GitHub stars (fixes 0+ and count drift)

vfanucci:
Draft — two points below need a call from the team before this is ready.

## The bug

`/enterprise` and `/1-0` shipped **`0+ Plugins`** in their static HTML. Confirmed live on 18 Aug 2026 (`curl https://kestra.io/enterprise` returns both `0+ Plugins` and `1900+ Plugins` on the same page).

The cause is not a failed fetch. `usePluginsCount()` fetches from a `.then()` on the **client**; `Dashboard.vue` (via `Hero.vue`, `client:load`) renders server-side while the ref is still `[]`, and the computed returns the literal `"0+"`. So it shipped on **every** build, with the API perfectly healthy — a guarded fetch or a fallback alone would not have fixed it. The fix is to stop fetching in the component and pass the value down from the Astro page.

`/1-0` had the same bug twice: `onezero/Header.vue` and `AiCopilot.vue` (`0+ Plugins & Safe Upgrades`).

## What this adds

`src/data/counts.ts` resolves each published figure while the page renders and inlines it into the HTML:

- **Committed fallback** — `src/data/counts.fallback.json` covers a failed fetch, so a network blip can no longer zero a published figure.
- **Build assertion** — a fetched-but-implausible plugin count throws rather than ships. It lives on the plugin accessor, not the shared resolver, so a bad plugin response cannot 500 on-demand pages (`/blueprints`) that only read a different counter.
- **Asymmetric floors** — plugins fail hard (`api.kestra.io` is ours and must answer); stars only warn. `api.github.com` rate-limits anonymous callers at 60/hour per IP, and `DISABLE_GITHUB` is deliberately set in `update-linux-snapshots.yml` — neither should be able to block a deploy.
- **Independent memos** — `/plugins/subgroups` is a ~3.4 MB payload; a page showing only a star count must not pull it.

Seven files had reimplemented `Math.floor(calculateTotalPlugins(...) / 100) * 100` with their own unguarded fetch, and `EveryPlugin.astro` defaulted to `"0+"` on failure. They now share the module, which also drops the per-page refetch.

The star count is rendered **server-side** in the header for the first time (the client component becomes a refresher), which also removes one `/api/github` request per page view.

## ⚠️ Point 1 for the team: the counting method

The brief asked for "total sub-plugins across all groups" and flagged that it yields 1,884 against a site claiming 1900+. Both numbers are real, they are just different methods:

| Method | Result | Display |
|---|---|---|
| Distinct plugin classes — **what the site already does** | **1,906** | `1900+` |
| Sum of the 9 element arrays — the brief's method | **1,884** | `1800+` |
| Sum of all 12 element arrays (incl. `aliases`, `appBlocks`, `dataFiltersKPI`) | 1,908 | `1900+` |

So `1900+` was **not** a hardcoded stale figure — `grep -rE "1,?900" src/pages src/components src/utils` returns nothing. It is already live and already true. Adopting the brief's method would have **lowered** a correct public claim to `1800+`.

**This PR keeps the existing method** (per @vfanucci). If the team prefers the stricter count, it is a one-line change in `getPluginCountRounded`.

Rounding stays **down**, never up: the figure carries a `+`, so flooring is what keeps the claim true as the real count drifts between builds.

## ⚠️ Point 2 for the team: editorial copy is not in this PR

The brief described "two /vs card descriptors". The real footprint of hardcoded plugin figures:

| Location | Occurrences |
|---|---|
| `src/contents/vs/*.yaml` | 61 |
| `src/contents/resources/**` | ~25 |
| `src/contents/orchestration/*.yaml` | 14 |
| blog titles / video titles (`1300+`, `1600+`) | dated content, leave alone |

These are rendered sentences and FAQ answers ("Kestra ships with 1400+ plugins covering dbt, Airbyte…"), not hero stats — templating indexed prose is a different job with a different review. Scoped to a **follow-up content PR** by decision; the inventory above is the worklist.

Also still client-side: `src/components/content/PluginCount.vue`, used as `:PluginCount` inside blog markdown. Its default no longer renders a zero (it falls back to the committed value), but feeding it a real build-time value means plumbing through the MDC renderer — same follow-up.

## Bonus: the blueprints count (brief point 7)

A blueprints endpoint **does** exist and is already wired live: `fetchTotalBlueprintsCount()` hits `/blueprints/versions/latest?size=1&page=1`. So `480+` was not hardcoded either — but it carried the same `catch → "0"` bug, which would have rendered `0+ Blueprints, ready to build.` as an H1. Hardened here via the same fallback.

## Verified

Full `astro build`, then grepped `dist/`. Note the brief's suggested grep (`"0+ Plugins"`) matches by substring — it hits `1300+ Plugins`, `1600+ Plugins`, `1,200+ Plugins` — so it can never reach zero. Anchored instead:

```
grep -rlE '(^|[^0-9])0\+ ?[Pp]lugins' dist/     -> 0 files
grep -rlE '(^|[^0-9])0\+ ?[Bb]lueprints' dist/  -> 0 files
```

Same value everywhere it is templated:

| Page | Rendered |
|---|---|
| `/` | `1900+ Plugins`, `1900+ plugins` |
| `/enterprise` | `1900+ Plugins` (was `0+`) |
| `/1-0` | `1900+ Plugins` (was `0+` ×2) |
| `/pricing` | `1900+ Plugins` (was `1,200+`) |
| `/overview` | `1900+ plugins` |
| `/vs` | `1900+ plugins` |
| `/features/code-in-any-language` | `1900+ integrations` |
| JSON-LD `SoftwareApplication` | `language-agnostic, 1900+ plugins` (was `1400+`) |

Header stars now in static HTML: `<span>27,850</span>`.

Both failure paths exercised by temporarily patching the resolver:

| Simulated | Result |
|---|---|
| API returns 200 with `[]` | **build fails** — `[counts] plugin count came back as 0, below the 1000 floor. Refusing to build…` |
| Fetch fails (404) | build succeeds, `[counts] using committed fallback plugin count (1906, as of 2026-08-18)` |

`npm run test:unit` 43/43 pass, `oxlint` clean on all touched files.

The `1400+ plugins` still visible on `/vs` comes from the content YAML descriptors — expected, that is the follow-up PR.

### DIFF
--- a/src/components/common/SchemaSoftwareApplication.astro
+++ b/src/components/common/SchemaSoftwareApplication.astro
@@ -1,4 +1,6 @@
 ---
+import { fetchTotalPluginsCount } from "~/utils/plugins/pluginCount"
+
 interface Props {
     id?: string
     name?: string
@@ -9,6 +11,8 @@ interface Props {
 }
 
 const { id, name, description, url, isVariantOf, featureList } = Astro.props
+
+const totalPlugins = await fetchTotalPluginsCount()
 
 const canonicalUrl =
     url ?? new URL(Astro.url.pathname, Astro.site ?? "https://kestra.io").toString().replace(/\/$/, "")
@@ -21,7 +25,7 @@ const schema: Record<string, unknown> = {
     "url": canonicalUrl,
     "description":
         description ??
-        "Open-source workflow orchestration platform for data pipelines, infrastructure automation, and AI workflows. YAML-native, language-agnostic, 1400+ plugins.",
+        `Open-source workflow orchestration platform for data pipelines, infrastructure automation, and AI workflows. YAML-native, language-agnostic, ${totalPlugins}+ plugins.`,
     "applicationCategory": "DeveloperApplication",
     "applicationSubCategory": "Workflow Orchestration",
     "operatingSystem": "Linux, macOS, Windows",
--- a/src/components/enterprise/Dashboard.vue
+++ b/src/components/enterprise/Dashboard.vue
@@ -61,8 +61,10 @@
     import Plus from 'vue-material-design-icons/PlusThick.vue';
     import CheckBold from 'vue-material-design-icons/CheckBold.vue';
 
-    import { usePluginsCount } from '~/composables/usePluginsCount';
-    const { totalPlugins } = usePluginsCount();
+    // Comes in from the Astro page, resolved at build time. Fetching it here
+    // (as usePluginsCount does) left "0+ Plugins" in the static HTML, since the
+    // fetch only resolves after hydration.
+    const props = defineProps<{ totalPlugins: string }>();
 
     const COLUMNS = [
         {
@@ -102,7 +104,7 @@
 
     const BOTTOM_FEATURES = computed(() => [
         "Declarative Workflow",
-        `${totalPlugins.value} Plugins`,
+        `${props.totalPlugins} Plugins`,
         "Event Driven & Scheduling",
         "Everything From the UI",
         "Business Logic In Any Language",
--- a/src/components/enterprise/Hero.vue
+++ b/src/components/enterprise/Hero.vue
@@ -20,7 +20,7 @@
                     />
                 </div>
             </div>
-            <Dashboard />
+            <Dashboard :total-plugins="totalPlugins" />
             <div class="logos-wrap">
                 <div class="logos">
                     <div
@@ -43,6 +43,9 @@
 <script setup lang="ts">
     import Link from '~/components/common/Link.vue';
     import Dashboard from '~/components/enterprise/Dashboard.vue';
+
+    defineProps<{ totalPlugins: string }>();
+
     import fortune500Logo from '~/components/enterprise/assets/fortune500.svg';
 
     const logos = import.meta.glob("/src/components/enterprise/assets/logos/*.svg", { eager: true });
--- a/src/components/features/core/EveryPlugin.astro
+++ b/src/components/features/core/EveryPlugin.astro
@@ -2,7 +2,7 @@
 import TaskIcon from "~/components/common/TaskIcon.vue"
 import { Icon } from "astro-icon/components"
 import { $fetchApiCached } from "~/utils/fetch"
-import { calculateTotalPlugins } from "~/composables/usePluginsCount"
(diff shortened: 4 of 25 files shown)


vfanucci:
Hello @iitzIrFan 

It's a draft and I don't know the good option on this one, let talk about it

iitzIrFan:
We have this in the pipeline :) ,
https://github.com/kestra-io/docs/pull/5446

---

Edit: 

What remains is the GitHub Stars part, which is not covered by #5446, so we can keep that as a separate follow-up.

---

vfanucci:
> We have this in the pipeline :) , #5446
> 
> Edit:
> 
> What remains is the GitHub Stars part, which is not covered by #5446, so we can keep that as a separate follow-up.

Thanks, I am following y #5446 

For Github stars, it's not a priority

iitzIrFan:
Closing as superseded by #5446, which landed on 17 Sept.

**Covered by #5446**
- The `0+ Plugins` SSR bug on `/enterprise` and `/1-0` (build-time prop into `Dashboard.vue`, `onezero/Header.vue`, `AiCopilot.vue`), the `0+` fallback in `EveryPlugin.astro`, the second raw fetch in `PlanTable.astro`, and the stale figures on `/pricing` and in the `SoftwareApplication` JSON-LD.
- Shared, memoized count utilities with a floor fallback and separate logging for an empty `200`. #5446 also templated the vs/orchestration YAML hardcodes that this PR deferred to a follow-up.
- The blueprint count fallback, plus the last hardcode on `/data`.

One deliberate difference: this PR throws on an implausible plugin count to fail the build. #5446 falls back to a floor instead, because the throwing variant returned 500 on `/` and `/get-started` during an API blip. Merging this now would reintroduce that.

**Not covered: server-rendered GitHub stars.** Main still renders a placeholder in the header and fetches `/api/github` after hydration. Per the thread above this is not a priority. When it is, a small standalone PR carrying only the four stars files (`GithubButton.vue`, `GithubStargazer.client.vue`, `layout/Header.vue`, `layout.astro`) will be simpler than rebasing this branch, which now conflicts in 18 files. For reference, there are also ten hardcoded star figures in content (six vs YAMLs at `26k+`, four resources articles) against a live count of ~28.1k.

PR #5344, opened 2026-08-19
feat(tracking): emit bookdemo_form_start on first field interaction

vfanucci:
## Why

The demo funnel currently has two points: the landing-page view (`content-view`) and a completed submit (`bookdemo_form`). Nothing in between. So a flat week on `/lp/workflow-orchestration` is ambiguous — is the ad message not landing, or are people starting the form and abandoning it? The form asks for four fields and rejects personal email domains server-side, so abandonment is a real candidate and we currently can't see it.

## What

Adds `bookdemo_form_start`, fired on the first field interaction, to PostHog and the dataLayer.

- `@focusin` on the `<form>` rather than `@input`: someone who tabs in and gives up at "number of employees" has still started the form, and that is the drop-off worth measuring.
- The handler ignores focus on anything that isn't an `input`, `select`, or `textarea`. `focusin` bubbles, and the form also contains the Privacy Policy link and the submit button — without the check, clicking the consent link would register as a form start.
- Guarded by a `started` flag so it fires **once per mounted form instance**: the first field focus emits, and every later focus change inside that same form does not.
- That guard is deliberately per-instance, **not** per-visitor. The `/lp/*` pages mount this component twice (hero + lower section), so one visitor who touches both forms emits two events. The metric should therefore be read as unique users in GA4/PostHog, not as a raw event count.
- Written straight to `window.dataLayer` for the same reason as the existing submit event: the vue-gtm plugin runs with `enabled: false` because GTM is loaded manually after cookie consent in `cookieconsent.ts`, so `gtm.trackEvent()` never reaches the dataLayer. Both events now go through one local `pushDataLayer()` helper instead of repeating the preamble and the explanation twice.
- Wrapped in try/catch, mirroring the submit event — analytics must never break the form.

No autofocus on any field, so the event cannot fire without a real user action.

## Scope

Reporting only. **No Google Ads conversion action is attached to this event** — it exists to give the funnel a denominator between "landed" and "submitted".

The dataLayer push is inert until a GA4 event tag and trigger for `bookdemo_form_start` exist in the GTM container. That container change is tracked separately; PostHog capture works as soon as this merges.

## Test plan

On `/lp/workflow-orchestration` (needs cookie consent accepted, otherwise GTM is not loaded at all):

1. Click into any field of the hero form → one `bookdemo_form_start` in `window.dataLayer`.
2. Click through the remaining fields of that same form → no second event.
3. Click the Privacy Policy link, or focus the submit button, without having touched a field → no event.
4. Scroll to the second form, click into it → one further event (separate mounted instance, expected — see the per-instance note above).
5. Submit → `bookdemo_form` follows as before.
6. Confirm no new Ads conversion pings to `ad.doubleclick.net` / `google.com/ccm` on the start event.

`npx oxlint src/components/demo/Meeting.vue` and `npx astro check` (447 files, 0 errors) both pass locally.

## Overlap with GA4 enhanced measurement — read before approving

Verified in Tag Assistant on `/lp/workflow-orchestration` (2026-08-20, after this PR was opened): **GA4 already sends a native `form_start` hit on this page**, via enhanced measurement "Form interactions" (driven by the `gtm.formInteract` events). So for GA4 specifically, this PR is partially redundant. That was not known when it was written, and it is a fair reason to push back on it.

What the event still buys, and why I'd still merge it:

- It lands in **PostHog**, which is where the rest of the demo funnel actually lives (`$pageview` from `analytics.js`, `bookdemo_form` from this component). GA4 receives **neither** `bookdemo_form` nor `hubspot-meeting-booked` — the container fires only Google Ads and LinkedIn tags on those events — so the funnel cannot be assembled there today.
- GA4's native `form_start` is not scoped to the demo form: it fires for any form on the page, and does not distinguish the hero instance from the one further down.

If the team would rather measure form starts in GA4 and skip the extra code, closing this is a legitimate outcome — the funnel would then just be split across two tools.

## Surrounding context

The Google Ads side of this funnel was reworked and published on 2026-08-20: the single conversion action that fired on both `bookdemo_form` and `hubspot-meeting-booked` was split into `Demo Form Submitted` (Submit lead form, primary) and `Demo Booked` (Book appointment, secondary), so form submits and booked meetings are no longer indistinguishable. Verified end to end in preview. This PR fills the remaining gap on the other side of the funnel: there is currently nothing measured between landing on the page and completing a submit.

Also confirmed during that test, relevant to this file: the tracking block correctly sits **after** the awaited `submitHubspotForm`, so a rejected submission (`BLOCKED_EMAIL` on a free email domain) emits no event at all.


src/components/demo/Meeting.vue
```diff
@@ -41,6 +41,32 @@
...
+            // with `enabled: false`, so the dataLayer has to be written here.
+            window.dataLayer = window.dataLayer || []
+            window.dataLayer.push({
+                event: "bookdemo_form_start",
+                noninteraction: false,
+            })
+        } catch (analyticsError) {
+            console.error("Demo form analytics error", analyticsError)
+        }
+    }
```
iitzIrFan, line 68: All four points folded into one applicable suggestion: the field-target gate (#1), plain `let` (#2), the `pushDataLayer` helper (#3), and the comment trim. After applying, the submit handler at line 176 should be switched over to `pushDataLayer("bookdemo_form")` too — that's the diff in the review body, since it sits on unchanged lines.

Verified locally: `oxlint` clean, `astro check` 0 errors across 447 files.

```suggestion
    // Push directly to the dataLayer: the vue-gtm plugin is initialized with
    // `enabled: false` (GTM is loaded manually after cookie consent in
    // cookieconsent.ts), so gtm.trackEvent() is a no-op and never reaches it.
    const pushDataLayer = (event: string) => {
        window.dataLayer = window.dataLayer || []
        window.dataLayer.push({ event, noninteraction: false })
    }

    /* `focusin` rather than `input` on purpose: someone who tabs in and gives
     * up on "number of employees" has still started the form, and that is
     * exactly the drop-off worth seeing. Fires once per mounted form, and only
     * for real fields — the consent link and submit button also bubble here. */
    let started = false

    const onFormStart = (e: FocusEvent) => {
        if (started) return

        const field = e.target as HTMLElement | null
        if (!field?.matches?.("input, select, textarea")) return

        started = true

        try {
            posthog.capture("bookdemo_form_start")
            pushDataLayer("bookdemo_form_start")
        } catch (analyticsError) {
            console.error("Demo form analytics error", analyticsError)
        }
    }
```

iitzIrFan (commented):
Verified everything in the description against the tree — the reasoning holds up. Confirming the load-bearing claims so they're on the record:

- `enabled: false` on the vue-gtm plugin at `src/vue-setup.ts:20`, and GTM loaded manually post-consent at `src/scripts/cookieconsent.ts:41`. I went further and checked the library: `trackEvent` gates on `this.options.enabled ?? false` (`@gtm-support/core/lib/index.js`), so it really does push nothing. The direct-to-dataLayer approach is correct, not a shortcut.
- The component genuinely mounts twice on `/lp/workflow-orchestration` — `Hero.astro:37` (`client:idle`) and `UseCases.astro:58` (`client:visible`).
- No `autofocus` on any field, so the event can't fire without user action. Confirmed.
- `focusin` over `focus` is right, since `focus` doesn't bubble to the form.
- `npx oxlint src/components/demo/Meeting.vue` is clean and `npx astro check` passes (447 files, 0 errors) — CI should be green.

Four suggestions, one of which I think is a real bug. Inline suggestion below covers all of them for the added block.

---

### 1. Focusing the Privacy Policy link counts as a form start

`Meeting.vue:337` puts `<a target="_blank" href="/privacy-policy">` **inside** the `<form>`, and `focusin` bubbles from it — same for the submit button. So clicking the consent link without ever touching a field emits `bookdemo_form_start`, which inflates the very denominator this event exists to provide. Gating on the target tag fixes it and makes the code match the comment's "first interaction with any field".

### 2. `started` doesn't need to be a `ref`

It's never read in the template, so a plain `let started = false` skips the reactive wrapper.

### 3. The dataLayer explanation is duplicated

The comment at lines 58-59 restates the four-line comment 115 lines below at `Meeting.vue:176`. Two copies of one explanation in one file will drift. Extracting the push lets the explanation live once and shortens the existing submit handler — this part touches unchanged lines so it isn't in the applicable suggestion:

```diff
         try {
             posthog.capture("bookdemo_form")
             hsq.push(["trackCustomBehavioralEvent", { name: "bookdemo_form" }])
-            // Push directly to the dataLayer: the vue-gtm plugin is
-            // initialized with `enabled: false` (GTM is loaded manually
-            // after cookie consent in cookieconsent.ts), so
-            // gtm.trackEvent() is a no-op and never reaches the dataLayer.
-            window.dataLayer = window.dataLayer || []
-            window.dataLayer.push({
-                event: "bookdemo_form",
-                noninteraction: false,
-            })
+            pushDataLayer("bookdemo_form")
```

Net effect with the inline suggestion applied: **+32/-9** instead of +27/-0, and the submit handler gets shorter.

### 4. The guard is per-instance, not per-visitor

> without the guard one visitor moving between them would inflate the count

The guard is a per-instance ref, so a visitor who touches both `/lp/*` forms still emits two events — which test plan step 3 correctly expects. Both statements are fine on their own; the description just needs to say the guard prevents repeat events from the *same form instance*, and that the metric should be read as unique users rather than raw event count. Worth being explicit so nobody later reads the raw count as "visitors who started".

---

### On the comments

Measured across the repo: Vue components average **0.55%** comment lines; `Meeting.vue` is at **6.9%** even before this PR. Not asking for a rewrite of the file, but on the added block specifically:

- **Keep** the `focusin` vs `input` rationale. It's the one genuinely non-obvious decision here, and without it someone "fixes" it to `@input` next quarter.
- **Drop** the funnel-motivation sentences (lines 44-46). That's exactly what the commit message and this description are for, and it goes stale the moment the funnel changes.
- **Drop** the duplicated vue-gtm note (lines 58-59), per #3 — the one at line 176 is the canonical copy.

### Out of scope, but worth knowing

This file is the only place in the repo that pushes to the dataLayer directly. Every other form still goes through `gtm?.trackEvent(...)` — `newsletterSubmit.js:40`, `hubspot.ts:50`, `identify.js`, plus `cloud/Form.vue`, `contact/Header.vue`, `early-adopter-program/Form.vue`, `partners/BecomeAPartner.vue`, `preview-access/Features.vue`, `airflow-2-eol-whitepaper/DownloadForm.vue`, `resources/ResourceCtaPair.vue`. By the same `enabled: false` reasoning this PR relies on, all of those are no-ops, so `newsletter_form`, `cloud_form`, `contact_form_submission` and friends never reach GA4. One consequence lands inside this file: `identify(em)` in the submit handler calls `gtm?.trackEvent({event: "identify"})`, which does nothing — only `cookieconsent.ts:73` gets an `identify` event through.

The real fix is probably `loadScript: false` + `gtm.enable(true)` after consent in `cookieconsent.ts`, which would make all ~10 call sites work and let this file drop the local helper entirely. Separate PR — a file-local helper is the right call for two call sites.

Last thing: is there a GA4 event tag and trigger for `bookdemo_form_start` in the GTM container? The dataLayer push is inert until then, so the GA4 half of the funnel doesn't land with this PR alone. PostHog capture works immediately either way. Worth a line in the description about who's picking that up.

PR #5361, opened 2026-08-21
fix(search): drop search results with null titles breaking the Blogs facet

### LINKED ISSUE #5360: [Bash2.0] [Docs] Search "Blogs" facet shows "No results found" despite non-zero result count
### Describe the issue

Bug: In the global search modal, searching a term like assets shows a Blogs (30) facet, but clicking it renders "No results found for the current search". Other facets (Docs, Blueprints, Plugins) work fine.

---

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5360

---

ref: 

https://github.com/user-attachments/assets/02d33781-7d0b-4627-929e-d74469708ff3

---

### DIFF
--- a/src/components/layout/Search.vue
+++ b/src/components/layout/Search.vue
@@ -259,6 +259,7 @@
     import ContentCopy from "vue-material-design-icons/ContentCopy.vue"
     import posthog from "posthog-js"
     import { $fetchApi } from "~/utils/fetch"
+    import { prepareSearchResults } from "~/utils/searchResults"
 
     export default {
         props: {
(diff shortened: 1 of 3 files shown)


MilosPaunovic (approved):
The fix is correct and verified end-to-end against the live search `API`: the index still contains `blogs/CLAUDE` with a null `title`, and with this change the `Blogs` facet renders the remaining 29 results with highlighting intact instead of "No results found". Two non-blocking notes: the facet badge still shows the server-side count (30) while the list renders 29, which will resolve itself once the indexer skips frontmatter-less files like `blogs/CLAUDE.md`; and `npm run test:unit` is not run by any `CI` workflow, so the new suite (like the existing `redirects.test.ts`) only runs locally - adding a vitest step to the Lint & Astro Check workflow would be a good follow-up.

PR #5414, opened 2026-08-25
fix(ci): skip contributor onboarding for authors who already have write access

MilosPaunovic:
Related to https://github.com/kestra-io/docs/pull/5356.

## What

The organization membership check added in https://github.com/kestra-io/docs/pull/5356 never worked. The bot token is scoped to a single repository, so `GET /orgs/kestra-io/members/{username}` answers `404` for every author, and the step could not tell that apart from a genuine non-member:

```
GET /orgs/kestra-io/members/soyasis - 404
```

Every author was therefore classified as an outside contributor and got the onboarding comments. It showed up first in the `kestra` repository, where https://github.com/kestra-io/kestra/pull/18569 got the "First PR merged 🎉" congratulation even though its author is an organization member with private membership - this repository runs the identical workflow and has the identical bug.

This reads the author's repository permission instead of their organization membership:

- `GET /repos/{owner}/{repo}/collaborators/{username}/permission` only needs `metadata: read`, which the repository-scoped installation token already has - no new app permission to grant.
- It separates the two populations cleanly. Every `kestra-io` member resolves to `admin`, every outside contributor to `read`, verified against both current teammates and recent external contributors to this repository.
- The call is no longer wrapped in a `catch`, so a token that cannot answer fails the workflow loudly instead of silently commenting on a teammate's `PR`.

The companion change for the `kestra` repository is https://github.com/kestra-io/kestra/pull/18573.

### DIFF
--- a/.github/workflows/welcome.yml
+++ b/.github/workflows/welcome.yml
@@ -11,7 +11,7 @@ permissions: {}
 
 jobs:
   welcome:
-    # author_association misses organization members with private membership, so the membership lookup below is the authoritative check.
+    # author_association misses organization members with private membership, so the permission lookup below is the authoritative check.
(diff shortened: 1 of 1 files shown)


iitzIrFan (approved):
LGTM!

PR #5415, opened 2026-08-25
seo(plugins): shift how-to headings so plugin pages have a single h1

vfanucci:
## Problem

Bing Webmaster Tools flags 63 pages under **"There are multiple `<h1>` tags on the page"** (high severity). 47 of the 50 URLs in the export are plugin pages, and all 50 confirmed ≥2 `<h1>` when checked against production.

On plugin pages the cause is the how-to section. Its markdown is authored in the plugin repositories and starts at heading level 1 (`# How to use the Ansible plugin`), so it renders as a second `<h1>` next to the page title:

```
h1: Ansible CLI                          <- page title
h1: How to use the Ansible plugin        <- how-to markdown
h2: Common properties
h2: Tasks
```

## Why shift the whole tree instead of just the first heading

Demoting only the leading `#` to `##` would fix the count, but the how-to body already uses `h2` for its own subsections — they would become siblings of their parent rather than children, flattening the section. These pages are part of the plugin LLM-visibility work, where heading nesting is what scopes an extractable passage, so the hierarchy is the point.

Shifting every level down one (`#`→`##`, `##`→`###`, …) fixes the duplicate `h1` *and* keeps the structure:

```
h1: Ansible CLI
h2: How to use the Ansible plugin
h3: Common properties
h3: Tasks
h3: Output capture and custom callbacks
```

Fixing this at render time keeps it to one file, instead of a coordinated change across ~30 plugin repositories.

## Fenced code blocks

A naive line-based regex would corrupt the flow examples: the YAML and shell snippets contain comments (`# Set your bootstrap servers`) that are indistinguishable from headings once you only look at line starts. The helper tracks ``` / ~~~ fences and leaves their contents untouched. Levels cap at 5 so nothing becomes `#######`.

## Verification

Checked on a dev server against this branch:

- `/plugins/plugin-kafka` — `h1` count 2 → 1; how-to now `h2` with `h3` children (Authentication, Common properties, Tasks)
- `/plugins/plugin-ansible` — same, and its fenced block renders unchanged
- **Table of contents unaffected.** It is not derived from rendered heading levels: `buildPluginPageProps.ts` builds it by hand with a constant `TOC_DEPTH = 2`, labels it via `extractFirstHeading()` on the raw markdown, and anchors it to the wrapper `<div id="how-to-use-this-plugin">` rather than the heading. The "How to use the Ansible plugin" entry still renders and still links correctly.

## Intentional visual change

The how-to title now renders at `h2` size and its subsections at `h3`. The markdown renderer has no per-level heading styles, so headings follow the global type scale — the section reads slightly smaller but stays hierarchically distinct.

## Out of scope

The remaining ~16 pages in the Bing report have a different cause: `<h1>` used as a section style in composable marketing components. `/use-cases/ci-cd` and `/1-0` carry three each (hero + section + CTA), and 61 components in the repo contain an `<h1>`. That needs a hero-vs-section convention and is worth its own PR.


src/components/plugins/PluginIndex.vue
```diff
@@ -114,6 +114,27 @@
...
+    // two <h1>. Shift the whole heading tree down one level so the section keeps its
+    // hierarchy under the page title. Headings inside fenced code blocks are skipped:
+    // the flow examples use YAML and shell comments that look exactly like headings.
+    const shiftHeadings = (markdown: string) => {
+        let insideFence = false;
+
+        return markdown
+            .split("\n")
+            .map((line) => {
+                if (/^\s*(```|~~~)/.test(line)) {
```
iitzIrFan, line 128: **Minor (non-blocking):** the fence toggle doesn't distinguish fence type or length, so a ` ``` ` line inside a `~~~` fence (or a nested 4-backtick fence containing ` ``` `) mis-toggles `insideFence`, and heading-like lines inside that code block would get shifted. I checked all 202 current plugin `longDescription`s and none trigger this — but since these docs are authored across ~30 external plugin repos, a future doc could regress silently. Remembering the opening fence's character (and only closing on a matching one) would close the gap.

src/components/plugins/PluginIndex.vue
```diff
@@ -114,6 +114,27 @@
         return plugin.value?.longDescription ?? plugin.value?.description
     });
 
+    // The plugin how-to markdown is authored in the plugin repositories and starts at
+    // heading level 1, which collided with the page title and left every plugin page with
+    // two <h1>. Shift the whole heading tree down one level so the section keeps its
+    // hierarchy under the page title. Headings inside fenced code blocks are skipped:
+    // the flow examples use YAML and shell comments that look exactly like headings.
+    const shiftHeadings = (markdown: string) => {
```
iitzIrFan, line 122: **Optional:** this is a pure string helper, and the repo already unit-tests exactly this kind of function in `src/utils/plugins/` (`compareVersions.test.ts`, `plugin.test.ts`, ...). Extracting it there with a small vitest suite (fences, `~~~`, info strings, h5 cap, `#hashtag` non-heading) would lock in the edge-case behavior. Keeping it inline does match the `extractFirstHeading` precedent in `buildPluginPageProps.ts`, so this is a preference, not a violation.

iitzIrFan (commented):
Reviewed and verified the approach — the render-time shift is the right interception point, and I confirmed the TOC claims against `buildPluginPageProps.ts`. I also scanned all 202 plugin `longDescription` payloads from `api.kestra.io/v1/plugins`: no setext headings, no indented headings, no `~~~` or nested fences, and all fences balanced — so the helper is safe for the entire current corpus. Two minor, non-blocking notes inline.

Issue #5424, opened 2026-08-26
Improve the /resources page template: h2-only TOC, mid-page and end-of-page CTAs, TL;DR box


vfanucci:
## Context

The `/resources/*` template (`src/components/resources/ResourceArticle.astro`) now serves **266 pages** and is one of our main organic entry points. Reference page: https://kestra.io/resources/data/iceberg-vs-delta-lake

Two problems today:

1. **The TOC is unreadable on long pages.** It renders every heading, h2 *and* h3. 242 of the 266 resource pages contain h3s, and the worst offenders have 25–30 of them (`etl-pipeline-tools`, `data-lineage`, `batch-vs-streaming-processing`…). The sidebar becomes a wall of links instead of a map of the page.
2. **There is almost no conversion path.** A resource page has zero CTA between the hero and the "similar resources" grid. The only CTA that exists (`ResourceCtaPair`, the gated-asset + demo block) is opt-in via front-matter `cta:` and is currently set on **1 page out of 266** (`dagster-to-kestra-migration`).

So we get the traffic, and then nothing happens with it.

---

## 1. Limit the TOC to h2 (must have)

Only h2 headings should appear in the resources TOC. h3s stay in the page (they're useful for scanning and for AI extraction), they just leave the sidebar.

Implementation: in `ResourceArticle.astro`, the `links` array is built from all `headings`:

```js
const links = headings.map((h) => ({ text: h.text, id: h.slug, depth: h.depth }))
```

Filter on `h.depth === 2`. Please do it **in `ResourceArticle.astro`, not in `BlogToc.vue`** — `BlogToc` is shared with `/blogs`, and we don't want to change blog behaviour in the same PR.

**Acceptance:** on `iceberg-vs-delta-lake` and on `etl-pipeline-tools`, the sidebar TOC lists only the h2s, scroll-spy still highlights the right entry, and the mobile collapsed TOC still works.

---

## 2. Mid-page CTA (must have)

A simple button pair, in the same style as the homepage hero (`src/components/home/Hero.astro`) and `ClosingCTA.astro`:

- **Get Started** → `/get-started` (`btn btn-primary`)
- **Book a Demo** → `/demo` (`btn btn-secondary`)

Wording exactly "Get Started" / "Book a Demo" (not "Get Started with Open Source" — too long inside body copy).

**Placement:** roughly mid-article, inserted after the h2 at index `ceil(nbH2 / 2)`, so it always lands on a section boundary and never splits a paragraph or a table.

[...]

PR #5427, opened 2026-08-26
docs: filter table of contents to include only H2 headings

iitzIrFan:
Part - 1: https://github.com/kestra-io/docs/issues/5424

---


vfanucci (approved):
Checked the preview on desktop and mobile: the resources TOC now lists only the h2s, scroll-spy highlights the right entry, and the collapsed mobile TOC still works. Blog TOC is untouched as requested.

Approving part 1. Two things before this can go in: please take it out of draft and sync with main (the branch is 64 commits behind).

PR #5440, opened 2026-08-28
fix(assets): serve /docs/tutorial/logos/kestra-logo.png (~73k 404/week from Gmail)

vfanucci:
## What

Adds `public/docs/tutorial/logos/kestra-logo.png` (a copy of `public/logo.png`).

## Why

`https://kestra.io/docs/tutorial/logos/kestra-logo.png` has never existed on the site, and it is currently the single largest source of 404s on the zone — about **5x every other 404 combined**.

Cloudflare Log Explorer, 20–27 Aug 2026:

| path | requests | status |
|---|---|---|
| `/docs/tutorial/logos/kestra-logo.png` | 73,347 | 404 |
| `/cdn-cgi/image/f=webp,q=80/…` (same asset, zone-side transform) | 71,916 | 404 |

So ~73k logical requests/week — the second line is the image-transform subrequest, not a second URL (ordinary `/docs/*.png` traffic shows the same ~1:1 pairing).

**145,261 of those 145,263 requests come from one user agent: `Mozilla/5.0 … (via ggpht.com GoogleImageProxy)`** — Gmail's image proxy. Referer is empty on 100% of them.

That means:

- This is **not a broken link on the site**. Nothing in this repo references the path, and no request carries a referer. The URL lives in an **email template**.
- SEO impact is nil (1 GPTBot hit, 1 img2dataset hit over 8 days).
- The actual damage is that **the Kestra logo renders broken in every email opened in Gmail**, roughly 18,000 times a day.
- Onset: 0 in mid-July, ~55/day from 29 Jul, 667 on 4 Aug, ~9k/day from 12 Aug, flat since. A template shipped in the last days of July points at a URL that never existed.

## Why fix it here rather than only in the email template

The offending template still needs to be corrected (it should point at `https://kestra.io/logo.png`, the convention already used by `console/…/email/send_code.html`, `send_added_to_organization.html` and `kestra-ee/…/mail/template.html`). We have not identified it yet — it is not in any kestra-io repo, so it is in HubSpot or Lemlist.

But fixing the template only helps future sends. Serving the asset here also un-breaks the logo in the mail **already sitting in recipients' inboxes**, and stops the retry traffic: Google's proxy never caches a 404, so it keeps re-fetching.

`public/logo.png` was chosen over the 220 KB brand asset given the request volume: 889x203, 14.5 KB.

## Verification

On the dev server:

- `/docs/tutorial/logos/kestra-logo.png` -> `200 image/png`
- neighbouring non-existent file -> `404` (control)
- the `/docs/tutorial` page route still returns `200`

## Out of scope

Same family, much smaller volume, not addressed here: `/docs/tutorial/logos/logo-light-version.png` (~6/day, referred — it is what `ui/src/components/content/CardLogos.vue` loads), `/docs/ui/kestra-logo-big.png` (~9/week), `/docs/icons/ks-logo.png` (~6/week).


iitzIrFan (approved):
Verified locally on the branch:

- `/docs/tutorial/logos/kestra-logo.png` → **200, `image/png`, 14,475 bytes**, and the served bytes are SHA-identical to `public/logo.png` (`ab6c56af`)
- neighbouring non-existent file in the same directory → 404 (control)
- `/docs/tutorial` page route → still 200

Also traced the production path: with `run_worker_first: true`, the Worker's extension check (`src/middlewares/worker.ts:198`) sends any extension-suffixed path straight to the assets handler, so this will serve correctly through Cloudflare, and the `/cdn-cgi/image/…` transform subrequest will start succeeding once the origin returns 200.

Follow-ups (non-blocking):
1. **Root cause**: the offending email template in HubSpot/Lemlist still needs to be found and pointed at `https://kestra.io/logo.png` — this PR un-breaks existing inboxes but new sends will keep using the wrong URL until then.
2. **Post-deploy check**: the ~73k/week 404s for this path should drop to near zero within a day or two; worth a quick Log Explorer look next week to confirm.
3. This file is an untracked copy of `public/logo.png` and will go stale if the logo is redesigned — fine for an email asset, just noting it.

PR #5442, opened 2026-08-28
fix: prevent overflow in kestra vs and Orchestration pages on mobile screens

### LINKED ISSUE #5382: [Bash2.0] Filter tabs overflow the page on mobile
## Describe the issue

On phones, the tool filter tab bar on the Orchestration and Kestra vs. pages is wider than the screen. Instead of fitting or scrolling on its own, it stretches the whole page wider than the phone, so the page scrolls sideways and the last tabs are cut off.

## Steps to reproduce

**1.** On a phone (or a browser window about 390px wide), open `kestra.io/orchestration`.

**2.** Scroll down to the row of tabs: "All tools", "Infrastructure", "Cloud", "Data"...

**3.** Try to swipe left/right on the page.

## Actual behaviour

The tab row is far wider than the screen, so the whole page scrolls sideways. On Orchestration the page runs about 230px past the screen edge, on `kestra.io/vs` about 275px. Only the first two or three tabs are visible, the rest sit off-screen, and there's empty space to the right of every section because the page got stretched.

Image

## Expected behaviour

- Keep the tab bar inside the screen: let it scroll on its own, or wrap it, so the page itself doesn't scroll sideways and all tabs stay reachable.

## Environment

- Site: kestra.io (Orchestration and Kestra vs. pages, mobile widths)

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5382

---

Ref. video: 

https://github.com/user-attachments/assets/270f4bdb-d784-445d-bbb1-d1508db3baee

---

### DIFF
--- a/src/pages/orchestration/index.astro
+++ b/src/pages/orchestration/index.astro
@@ -596,6 +596,14 @@ const faqItems = [
         margin-bottom: 1.5rem;
         border-bottom: 1px solid var(--ks-border-secondary);
         width: fit-content;
+        max-width: 100%;
+        overflow-x: auto;
+        -ms-overflow-style: none;
+        scrollbar-width: none;
+
+        &::-webkit-scrollbar {
+            display: none;
+        }
     }
 
     .toggle-btn {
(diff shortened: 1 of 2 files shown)


RemiBarthe (approved):

iitzIrFan:
thanks @RemiBarthe 🙌

PR #5060, opened 2026-07-02
fix(blueprints): make blueprint count dynamic and consistent site-wide

fdelbrayelle:
## Summary
- `/get-started` showed "250+ Blueprints" (stale, missing space in one spot) while `/blueprints` shows the real live count (399). Fixed 4 hardcoded copy spots to pull the actual total from the API instead.
- Added `fetchTotalBlueprintsCount()` (`src/utils/blueprints/blueprintCount.ts`), mirroring the existing plugin-count pattern (`fetchTotalPluginsCount`): floors the real count to the nearest 10 and callers append `+`, so e.g. 399 → "390+" (same "at least X" semantics used for plugin counts, just nearest-10 instead of nearest-100 given the smaller magnitude).
- Updated: `src/pages/get-started.astro`, `src/components/home/Blueprints.astro`, `src/components/get-started/SetupSteps.astro`, `src/components/common/BlueprintsRender.astro`.

## Test plan
- [x] `oxlint` clean on all touched files
- [x] Confirmed no remaining hardcoded blueprint-count strings (`grep -ri "250+" src | grep -i blueprint` → empty)
- [ ] Visually spot-check `/get-started` and homepage blueprint sections after deploy preview

PR #5446, opened 2026-08-30
fix(website): single source of truth for plugin and blueprint counts

### LINKED ISSUE #5387: [Bash2.0][Website] Plugin count is inconsistent across the site, six different values including a broken "0+ Plugins"
### Describe the issue

The plugin count is stated with six different values across kestra.io. The canonical https://kestra.io/plugins page says 1900+, which makes the others wrong:

- 1900+ on home, /plugins, /overview, /enterprise hero, /features, /data, /vs and the /pricing comparison table.
- 1,200+ in the /pricing Open Source section.
- 1,700+ on /cloud, "language-agnostic tasks with 1,700+ plugins".
- 1400+ on /vs/google-workflows, and in the SEO meta description on both home and /pricing.
- 1200+ on /vs/google-workflows, "1200+ other tools".
- 0+ on the /enterprise Open Source edition card, which reads "0+ Plugins" and looks like a broken placeholder value.

Some pages contradict themselves. /vs/google-workflows shows 1400+, 1200+ and 1900+ on one page. /enterprise shows 1900+ in the hero and 0+ on the Open Source card. Home and /pricing show 1900+ and 1,200+ in the body while their SEO meta says 1400+.

### Visuals and Reproducibility

Image

1. Open https://kestra.io/plugins and note the count, 1900+.
2. Open https://kestra.io/pricing. The Open Source section says 1,200+.
3. Open https://kestra.io/cloud. It says 1,700+.
4. Open https://kestra.io/vs/google-workflows. It shows 1400+, 1200+ and 1900+ on the same page.
5. Open https://kestra.io/enterprise. The hero says 1900+ and the Open Source card says "0+ Plugins".

### Environment

Site: https://kestra.io

### LINKED ISSUE #5392: [Bash2.0][Website] Blueprint count is inconsistent (680+ vs 480+ on the same /blueprints page, plus 200+ and 576 elsewhere)
### Describe the issue

The blueprint count contradicts itself on the same page. https://kestra.io/blueprints says "680+ Blueprints, ready to build." in the hero, while the "Start From ... Blueprints" section further down the same page says "480+ Blueprints".

The count also disagrees across pages. The homepage and /get-started say 480+, /data says "200+ production-ready blueprints", and /plugins lists 576 blueprints for the Core plugin alone.

### Visuals and Reproducibility

Image

Image

Image

1. Open https://kestra.io/blueprints. The hero says 680+ and the "Start From" section lower down says 480+.
2. Compare with https://kestra.io (480+), https://kestra.io/get-started (480+), https://kestra.io/data (200+) and https://kestra.io/plugins (Core alone lists 576).

### Environment

Site: https://kestra.io

iitzIrFan:
Closes #5387
Closes #5392

---

## What was wrong

The plugin count appeared with six different values across the site (1900+,
1,700+, 1,400+, 1,200+, 1300+, and a broken **"0+ Plugins"** on /enterprise),
and /data claimed "200+ production-ready blueprints" while the catalog holds
690. The canonical values were already computed dynamically — the stale
numbers were pages that never adopted the existing utilities.

## What this changes

**Plugins** — everything now reads `fetchTotalPluginsCount()` (distinct task
classes from `/plugins/subgroups`, floored to hundreds and formatted with a
thousands separator → currently **2,000+**):

- `/pricing` Open Source card and comparison table (was 1,200+ and a second
  raw fetch in `PlanTable.astro`), `/cloud` (was 1,700+), and the
  SoftwareApplication JSON-LD used on home/pricing/features/enterprise
  (was 1400+).
- Two leftover hardcodes on the software-providers and public-services
  use-case pages (which already fetched the count for other sections).
- **`/enterprise` "0+ Plugins" bug**: `Dashboard.vue` used the client-side
  `usePluginsCount()` composable, whose `"0+"` initial value is what SSR
  renders. The page now fetches the count at build time and passes it down
  as a prop.
- **`/1-0`** had the same SSR `0+` pattern in `Header.vue` and
  `AiCopilot.vue` — both now receive the build-time count as a prop.
- **`/features`** (`EveryPlugin.astro`) had the last remaining `0+`
  fallback; it now uses `fetchTotalPluginsCount()` (the carousel keeps its
  own metadata fetch and graceful degradation).
- **`/lp/workflow-orchestration`** hardcoded 1,800+ in `Platform.astro` and
  `Stack.astro`.
- **`/two-zero`** hardcoded `2,000+` in the hero key numbers
  (`twozero/Hero.astro`) and `Browse 2,000+ plugins` in the Day-Two
  Operations CTA (`pages/two-zero.astro`).
- **vs pages**: the `/vs/[slug]` template already substituted a
  `{totalPlugins}` placeholder, but only in one field. That substitution now
  covers the whole YAML entry (headings, feature lists, FAQ answers, SEO
  meta) and the `/vs` index cards. All 105 hardcoded Kestra plugin totals
  across 34 vs YAMLs were converted to the placeholder — each one audited to
  confirm it means Kestra's total. Competitor figures ("1,000+ community
  nodes", "700+ pre-built connectors", "400+ built-in") are untouched.
- **orchestration pages**: the same placeholder + whole-entry substitution
  was applied to `/orchestration/[slug]` and the index, converting the
  hardcoded totals in 14 orchestration YAMLs.

**One counting and rounding rule** — `calculateTotalPlugins()` and a new
`formatPluginCount()` live in `utils/plugins/pluginCount` next to the fetch,
and every consumer uses them: the `usePluginsCount()` composable, the five
prerendered pages that previously fetched `/plugins/subgroups` and floored
locally (`overview` and the four code-in-any-language pages), and
`/plugins`. `FeaturePlugins.astro` had the same local floor but is not
imported anywhere in `src`, so it is deleted rather than migrated. The
duplicate
`calculateTotalPluginCount()` in `pruneForClient.ts` is deleted. The
floor-to-hundred expression now exists in exactly one place. Contract:
`fetchTotalPluginsCount()` and `formatPluginCount()` return the bare
formatted number and callers append `+`; `usePluginsCount()` returns it
with the `+` (documented on both).

`/plugins` is `prerender = false`, so it derives the count per request from
the groups it already fetches rather than the build-time memo — otherwise the
memo would freeze the value for the lifetime of a Worker isolate.

**Blueprints** — `/data` now uses the existing `fetchTotalBlueprintsCount()`
(currently **690+**). That was the last hardcoded blueprint count; everything
else was already dynamic since #5060.

`/blueprints` is `prerender = false`, so like `/plugins` it calls the
un-memoized `loadTotalBlueprintsCount()` instead of the build-time memo,
which would otherwise freeze its headline for the lifetime of a Worker
isolate. It falls back to `BLUEPRINT_COUNT_FLOOR` there, the way the rest of that page
already degrades when the API fails, rather than taking the page down with it.

`BlueprintsRender.astro` was a `fetchTotalBlueprintsCount()` consumer that
nothing imports — dead on `main` before this branch — so it is deleted as
well. Its one import, `BlueprintsCarousel.astro`, is still used by
`home/Blueprints.astro`. That leaves four consumers of the memoized count,
all prerendered.

**Reliability** — both count utilities share the same pattern, since their
values are rendered in page copy and SEO markup:

- module-level request memoization (one request per build, shared by all
  callers);
- transient failures are retried by `internalFetch` (#5637), which covers
  every call site rather than just these two;
- after the final attempt they fall back to `PLUGIN_COUNT_FLOOR` (2000) and
  `BLUEPRINT_COUNT_FLOOR` (690) — a plausible round number rather than either
  "0+ plugins" or a discarded build. Both floors sit at or below the live
  total on purpose: a floor must never read as a precise-but-stale figure;
- a `200` carrying an empty or nonsensical payload takes the same fallback,
  but logs separately, since that points at an API change rather than a blip;
- failed requests are not cached, so the next caller retries and a brief
  outage cannot freeze the floor across a whole build;
- `replaceTotalPluginsPlaceholder()` only walks plain objects, so a `Date`
  or `image()` field added to a schema later passes through instead of
  silently collapsing to `{}`.

**On failing loudly**: an earlier revision of this branch let both utilities
propagate, on the reasoning that a wrong number baked into copy and JSON-LD is
worse than a failed build. Review pointed out that this reintroduced exactly
what #5631 is about — with the API down, `/` and `/get-started` returned 500,
so one blip on a decorative marketing number could discard a 2,500-route
build. The floor resolves both concerns: retry handles the transient case,
the floor handles the residual, and neither "0+" nor a red build stays
possible. The two per-request pages (`/plugins`, `/blueprints`) degrade the
same way.

**OSS vs EE wording** — the total from `/plugins/subgroups` includes
Enterprise-only plugin classes, so copy that scoped the number to the
open-source edition was reworded during review: the netbox FAQ no longer
attaches a count to its OSS sentence, "N+ open-source plugins" phrasings
dropped the qualifier, and edition-scoped copy says "most of Kestra's
{totalPlugins}+ plugins". That hedge now also applies to the `/pricing`
Open Source card bullet and the comparison-row tooltip. Project-level claims
("Open source with 26k+ GitHub stars and {totalPlugins}+ plugins") keep the
bare total.

## Reconciled with #5637

#5637 landed the bounded retry inside `internalFetch` and converted the
decorative build-time fetches, and it overlapped this branch in six files.
Merged in `4dcb4d1`:

- Five content conflicts (`EveryPlugin`, `PlanTable`, `overview`,
  `vs/[slug]`, `pluginCount`) resolve to this branch: #5637 did a smaller
  version of the same consolidation, and the formatted-string contract here
  supersedes its `Promise<number>`.
- `FeaturePlugins.astro` was a modify/delete. The deletion stands — the file
  still has no importers.
- `src/utils/fetch.ts` auto-merged **without a conflict marker** and left
  `RETRY_BASE_DELAY_MS` declared twice, which does not compile.
  `$fetchApiCachedWithRetry` and its two constants are dropped: retry now
  lives in `internalFetch` for every call site, and keeping both would nest
  retries 3x3 = 9 attempts per dead URL. `errorStatus` stays —
  `$fetchApiCachedOptional` uses it. The two count utilities go back to
  `$fetchApiCached`.

The 2.0 homepage revamp (#5493) also landed after this branch opened and
hardcoded `"2000+ Plugins"` in the hero's key-facts list, so the homepage was
rendering `2000+ Plugins` above `2,000+ plugins`. Converted in `7930f28`.

## Supersedes #5686

The 2.0 page (#5553) landed after this branch opened, carrying two more
hardcodes. #5686 proposed converting them — along with eight files this
branch already covers identically — against `main`'s util, where
`fetchTotalPluginsCount()` still returns a bare number and falls back to `0`.
That would have rendered `2000+` without the thousands separator and
reintroduced `0+ Plugins` (#5387) on /pricing, /, /cloud, /two-zero and in the
shared `SoftwareApplication` JSON-LD on an API blip. The two unique
conversions are picked up here in `d4f7f11`, against this branch's
formatted-string contract, so both spots keep their separator and degrade to
`PLUGIN_COUNT_FLOOR`, which leaves nothing in #5686 that this branch does not
already carry.

## Intentionally not changed

- Per-plugin blueprint counts on /plugins (e.g. Core's 576) and the
  category/pattern tile counts on /blueprints — legitimate subsets.
- Blog posts and resources articles (editorial content; blogs already use
  the `:PluginCount` component where a live count is wanted).
- The `/pricing` Open Source card bullet still reads "Most of Kestra's
  {totalPlugins}+ Plugins". Raised in review as slightly awkward in a
  checklist; left as a product wording call.

## Verification

- `oxlint` clean, `vitest` 379/379, `astro check` 0 errors.
- The count tests mock the virtual `astro:env/client` module and stub global
  `fetch` (rather than mocking `~/utils/fetch` wholesale), so the real
  retry/backoff and HTTP-error paths are what run under test. Tests cover
  rounding, the thousands separator, memoization, placeholder replacement
  including the `Date` guard, retry on a rejected fetch and on a 502,
  backoff timing, the floor fallback on exhausted retries and on an empty
  `200`, and that a floored failure is not memoized.
- With `API_URL` pointed at a dead port, `/`, `/get-started`, `/data`,
  `/use-cases`, `/pricing` and `/overview` all render 200 showing `2,000+`
  and `690+`. On the pre-floor revision `/` and `/get-started` returned 500.
- All modified YAMLs re-validated with js-yaml; a mechanical diff audit
  confirms every converted line changed only number → placeholder (plus
  required quoting), with competitor figures untouched.
- Plugin/blueprint counts independently recomputed from the API (2017
  distinct classes → 2,000+; blueprints total 695 → 690+) and matched
  against the rendered pages.
- Rendered checks on home, /pricing, /cloud, /enterprise, /features, /1-0,
  /data, /overview, /plugins, /two-zero, the code-in-any-language pages,
  /lp/workflow-orchestration, /vs and /orchestration pages: one consistent
  `2,000+` everywhere (no bare `2000+`), no raw `{totalPlugins}` in any
  HTML, no "0+ Plugins", /data shows 690+, no console/hydration errors.


iitzIrFan (commented):
A few non-blocking follow-ups I noticed:

* It may be worth memoizing `fetchTotalPluginsCount()` at module level. `$fetchApiCached` doesn't deduplicate the build-time Node requests, so the current implementation can make many independent API calls during a build. More importantly, a transient failure could theoretically result in one page receiving `0` while others receive the actual count. This is partly pre-existing, since the vs pages already called the utility independently, but this PR increases its usage.
* We may also want to revisit the `"0"` fallback in `fetchTotalPluginsCount()`. Now that the value is used in SEO/meta and JSON-LD content as well, silently generating `0+ Plugins` on an API failure could recreate the inconsistency we're trying to eliminate.
* A small unit test for `replaceTotalPluginsPlaceholder()` would be useful, covering nested objects/arrays and multiple occurrences.
* I also noticed the same stale plugin-count pattern remains in ~15 `src/contents/orchestration/` YAMLs, and `/1-0` appears to have the same client-side `usePluginsCount()` SSR `0+` pattern that was fixed here. I'd handle both separately rather than expanding this PR.

iitzIrFan (commented):
Remaining work:

* `/1-0` still has the SSR `0+ Plugins` pattern in `Header.vue` and `AiCopilot.vue`.
* 13 `src/contents/orchestration/` YAMLs still contain hardcoded `1400+` plugin counts.
* `/lp/workflow-orchestration` still has hardcoded `1800+` plugin counts in `Platform.astro` and `Stack.astro`.

flcarre (commented):
Thanks for taking this on — and for resisting the obvious shortcut. The central question with a PR like this is whether the "single source of truth" is a derived value or just one hardcoded number instead of sixty, and I checked: it's genuinely derived. `loadTotalPluginsCount()` fetches `/plugins/subgroups` and counts distinct `cls` values, `loadTotalBlueprintsCount()` reads `total` from the blueprints endpoint, both are memoized per build so ~28 call sites share one request, and failures are deliberately not cached. There's no static constant anywhere in the new path, and removing the `"0+"` degradation that caused the `/enterprise` bug is the right call.

I also verified the bulk YAML edit is genuinely mechanical: every file has an exactly 1:1 removed/added line count, all 121 added YAML lines contain `{totalPlugins}`, there's no whitespace churn, and all 13 values that gained a leading `{` are correctly quoted. The JSON-LD is right too — I checked all four callers of `SchemaSoftwareApplication.astro`; three pass no `description` so the dynamic default ships, and `enterprise/index.astro` passes its own, which contains no count. And the `+` handling is correct at every migrated call site: `usePluginsCount()` returns `"1900+"` with the plus while `fetchTotalPluginsCount()` returns it bare, and all three Vue components re-add it. No double `+`, none lost.

Three things before merge.

## 1. `netbox.yaml` — the substitution changes what the sentence means

```
- The open-source edition includes nearly every other Kestra plugin (1300+)
+ The open-source edition includes nearly every other Kestra plugin ({totalPlugins}+)
```

That `1300+` is specifically the **OSS-only** count, in a FAQ answer whose entire point is that `io.kestra.plugin.ee.netbox` is Enterprise-only. `fetchTotalPluginsCount()` counts every class returned by `/plugins/subgroups`, EE included — so the page now claims the open-source edition ships the full total, contradicting itself one clause earlier.

This one should stay a literal, or be reworded. It's the single semantic miss in an otherwise clean bulk edit.

## 2. The `0+` bug is still live on `/features`

`src/components/features/core/EveryPlugin.astro` still opens with `let totalPlugins = "0+"`, does its own `Math.floor(count / 100) * 100`, and renders `Choose from {totalPlugins} plugins` — the exact failure this PR fixes on `/enterprise`, untouched.

More broadly, the rounding rule is re-implemented in four `.astro` files (`EveryPlugin.astro`, `PlanTable.astro`, `FeaturePlugins.astro`, `overview.astro`) plus `usePluginsCount.ts`, which returns the count *with* a trailing `+` where `fetchTotalPluginsCount()` returns it without. That's six implementations of one rule, so "single source of truth" isn't quite true yet. Exporting one `formatPluginCount(count)` and pointing them all at it would finish the job — at minimum, please kill the `"0+"` initialiser.

## 3. One transient API blip now fails the whole production build

`$fetchApiCached` has no retry, and the `cf: {cacheTtl}` hint applies to the Cloudflare Workers runtime rather than the Node/Astro static build. Combined with the new `throw`, a single failed `GET /plugins/subgroups` fails the build across all 23 plugin-count call sites and 5 blueprint ones. Previously it shipped a wrong number.

Failing loudly is the right trade — but not on one attempt. Two or three tries with a short backoff inside `loadTotalPluginsCount()` / `loadTotalBlueprintsCount()` before rethrowing would keep the guarantee without making the build fragile.

## Minor: the PR description is stale

The "Intentionally not changed" section lists three things as follow-ups that the diff actually changes — the workflow-orchestration landing page's `1,800+` (`Platform.astro`, `Stack.astro`), the `/1-0` client-side count (converted to build-time props in `1-0.astro`, `onezero/Header.vue`, `onezero/AiCopilot.vue`), and the orchestration YAMLs' `1400+` (14 files). The code is fine; it's the description that's out of date, and reviewers use that section to decide what *not* to check.

One thing I couldn't verify: the rendered-output checks in the PR body. The Cloudflare preview sits behind Cloudflare Access and returns a sign-in page.

iitzIrFan (commented):
Hey made the requested follow-ups: fixed the NetBox OSS wording, moved `/features` to the `canonical plugin count` and `added retry/backoff` for the count fetches. Also updated the PR description :)
Could you take another quick look when you get a chance? 
Thanks!

elevatebart (commented):
Reviewed on `3dd85b6`. Verified locally: `astro check` 0 errors, `vitest` 91/91, `oxlint` clean, all 34 vs + 14 orchestration YAMLs parse, 121 placeholders with none left as an unquoted scalar, and every remaining number in those YAMLs is a competitor figure. The mechanism is right: build-time fetch with prop-drilling replaces the SSR-`0+` composable, and whole-entry substitution replaces the one-field `.replace`.

Six notes, none blocking. The first two are small enough to fold in here.

- Thousands separator is lost (`src/utils/plugins/pluginCount.ts:32`). The util returns `"1900"`, so copy that read `1,700+` (cloud.astro), `1,200+` (price/Header.astro, use-cases) and `1,800+` (lp/*) now renders `1900+`, against the separator style used elsewhere. Format once in the util: `return new Intl.NumberFormat("en-US").format(rounded)`. YAML placeholders then pick it up too, and blueprints are unaffected either way.

- Two `+` conventions coexist. `usePluginsCount` (`src/composables/usePluginsCount.ts:32`) bakes the plus into the value; `fetchTotalPluginsCount` returns the bare number and 40 call sites append it. That divergence is what produced the `0+ Plugins` this PR fixes, so it is worth either returning the display string from the util or documenting the contract on both. Related: `/pricing` still renders the count from two paths, the new util in `price/Header.astro:55` and a raw fetch plus local floor in `price/PlanTable.astro:8-9`. Converting the second is two lines on a page already touched here, and it is what keeps the single-source claim partial.

- `fetchWithRetry` is duplicated verbatim (`pluginCount.ts:11`, `blueprintCount.ts:9`), same body and same constants. It belongs next to `$fetchApiCached` in `~/utils/fetch`. A PR about single sources of truth shipping a copy-pasted helper will get quoted back at you.

- `replaceTotalPluginsPlaceholder` structurally clones `entry.data` (`pluginCount.ts:64`). Safe today, since neither `src/schemas/vs.ts` nor `orchestration.ts` has `z.date()` or `image()`. Add either and `Object.entries` returns `[]`, so a `Date` silently becomes `{}` with no type error. Cheap guard: `Object.getPrototypeOf(value) === Object.prototype` in that branch.

- Loud failure is the right call for prerendered pages, worth stating as a trade-off: `features/core/EveryPlugin.astro` used to degrade gracefully on that endpoint and now takes the page down with it. If any of these pages ever gets `prerender = false`, the module-level memo also freezes the count for the Worker isolate's lifetime.

- Four comments in the two utils run 3 to 5 lines (`pluginCount.ts:8-10`, `36-40`, `blueprintCount.ts:6-8`, `35-38`) against the 2-line convention. Content is good, just long.

Not a defect, your call: the OSS pricing card and comparison-table row (`price/Header.astro:55`, `pricingData.ts:57`) now assert the full `/plugins/subgroups` total, which includes `io.kestra.plugin.ee.*` classes, in an edition-scoped context. The vs YAMLs got the "most of Kestra's" hedge and the pricing cards deliberately did not. Consistent with the reasoning in the description, but it is a claim about the free edition, so worth a conscious yes.

---

iitzIrFan:
@elevatebart thanks for the thorough pass. Follow-ups are in 12f4b7f08 (on top of the main merge), and the PR description is updated to match.

- **Thousands separator**: `fetchTotalPluginsCount()` now formats with `Intl.NumberFormat("en-US")`, so every call site and all 121 YAML placeholders render `1,900+`.
- **`+` convention / PlanTable**: `PlanTable.astro` uses the shared util; `getSections` takes a string. The contract is documented on both `fetchTotalPluginsCount()` (bare) and `usePluginsCount()` (with `+`).
- **`fetchWithRetry` dedupe**: now a single `$fetchApiCachedWithRetry()` next to `$fetchApiCached` in `~/utils/fetch`.
- **Prototype guard**: `replaceTotalPluginsPlaceholder()` only walks plain objects; a `Date` passes through intact (tested).
- **Comments** trimmed to two lines.
- **OSS pricing card**: went with the hedge rather than a bare yes. The Open Source card bullet reads "Most of Kestra's 1,900+ Plugins" and the comparison-row tooltip says most are available in both editions.
- **Trade-off** for `EveryPlugin.astro` failing loudly is now stated in the description.

I also pulled the deferred consolidation in, since the separator change would otherwise have left `/overview`, `/plugins`, and the code-in-any-language pages rendering a bare `1900+` next to `1,900+`. `calculateTotalPlugins()` and a new `formatPluginCount()` live in `utils/plugins/pluginCount`; the composable, the six prerendered pages that fetched and floored locally, and `/plugins` all use them, and `calculateTotalPluginCount()` in `pruneForClient.ts` is gone. The floor-to-hundred expression now exists in exactly one place. `/plugins` is `prerender = false`, so per your isolate note it derives the count per request from the groups it already fetches rather than the memo.

One test change worth knowing about: moving the retry into `~/utils/fetch` meant the old whole-module mock would have hidden it, so both count test files now mock the virtual `astro:env/client` module and stub global `fetch`. The real retry/backoff and HTTP-error paths run under test, with new cases for the separator, a 502 retry, the `Date` guard, and the two pure helpers.

`vitest` 297/297, `oxlint` clean, `astro check` 0 errors. Rendered checks show one consistent `1,900+` on every page listed in the description, no bare `1900+`, no `0+ Plugins`, no raw placeholder.

elevatebart (approved):

MartinRst (approved):

elevatebart (approved):
Approving. One follow-up worth doing before or right after merge, plus a few optional notes. Read the full utility diff, every page/component consumer, the tests, and a spot-check of five YAMLs (windmill, camunda, google-workflows, airflow, netbox).

### Should fix

`/blueprints` regresses the same way you fixed `/plugins`. `src/pages/blueprints/index.astro` is `prerender = false` and calls `fetchTotalBlueprintsCount()` per request (the `headlineCount` in the `Promise.all`, rendered in the H1). The new module-level memo freezes that value for the lifetime of the Worker isolate, where it previously refetched under the 1 h Cloudflare cache. Second effect: `fetchFullCatalog()` degrades gracefully on API failure, so the page used to survive an outage with "0+"; now the count throws after retries and 500s the whole page.

Fix mirrors what `/plugins` does: export the un-memoized `loadTotalBlueprintsCount()` and call that from the SSR page, or read `first.total` from the catalog response the page already fetches.

### Optional

- "Fail loudly" has a hole. A 200 with an unexpected body still ships "0+": `{ total = 0 }` in `blueprintCount.ts`, and an empty array through `calculateTotalPlugins`. The "treats a missing total as zero" test locks that in. Throwing on 0 / missing total would close it.
- `src/components/layout/FeaturePlugins.astro` has no importers anywhere in `src`. Deleting beats migrating it.
- `$fetchApiCachedWithRetry` retries 4xx as well as network/5xx, so a 404 costs 1.5 s before failing. Gating on `status >= 500 || !status` would keep retries for transient errors only.
- Wording: the `/pricing` Open Source card bullet "Most of Kestra's 1,900+ Plugins" reads a bit awkwardly in a checklist. Product call, not a code issue.

### Looks right

Memo resets on failure so concurrent callers share the rejection and the next call retries. `getStaticPaths` scope includes the count, so the incremental page cache invalidates when it changes. Placeholder walk only recurses plain objects. YAML conversions kept competitor figures and quoted values that start with `{`. Tests run the real retry path under fake timers rather than mocking the fetch layer.

---

iitzIrFan (commented):
other PR comment!

iitzIrFan:
Thanks — all four addressed in `8e344b75d`.

**`/blueprints` regression.** You were right about both effects. `loadTotalBlueprintsCount()` is now exported and the SSR page calls that instead of the memo, so the headline no longer freezes for the isolate's lifetime. It falls back to `0` on failure the way `fetchFullCatalog()` already does, so an outage degrades the page rather than 500ing it. `fetchTotalBlueprintsCount()` is unchanged for the six prerendered callers.

I went with the exported-loader option rather than reading `first.total` from the catalog: `fetchFullCatalog()` caps at 5 pages of 200 and discards `total`, so deriving the headline from `catalog.length` would under-report once the catalog passes 1000.

**The "fail loudly" hole.** Closed on both sides. Each utility now throws when the value it is about to return would be `0` — that covers the missing `total` and the empty array, and also a total small enough to round to zero, which would ship `0+` just the same. The "treats a missing total as zero" test that locked in the old behaviour is now a throw assertion, plus a rounds-to-zero case on each. Both assert a single fetch, since the request itself succeeded and shouldn't be retried.

**`FeaturePlugins.astro`.** Deleted. It was unimported on `main` too, so this branch was migrating a corpse. Nothing else referenced its icons — they resolve dynamically through `TaskIcon.astro`.

**Retry gating.** Now breaks unless `!status || status >= 500`, so a 404 fails on the first attempt. While doing it I noticed the status extraction already existed verbatim in `$fetchApiCachedOptional` — given your note about copy-pasted helpers, it is now a shared `errorStatus()` used by both. Covered by a 4xx-vs-5xx test in `fetch.test.ts`.

**`/pricing` wording** — left as-is, agreed it's a product call rather than a code issue.

One thing worth a conscious yes: the utilities now throw to avoid shipping `0+`, while `/blueprints` explicitly catches and shows it. That's deliberate — on a prerendered page a wrong number is baked in until the next deploy, but on the live page `0+ Blueprints` over an already-empty grid beats a 500. Happy to flip it if you'd rather the page fail.

`astro check` 0 errors, `vitest` 355/355, `oxlint` clean. PR description updated.

iitzIrFan:
One addition after the comment above, pushed as `054283ac5`.

`BlueprintsRender.astro` turned out to be in exactly the same state as `FeaturePlugins.astro` — a `fetchTotalBlueprintsCount()` consumer that nothing in the repo imports, and already dead on `main` before this branch — so it is deleted too. `BlueprintsCarousel.astro`, its only import, is still used by `home/Blueprints.astro`, so nothing else is orphaned by it.

That leaves four consumers of the memoized count, all prerendered (`home/Blueprints.astro`, `data/Actual.astro`, `get-started/SetupSteps.astro`, `get-started.astro`), with `/blueprints` the only per-request caller and the sole user of the un-memoized loader.

`astro check` 0 errors, `vitest` 355/355, `oxlint` clean. Description updated to match.

iitzIrFan:
Holding this one until #5637 merges, and then reworking part of it. Context, since the interaction isn't obvious from either diff.

#5637 (closes #5631) puts bounded retry inside `internalFetch` and converts the decorative build-time fetches to `$fetchApiCachedOptional`, under the rule "decorative sections degrade, route generation does not."

This PR takes the opposite position for the two count utilities, on purpose — `fetchTotalPluginsCount` and `fetchTotalBlueprintsCount` log and rethrow so a build fails loudly rather than shipping `0+`. I still think never shipping `0+` is right; fixing that exact bug on `/enterprise` is part of why this PR exists. But as written it reintroduces the failure mode #5631 is about.

Pointing `API_URL` at a dead port and running the dev server:

| page | main | this branch |
|------|------|-------------|
| `/` | 200 | **500** |
| `/get-started` | 200 | **500** |
| `/data` | 200 | **500** |

So this branch makes `/plugins/subgroups` and the blueprints-count endpoint hard build dependencies. It also partly neuters #5637: in merged code `home/Blueprints.astro` awaits `fetchTotalBlueprintsCount()` before the carousel data, so the `:not(:has(.blueprint))` guard added over there never gets a chance — the page throws at the heading count.

## What I'll change here

A floor constant resolves both concerns at once. Retry (from #5637) covers the transient case, which is the common one; the floor covers the residual. Neither `0+` nor a red build stays possible.

So once #5637 is in, this branch:

1. Deletes `$fetchApiCachedWithRetry` and its two constants, and points `pluginCount.ts` and `blueprintCount.ts` back at `$fetchApiCached`. Retry lives in `internalFetch` after #5637, so this is redundant — and leaving both nests retries 3×3 = 9 attempts per dead URL. The version here is also weaker: it breaks out on any status below 500, so it never retries 408/425/429, and it ignores `Retry-After`.
2. Replaces the two rethrows with the floor.
3. Takes the `FeaturePlugins.astro` deletion over #5637's edit to it — the file has no importers on main, so the deletion is still correct.
4. Reconciles `getSections(totalPlugins: string)` here against `number` in #5637. The formatted-string contract wins; that's the point of this PR.

## Why this order

#5637 first. It's smaller, it's pure infrastructure with no content decisions in it, and landing it makes this PR shrink rather than grow. The reverse order ships the weaker retry into `main` and then requires #5637 to remove it. This branch is also the one most exposed to the flake — open since 2026-08-30, ~16 Deploy runs, so per #5631 it rolls the dice roughly 8x more often than a same-day PR.

Worth knowing for whoever resolves: `src/utils/fetch.ts` auto-merges **without a conflict marker** and leaves `RETRY_BASE_DELAY_MS` declared twice, which fails `oxlint`. The other five conflicts (`pluginCount.ts`, `vs/[slug].astro`, `EveryPlugin.astro`, `PlanTable.astro`, `overview.astro`) plus the `FeaturePlugins.astro` modify/delete all surface loudly. And the trap isn't the duplicate declaration — lint catches that. It's resolving it by renaming the constant and leaving `$fetchApiCachedWithRetry` in place: green CI, 9 attempts per dead URL.

## Open question

What should the floor be? It's visible marketing copy on `/` and `/pricing`, so I don't want to pick it unilaterally. Current live value is 1,900. A conservative round number that can't read as stale — 1,500? — would do, since it only ever appears if both the request and its retries fail.

Related: the rethrow also fires on a *successful* 200 whose payload counts to zero. A floor covers that too, but it is a different signal — an empty-but-valid payload suggests a real API change rather than a blip, so it may be worth keeping that one loud.

iitzIrFan:
@elevatebart — re-review when you have a moment, scoped to the three commits since your approval rather than the whole branch:

- `4dcb4d1` — merge of main after #5637. Six conflicts, all resolved to this branch; `$fetchApiCachedWithRetry` dropped so retry lives only in `internalFetch`.
- `7930f28` — the 2.0 hero's hardcoded "2000+ Plugins". No test behind this one.
- `663c5a1` — **the one that actually needs your eyes.** You both approved "propagate the failure, fail the build loudly." This reverses that to a floor fallback: `PLUGIN_COUNT_FLOOR` 2000, `BLUEPRINT_COUNT_FLOOR` 690. Same question you raised on #5637, and you offered to wire up the floor if someone named the value — so this is me naming it and implementing it rather than handing it back.

Two calls inside that commit I'd particularly like challenged:

1. An empty-but-valid 200 now takes the floor too, rather than staying loud as we'd both mused. It keeps a separate error log naming it as a likely API change. Reasoning: if the rule is "never abort a build over a decorative number", a schema change shouldn't be the one exception — but that failure mode is now log-only.
2. The floored result still isn't memoized, so a mid-build recovery gets the real number, at the cost of a retry cycle per consumer during a sustained outage.

Verification: 379 tests, oxlint clean, astro check 0 errors, Build site green. With `API_URL` on a dead port, `/`, `/get-started`, `/data`, `/use-cases`, `/pricing` and `/overview` all render 200 showing 2,000+ and 690+ — on the pre-floor revision `/` and `/get-started` returned 500. The CI build log shows zero retries and zero floor fallbacks, so the deployed preview carries real values, not the floor.

@MartinRst — flagging separately in case your approval was on the copy side: 2,000+ is what the site would show visitors if the API were down through a deploy. Happy to change the number if that reads wrong to you.

vfanucci:
Thanks Irfan !

PR #5452, opened 2026-08-31
chore(plugins): drop the dead pre-canonicalUrl helpers and lock the payload order

vfanucci:
Follow-up to the plugin-canonical-URL stack (#5244 / #5245 / #5246), opened per @iitzIrFan's review on #5244 — the two items he waived rather than blocking the base PR on.

### Dead code

`canonicalUrl.ts` superseded the old mapping helpers. Nothing imports them any more outside their own tests (`git grep` over `main` returns only `plugin.ts` and `plugin.test.ts`):

- `buildSubGroupSegmentPredicate`, `buildPluginMappings` and the `PluginMappings` type are removed from `src/utils/plugins/plugin.ts`.
- The `slugify` import goes with them — `buildPluginMappings` was its last user in that file.
- `plugin.test.ts` only ever covered `buildSubGroupSegmentPredicate`, so the file is removed rather than emptied.

### Payload-order fixture

`buildPluginUrlIndex` ranks entries OSS-over-EE and subgroup-over-root, but an equal-rank tie is first-seen-wins, so the precedence rides on the order the `/plugins/subgroups` payload lists its entries in. Three cases in `canonicalUrl.test.ts` now build an index from a plugin root entry that repeats the classes its subgroups declare, and assert `clsToUrl` is identical with the root first, the root last, and the array reversed.

This is not a bug fix — the subgroup rank already wins today. It locks the behaviour so a reordering on the API side can't silently move canonical plugin URLs, which is what produced the 123 redirecting sitemap URLs #5244 fixed.

### Verification

- `vitest run`: 72 passed (8 files) — was 72 with the 4 deleted cases replaced by 3 new ones.
- `oxlint src/utils/plugins/`: clean.
- `astro check`: 0 errors, 0 warnings across 454 files.

No behaviour change: nothing outside the deleted tests called the removed helpers.


vfanucci:
Pushed a second commit that is not part of the agreed follow-up, flagging it explicitly.

`#5246` introduced a `parentSlug` local in the no-content redirect of `plugins/[...slug].astro`. `astro check` doesn't count the reads inside that block, so it reports `ts(6133): 'parentSlug' is declared but its value is never read` — and the Lint & Astro Check job runs reviewdog with `-filter-mode=nofilter -fail-level=any`, so that single hint fails the job on **every** open PR, not just ones touching this file. This PR's first CI run is the proof.

The workflow was added after `#5246`'s branch was last pushed, so it never ran there and the diagnostic reached main unseen.

Fix is to inline the expression. Redirect is unchanged — verified over the slug shapes:

| `splitRouteSlug` | before | after |
|---|---|---|
| `["plugin-airtable"]` | `/plugins` | `/plugins` |
| `["plugin-gcp","google-cloud-storage-gcs"]` | `/plugins/plugin-gcp` | `/plugins/plugin-gcp` |
| `["plugin-gcp","gcs","io.kestra.x"]` | `/plugins/plugin-gcp/gcs` | `/plugins/plugin-gcp/gcs` |
| `[]` | `/plugins` | `/plugins` |

`astro check` now reports 0 errors, 0 warnings, 0 hints across 454 files; Lint & Astro Check is green. Happy to split this into its own PR if you'd rather keep the cleanup single-purpose — but as it stands it unblocks the whole PR queue, so merging it here is the fastest path.

iitzIrFan (commented):
Could you please resolve the merge conflicts and update the PR before review? 
Thanks!

PR #5459, opened 2026-09-01
fix(ui): align rotating word baseline in hero heading (#5435)

### LINKED ISSUE #5435: "One Platform to _ All Your Workflows" - Changing word isn't inline with the rest of the words
### Describe the issue

It's slightly higher than the others

Image

mohitjeswani01:
## Summary
Fixes the vertical alignment of the rotating word ("Run" / "Control" / "Govern") in the homepage hero heading so it sits inline with the same baseline and size as the surrounding static text, instead of appearing bigger and offset upward.

## Root cause
`.animated-word` in `AnimatedWord.vue` used a fixed `1.2em` height/line-height that didn't match `.hero h1`'s actual line-height, pushing the rotating word out of alignment. Also added an explicit `line-height` to `.hero h1` at the mobile breakpoint, which previously had none and relied on a browser default — needed so both files' height math can match exactly at each screen size.

## Changes
- `src/components/home/AnimatedWord.vue`: `.animated-word` and `.word` height/line-height now exactly match `.hero h1`'s line-height at both mobile and desktop breakpoints
- `src/components/home/Hero.astro`: added explicit `line-height: 2.375rem` for `.hero h1` at the mobile breakpoint

## Screenshots
**Desktop**
image

**Mobile**
image

Closes #5435

### DIFF
--- a/src/components/home/AnimatedWord.vue
+++ b/src/components/home/AnimatedWord.vue
@@ -34,20 +34,25 @@
         display: inline-block;
         overflow: hidden;
         vertical-align: bottom;
-        height: 1.2em;
+        height: 2.375rem;
         position: relative;
         // force a fixed width to prevent layout shift; adjust as needed based on longest word
         width: 115px;
         @include media-breakpoint-up(md) {
+            height: 3.9375rem;
             width: 200px;
         }
     }
 
     .word {
         display: block;
-        height: 1.2em;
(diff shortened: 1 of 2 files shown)


iitzIrFan (approved):
Made a small follow-up to use `1lh`. 
LGTM now, thanks!

PR #5463, opened 2026-09-01
fix(styles): use white text for cookie consent accept button (#5434)

### LINKED ISSUE #5434: Cookie Accept button should have white text
### Describe the issue

Image

mohitjeswani01:
## Description
Fixes #5434

Changes the Cookie Consent "Accept" button text color (`--cc-btn-primary-color` and `--cc-btn-primary-hover-color`) from `var(--ks-content-primary)` to `$white` in `src/assets/styles/app.scss`. 

Previously, in light mode, `--ks-content-primary` evaluated to black text on the purple button background. Using `$white` matches the primary CTA styling used across the site.

## Screenshot
image


iitzIrFan (approved):

MilosPaunovic:
@iitzIrFan Can we merge this maybe?

iitzIrFan (commented):
yes @MilosPaunovic , safe to merge, small note from Claude for any further investigation. I manually investigated it yesterday and did another pass with the agent today. Merging it now.


> Verified this beyond the diff — it's isolated. --cc-* properties are declared in exactly one place in the repo (the #cc-main block in app.scss) and are read only by the vendored vanilla-cookieconsent stylesheet, so nothing outside the consent widget can be affected. Compiling app.scss before/after produces a diff of exactly these two declarations, and a computed-style diff across the rendered widget (consent + preferences modal, both themes) changes exactly four elements: "Accept", "Accept all", "Reject all", and the Accept label span. Toggles, links, secondary buttons and the panel background are untouched — the cc--darkmode aliases for the toggle knob/icon colours resolve against <html>, not #cc-main, so the override doesn't reach them.
> 
> Dark mode is a no-op (--ks-content-primary was already #ffffff there); light mode goes from 2.77:1 to 6.71:1, so this is a clear AA win. $white is the right choice here rather than --ks-content-primary-inverse, which would flip to black in dark mode while --ks-background-button-primary stays purple — and it matches what .btn-primary already does.

iitzIrFan:
Thanks @mohitjeswani01 🙌

PR #5468, opened 2026-09-02
docs(fix): generalize storage types

yuri1969:
Made the working less confusing for non-K8s folks.


aj-emerich (approved):
LGTM - thanks @yuri1969, makes sense to keep in general

PR #5489, opened 2026-09-03
docs(scripts): document the execution context file for script tasks

fdelbrayelle:
Documents the feature added in kestra-io/kestra#18958 (backend) and kestra-io/libs#40 (Python client). Neither PR touches docs, so without this the property and the `context` object ship undiscoverable.

New page: **Scripts → Execution Context** (`/docs/scripts/execution-context`), placed after Logging.

## What it covers

- **Why** — a Pebble expression pasted into script source is not valid Python, so a script kept as a Namespace File fails IDE linting, and the value breaks outright when it holds a quote or a newline.
- **How to enable** — `executionContext: true`, with a runnable flow example.
- **What the file contains** — table of the keys (`inputs`, `outputs`, `labels`, `vars`, `trigger`, `flow`, `execution`, `task`, `taskrun`), plus why `envs` and `globals` are deliberately excluded.
- **Reading it from Python** — the `context` object, nested traversal including objects inside lists, the mapping protocol, the bracket form for keys colliding with a method name, and the errors raised for a missing key or a missing file.
- **Reading it from another language** — it is plain JSON, so a Node example reads it with no Kestra client.
- **Secrets and lifecycle** — the file holds `SECRET` inputs in plaintext, which is why it is disabled by default, why the name is reserved and never collected by `outputFiles` / `cache.patterns` / `UploadFiles`, and why it is deleted once the task's commands are done.

## Accuracy checks

Three claims in the page I verified against the source rather than assuming:

- **Task coverage.** All 18 language tasks in `kestra-io/plugin-scripts` extend `AbstractExecScript`, where the property is declared, so they inherit it — the page lists them rather than only naming Python.
- **The `kestra:///` URI is not a readable path.** My first draft did `pd.read_csv(context.outputs.extract.uri)`, which cannot work. The example now maps the file in with `inputFiles` and uses the context for metadata, with a note making the distinction explicit.
- **Internal links.** `/docs/concepts/namespace-files`, `/docs/scripts/input-output-files` and `/docs/scripts/working-directory` all resolve — directory numeric prefixes are stripped, confirmed against `task-runners/04.types` → `/docs/task-runners/types`.

Frontmatter matches the `docs` collection schema in `src/content.config.ts` and the sibling pages' shape. `<ChildCard />` on the section index picks the page up automatically, so no nav file needed changing.

## Merge order

Worth holding this until kestra-io/kestra#18958 lands, so the docs do not describe a property that is not yet released. Happy to adjust the wording if the property shape changes — it is still an open question on that PR whether it stays a bare `executionContext: true` or moves to a nested `enabled:` like `namespaceFiles`.

### DIFF
--- a/src/contents/docs/16.scripts/09.execution-context/index.md
+++ b/src/contents/docs/16.scripts/09.execution-context/index.md
@@ -1,0 +1,134 @@
+---
+title: "Execution Context in Scripts: Read Metadata as Data"
+h1: Read the Execution Context from Python Scripts
+sidebarTitle: Execution Context
+icon: /src/contents/docs/icons/dev.svg
+description: Read inputs, outputs, labels and variables inside a script from a JSON file instead of interpolating Pebble expressions into your code.
+---
+
+Read the execution metadata from inside your script.
+
+## Why read the context as data
+
+A script task can reach the execution metadata through Pebble expressions:
+
+```python
+csv_path = "{{ outputs.extract.uri }}"
+```
+
+That works, but the expression is not valid Python, so a script kept as a [Namespace File](/docs/concepts/namespace-files) and edited in an IDE fails linting. It also mixes orchestration into business logic, and it breaks outright when a value contains a quote or a newline, because the value is pasted into your source before the interpreter sees it.
+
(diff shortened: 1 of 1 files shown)


aj-emerich:
Cherry picking into the 2.0 mega

PR #5520, opened 2026-09-08
build: enable Astro incremental static builds

elevatebart:
Turns on [`experimental.incrementalBuild`](https://docs.astro.build/en/reference/experimental-flags/incremental-build/). Astro skips re-rendering a static page when its `cacheKey` and its module dependency graph both match the previous build. Pages without a `cacheKey` are rendered every time, so the flag alone does nothing: the work is in the keys.

### What changed

- `astro` 7.0.0 -> 7.3.1 (`@astrojs/markdown-remark` to its 7.3 peer). The flag landed in 7.2, so the pinned 7.0.0 lockfile could not use it.
- `src/utils/incrementalCacheKey.ts`: a key is the entry digest plus a scope digest of the build-time inputs a page renders but does not own.
- Keys on the collection routes: docs, blogs, resources, legal, vs, orchestration, customers, plus the three raw-markdown twins.
- `node_modules/.astro` cached in the deploy workflow. It holds the manifest and the previous build's page output, and restore copies from there into `dist`, so `dist` itself does not need caching.

### Why the keys are not just the entry digest

A doc page renders more than its own file: the sidebar, prev/next and `<ChildCard />` all read every doc's frontmatter, and the shared layout bakes in the latest docs version fetched from the API. Keying on `doc.digest` alone would ship a stale sidebar the first time a page is renamed. So:

| Input | Digest used | Effect |
| --- | --- | --- |
| The page's own file | `entry.digest` | Body edit invalidates that page only |
| Sibling frontmatter (sidebar, cards, similar resources) | `collectionMetadataDigest` | Any frontmatter or membership change invalidates the collection |
| Sibling bodies (related customer stories) | `collectionContentDigest` | Any change in the collection invalidates it |
| Latest docs version from the API | `layoutDigest` | A release invalidates every keyed page |

Entries whose loader reports no digest get no key and are always rendered, rather than being cached on a guess.

Practical shape for this repo: editing the body of a doc or a blog post re-renders that one page, so a typical content PR should hit the cache for nearly everything. Adding, renaming or retitling a doc invalidates all 518 of them, because the sidebar genuinely changed.

### Where the cache actually hits

Preview builds set `NO_IMAGE_OPTIM` and `PREVIEW`, which change image URLs and gated content. Neither is part of Astro's config hash, so a shared cache would restore production HTML into a preview deploy. The key carries a `BUILD_VARIANT` segment to keep them apart, and the consequence is worth being blunt about: **main never builds the preview variant, so the first build of any PR is always cold.** The wins are pushes to `main` (each deploy reuses the previous one) and the second and later pushes within a PR. There is no way around that split without shipping wrong image URLs.

### Measurements

Three deploy runs on this PR, same workflow, same runner class:

| Run | Commit touches | Build step | Astro `Server built in` | Prerender phase |
| --- | --- | --- | --- | --- |
| [1](https://github.com/kestra-io/docs/actions/runs/34208405922) | everything (cold, no cache) | 4m36s | 4m33s | 2m23s |
| [2](https://github.com/kestra-io/docs/actions/runs/34209441840) | remark plugins, i.e. every content page's module graph | 3m56s | 3m54s | 2m31s |
| [3](https://github.com/kestra-io/docs/actions/runs/34210226152) | workflow only, nothing Astro compiles | 3m21s | 3m18s | **1m36s** |

Run 2 is the instructive one: its 39s saving came from the restored content store, not from skipped pages, because touching a remark plugin changes the dependency hash of every markdown route. Run 3 is what a normal content PR looks like: prerendering drops from 2m23s to 1m36s, about 35%, and the whole build from 4m36s to 3m21s.

The runner log is truncated to its last 5,000 lines, so exact page counts are lower bounds. In that window run 3 restored 894 pages (docs, resources, vs, legal) and re-rendered 144, nearly all of them routes with no `cacheKey`: `/use-cases`, `/features`, `/partners`, `/videos`, the sitemaps.

Cost: the saved cache is **~665 MB**, mostly the page copies that make restoring possible. Against GitHub's 10 GB repo limit that is roughly fifteen entries before LRU eviction starts.

### Open item

The 39 `/orchestration/*` pages have a `cacheKey` but re-rendered in run 3 anyway. That route's key folds in a digest of the blogs collection (the page renders the blog list), and something in that digest is evidently not stable across builds, most likely an `image()` field in the blog frontmatter. Worth chasing before anyone relies on that route being cached; it does not affect docs, blogs or resources, which do restore.

### Verification

`astro check` clean at 7.3.1, 292 unit tests pass, and a throwaway 7.3.1 project confirms a second build restores keyed pages and that a changed key re-renders them. The full site build could not run in the sandbox (prerendered pages fetch `api.kestra.io`, which was blocked), hence the reliance on CI.

Note for sequencing: #5519 also bumps `@astrojs/markdown-remark` and moves `@astrojs/mdx` to ^8. Whichever of the two lands second needs a lockfile resolution.

https://claude.ai/code/session_017kv1o69jUtuvZJEvBUvoCo


elevatebart:
Closing: folded into #5519, which carries the same change (the cache-key util and its tests, the keyed routes, the `incrementalBuild` flag and the workflow cache step) on top of its dependency bump.

The measurements and the open item on `/orchestration/*` are in this description if they are useful over there.

---

elevatebart:
Landed via #5519 instead, so closing this as superseded rather than stale.

All three commits here (`8c9b56b`, `0b98fba`, `6e1b05f`) were cherry-picked onto `deps/20260908` and squashed into main as 912271a. Both branches touched `package.json` and the lockfile, so one of them had to resolve against the other; folding this one into the dep bump meant resolving it once. The `astro` 7.0.0 to 7.3.1 bump this PR needed is subsumed by that branch's wider bump.

Two things measured on `deps/20260908` that are worth carrying forward, since they qualify the numbers in the description above:

**The saving is narrower than it looks.** Four Deploy `Build` runs on that branch: 6m08s without incremental, 4m00s cold, 3m04s on an exact-key hit, and 4m44s on the first build after merging `main` in. Only the 4m00s to 3m04s pair controls for anything, being the same commit re-run with nothing but the cache differing, and that is a best case rather than a typical one. The 4m44s run did restore from the `astro-build-preview-<pr>/merge-` prefix, but `main` had brought 11 commits, so most page keys had moved on and the 10s restore plus 7s save became overhead on a near-cold build. The same workflow has ranged 3m04s to 6m08s across five runs, with `Install dependencies` alone swinging 19s to 41s on identical work, so the noise band is wider than the ~56s effect.

**The cache key is not the branch ref.** On a `pull_request` event `github.ref_name` is `<pr>/merge` and `github.sha` is the ephemeral merge commit, so entries key as `astro-build-preview-<pr>/merge-<merge-sha>`. It works, since the restore prefix matches within a PR and resolves to `astro-build-production-main-` on main. But that merge SHA changes whenever **main** moves, not only when the PR is pushed, so the exact key misses on every base-branch commit and entries accumulate faster than the push count. At ~700 MB each against a 10 GB budget that is roughly fourteen entries before LRU eviction starts, and an evicted entry is a silent cold build.

The open item from the description, the 39 `/orchestration/*` pages that carry a `cacheKey` but re-render anyway, is untouched and still worth chasing.

---

PR #5521, opened 2026-09-08
fix(blueprints): degrade gracefully when the API can't build a graph

elevatebart:
### The 500

`/blueprints/ai-github-daily-healthcheck` returns a 500. The cause is upstream:

```
GET https://api.kestra.io/v1/blueprints/ai-github-daily-healthcheck/versions/latest/graph
422 Invalid entity: Unrecognized field "chatConfiguration"
    (class io.kestra.plugin.ai.domain.ChatConfiguration), not marked as ignorable
```

The API's plugin version is behind the blueprint's, so it can't deserialize the flow well enough to build a graph. The blueprint's own payload is fine. The docs page fetched the graph outside a try/catch, so one unhappy blueprint took the entire page with it.

This will happen again on every blueprint that lands ahead of the API's plugin jars, so the page should survive it.

### What changed

- `$fetchApiCachedOptional` in `src/utils/fetch.ts`: resolves to `undefined` and logs a warning instead of throwing. For decorative data only.
- The graph, the tag list and the per-tag related rows now use it. The blueprint payload itself still redirects to `/404` on failure, unchanged.
- The `.topology` column always renders. Without a graph it shows a "Diagram unavailable" panel styled like the topology card, pointing the reader at the YAML, rather than collapsing and leaving a gap where a diagram should be.
- The three category-listing fetches (`/blueprints/{category}`) got the same treatment. Its grid already had an empty state, but the copy would have lied about a failure ("no blueprints tagged with this category yet"), so it now tells a fetch failure apart from an actually empty category.
- Dropped `BROKEN_BLUEPRINTS` from `src/pages/sitemap/blueprints.xml.ts`. It was a hardcoded list of blueprints excluded from the sitemap purely because their pages 500'd. Both entries now return 200 from the graph endpoint, and the page no longer cares either way.

Degradation order: graph fails, you get the placeholder; tags fail, you lose the header pills and the "More ..." rows and keep everything else.

### Verified locally

| URL | before | after |
| --- | --- | --- |
| `/blueprints/ai-github-daily-healthcheck` (graph 422) | 500 | 200, placeholder panel, no `.topology` island |
| `/blueprints/azure-blob-backup-retention-cutover` (graph 200) | 200 | 200, real topology |
| `/sitemap/blueprints.xml` | 200 | 200 |
| `/blueprints/ai`, `/blueprints/data`, `?q=` hit and miss | 200 | 200, unchanged |

With every optional fetch forced to throw, `/blueprints/ai` returns 200 with the "could not load blueprints" message and `/blueprints/ai-github-daily-healthcheck` returns 200 with the placeholder panel.

`tsc --noEmit` reports nothing new on the touched files (the repo has pre-existing errors in `versionedNavigation`).

### Not in scope

Server islands were considered and rejected for this: `server:defer` has no per-island error boundary, so a throwing island returns a failed response and renders nothing. It would swap a loud 500 for a silent blank slot plus a round trip. They'd help these pages for caching, not resilience, and that's a separate change.

### DIFF
--- a/src/pages/blueprints/[id].astro
+++ b/src/pages/blueprints/[id].astro
@@ -1,7 +1,7 @@
 ---
 export const prerender = false
 import { Icon } from "astro-icon/components"
-import { $fetchApiCached } from "~/utils/fetch"
+import { $fetchApiCached, $fetchApiCachedOptional } from "~/utils/fetch"
 import Layout from "~/components/layout.astro"
 import Header from "~/components/blueprints/Header.astro"
 import BlueprintHero from "~/components/blueprints/BlueprintHero.astro"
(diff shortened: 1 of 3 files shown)


brian-mulier-p (approved):
LGTM

PR #5532, opened 2026-09-08
perf(blogs): lazy-load blog card images on the listing page

elevatebart:
### The problem

`/blogs` rewrites to `/blogs/all`, which renders **all 197 blog posts** into the DOM. `BlogGrid` hides all but 8 with a `display:none` class, and `BlogCard` emitted no `loading` attribute. `display:none` does not stop an image fetch, so the browser downloaded every card image to show eight of them.

The Lighthouse Benchmark has been reporting ~14 s LCP and 15-20 s Speed Index for the Blog Index on `main`, while every other page sits between 0.75 s and 2.8 s. This is why. FCP (~0.6 s) and TBT (~20-95 ms) were always fine, which is what pointed at the request pile rather than at JS.

Measured on the built page: **213 `` tags, 211 image requests, 247.9 MB transferred.** The only two lazy images on the page came from the header.

Production has the same bug in milder form: `kestra.io/blogs` today pulls **202 images / 5.2 MB** (avg 26 KB) through `/cdn-cgi/image`. The customer-stories card (`src/components/stories/Card.vue`) already sets `loading="lazy"`, which is why the other card-grid listing pages never showed this.

The LCP element is the hero `section.mesh`, not a card image — it was just stuck behind the request pile. Lighthouse's LCP breakdown shows element render delay going **1227 ms → 28 ms**.

### The fix

Lazy-load every card image; keep the first one eager so the one image that is always above the fold is not deferred. Two lines of substance.

### Verification

`scripts/lighthouse-benchmark.mjs` against `wrangler dev --local`, same machine, both variants of the same build, `NO_IMAGE_OPTIM=true` to match the CI build:

| | before | after |
|---|---|---|
| `/blogs` LCP | 10.70 s | **6.08 s** 🟢 |
| `/blogs` Speed Index | 1.28 s | **0.72 s** 🟢 |
| image requests | 211 | **17** |
| image bytes | 247.9 MB | **13.2 MB** |

`/blogs` is the only page the benchmark flags as changed; the other 17 are unchanged. The local baseline (10.70 s) tracks CI's 14.0/14.5 s, and the other pages land in CI's 0.75-2.8 s range, so the two environments are comparable.

### Expect CI to improve but not reach parity

`/blogs` will remain the worst LCP on the benchmark, and that part is not real user-facing latency.

`NO_IMAGE_OPTIM=true` selects the Cloudflare adapter's `passthrough` image service, whose `transform` returns the input buffer untouched and deletes the format option. So the `getImage({ width: 800, format: "webp" })` call in `src/pages/blogs/[category].astro` is a no-op, and all 201 assets ship as original JPEG/PNG (avg 1.26 MB, largest 4.8 MB). The one eager card image is a 2.4 MB original; production serves it at 26 KB.

There is no way to benchmark the production image path locally or in CI: the `cloudflare` image service only emits `/cdn-cgi/image/` URLs that Cloudflare's edge resolves, so any `wrangler dev` run necessarily measures unresized originals. Worth knowing when reading Blog Index numbers in future runs.

### One thing to flag

The Blog Index visual-regression baselines (`tests/visual-regression.spec.ts-snapshots/Blog-Index-*-linux.png`) may need regenerating — `fullPage` screenshots and `loading="lazy"` interact badly, so below-fold cards can capture blank. Those tests only run via the manual `update-linux-snapshots` workflow, so nothing breaks here; someone may want to dispatch it after merge.


iitzIrFan (approved):
LGTM, approving.

**Verified locally.** The preview is behind Cloudflare Access, so I checked the branch out in a worktree, ran `astro dev`, and inspected the server-rendered `/blogs/all`:

| | main | this PR |
|---|---|---|
| grid cards rendered | 200 | 200 |
| grid images `loading="lazy"` | 0 | 199 |
| grid images `loading="eager"` | 0 | 1 (card 0) |
| hidden cards (`index >= 8`) that are lazy | n/a | 189 / 189 |

The attributes are in the SSR HTML (BlogGrid is `client:idle`), so the fix applies on first paint without waiting for hydration. A lazy image inside a `display: none` box has no layout box and is never fetched, which is exactly what stops the 200-image download pile.

**Optional nit:** the grid is two columns from `md` up, so cards 0 and 1 are both above the fold on desktop. `:eager="index < 2"` would be slightly more precise. Low impact since browsers fetch in-viewport lazy images almost immediately; fine to leave as is.

**Post-merge:** as flagged in the description, `tests/visual-regression.spec.ts` uses `networkidle` + `fullPage` without scrolling, so the Blog Index baselines will capture blank below-fold cards. Someone should dispatch `update-linux-snapshots` after this lands.

Follow-up idea, out of scope here: slicing `filteredBlogs` to `visibleCount` in the `v-for` instead of hiding 189 cards with a class would remove the root cause and also fix the snapshot interaction.

PR #5552, opened 2026-09-09
fix(two-zero): serve the flattened hero gradient to Firefox only

elevatebart:
Fixes the banding that appeared on the `/two-zero` hero after #5540 flattened `hero-bg.svg` to WebP.

**Superseded by #5553**, which deletes the raster instead of choosing which engine gets which one. Prefer that; this PR is the narrow fix.

### What the artifacts actually are

VP8 macroblock terracing, not 1-LSB contouring. The ramp goes piecewise-constant in 16x16 blocks and the block edges line up into a staircase, worst across the dark upper-left where the gradient is shallowest. Each step is only 1 to 2 code values, but the eye integrates over a whole block, so it reads as a hard edge.

At 5x gain, same crop of the upper-left, top to bottom: original SVG, shipped WebP (33KB), WebP re-encoded at q95 (52KB), CSS `radial-gradient` equivalent. Only the two vector paths are clean.

### Why dither was the wrong fix

The first commit here added a noise layer, and the run-length statistics said it worked (longest flat run 203px to 5px). They were measuring the wrong thing. Zero-mean noise randomizes individual pixels, not block means: 3/255 of amplitude averaged over a 16x16 block shifts the mean by 0.19/255, against a step of 1/255. The staircase survives, which a 5x-gain crop shows plainly and a histogram of run lengths does not. Re-encoding is no better, already established at q95 in #5540.

### The change

Only Gecko re-rasterizes SVG-as-image with filters on every raster-scale change, which is what made the hero stagger there in the first place. So only Gecko gets the flattened asset:

- `hero-bg.svg` is restored and is what every engine gets by default.
- A Gecko-only `@supports` swaps in `hero-bg.webp` and enables the grain.
- The grain layer stays on the WebP path. It does nothing for the terracing, but it does fix the fine contouring in the same asset. 2.1KB, static, no blend mode, so nothing re-rasterizes and the drifting backdrop keeps its compositor-only animation.

`hero-squares.webp` keeps its flattening from #5540: only its alpha is ever read, so there is no ramp in it to terrace.

### The `@supports` condition, and a correction

An earlier revision of this PR used `@supports (-moz-orient: inline) or (-moz-appearance: none)`, on the reasoning that two Gecko-only properties are safer than one. That was the bug, and the claim in this description that both are false in Chromium was true of the source and false of the build. The emitted CSS read:

```css
@supports (-moz-orient:inline) or (appearance:none){ .backdrop{background-image:var(--heroBgFlat)} }
```

`-moz-appearance` came out of the build as `appearance`, which every engine supports, so the condition was always true and Chrome resolved `.backdrop` to the WebP on the preview. `-moz-orient` survived untouched, having no unprefixed name, so `93f5bac` narrows the condition to that alone. Verified through the build this time rather than from source: the emitted rule is Gecko-only, Chromium on that stylesheet resolves the backdrop to the SVG, and Chrome on the redeployed preview agrees.

Worth taking as a general lesson: a `@supports` condition on a `-moz-`-prefixed property is only Gecko-only if the property has no unprefixed equivalent for the build to rewrite it into.

### Measured

Longest flat run and share of a scanline flat over 8px, blue channel, 3 scanlines at 2000px wide:

| | longest | flat ≥8px |
|---|---|---|
| WebP, as shipped | 203px | 87% |
| WebP + grain (Firefox path) | 5px | 0% |
| SVG (every other engine) | 16px | 9% |

### Remaining caveat

Firefox is still served the terraced WebP here. That is this PR's trade, not a defect: it buys clean rendering for every other engine and keeps the Firefox scroll fix from #5540. Removing the terracing for Firefox too means removing the raster, which is #5553.

https://claude.ai/code/session_01JQW8A12pBdMRGpcg2GERuP


elevatebart:
Closing unmerged in favour of #5553, merged as 91bf50f.

That deletes `hero-bg.webp` rather than choosing which engine gets it, so both the grain layer and the Gecko-only `@supports` here are moot. The `@supports` finding is worth keeping in mind though: a condition on a `-moz-`-prefixed property is only Gecko-only if the property has no unprefixed equivalent for the build to rewrite it into. `-moz-appearance` came out of the build as `appearance` and handed the WebP to every engine; `-moz-orient` survived intact.

---

PR #5561, opened 2026-09-09
fix(visual): unblock WebKit snapshots and bound the workflow

elevatebart:
## What was happening

[Run 11](https://github.com/kestra-io/docs/actions/runs/34344429871/job/102442551807) of `Update Linux Visual Snapshots` had to be cancelled by hand after 26m52s. All 18 `desktop` tests wrote in 2m38s, then **16m45s of complete silence** until someone killed it. Runs 10 and 11 both died this way; run 9 in April was the last success.

Two separate problems, and the second is the real one.

## 1. The root cause: CSP upgraded every subresource to https

`content-security-policy.config.js` sets `upgrade-insecure-requests`, and `src/middlewares/worker.ts` only stripped it when `import.meta.env.DEV`. But `wrangler dev` serves the **production** build, so `DEV` is false there and the visual tests received the directive over `http://localhost:8787`. Every subresource got rewritten to `https://localhost:8787`, nothing serves TLS on that port, and the page never reached `load`.

The trace from run 12 shows it exactly: the HTML came back in **7ms**, then all 19 subresources sat `cancelled` with zero bytes and no content type until the navigation timeout killed them.

Chromium survives because it exempts `localhost` from the upgrade as a potentially-trustworthy origin. WebKit has no such carve-out, which is precisely why all 18 desktop tests passed while every `tablet` and `mobile` test failed.

Worth noting that `import.meta.env.DEV` could never be true in that file at all: `astro.config.mjs:43` selects the **node** adapter for `astro dev`, so the Cloudflare worker entry never runs during dev. The guard was dead code, which is why nobody noticed the header was always being sent. Two sibling guards in the same function were dead for the same reason and are fixed here too: `x-frame-options` always emitted `DENY`, and `strict-transport-security` was always sent, including over http where browsers ignore it.

The fix keys all three, and the localhost allowlist, off the request scheme. Production is https, so the emitted headers there are unchanged.

### A narrower version was tried and does not work

Scoping the check to loopback hostnames (`url.protocol === "http:" && new Set(["localhost","127.0.0.1","[::1]"]).has(url.hostname)`) looked strictly better: same carve-out Chromium implements, no behaviour change for any non-loopback request. Run 14 says otherwise — it failed 34/54, identical to the no-fix baseline, so the predicate evaluates false under `wrangler dev`.

Since run 13 proves `url.protocol === "http:"` is true there, `url.hostname` must be something other than those three values, but I could not observe what without instrumenting a run. Left at the scheme check, which is verified. Anyone who wants the narrower predicate should log `request.url` from the worker under `wrangler dev` first.

## 2. Bounds, so a failure is never a silent hang again

- `timeout-minutes: 30` on the job, `globalTimeout` of 20 minutes in Playwright
- `timeout: 90_000` per test, plus `navigationTimeout` and `actionTimeout`
- `retries: process.env.CI ? 1 : 0`, and `trace: "retain-on-failure"`. One retry absorbs a crashed or OOM-killed browser process without a manual re-dispatch; `globalTimeout` covers a wedged one. Trade-off worth knowing: on a *systemic* failure like runs 12 and 14 (34 tests × 45s), the retry pushes the test step past `globalTimeout`, so it reports a partial failure list rather than all of it. The traces still upload, and a trace is what identified the CSP bug
- `github` reporter in CI for per-test progress, and the html report uploads on `failure()` — this is what made the root cause findable
- the snapshot artifact uploads on `always()`, so even a cancelled run hands back what it wrote

## 3. Fixed cost

- `fetch-depth` back to the default 1. The push is `HEAD:branch`, it never needed full history
- `~/.cache/ms-playwright` cached, installing `chromium webkit` only, which is all three projects use. Apt layers aren't cacheable, so a cache hit still runs `install-deps`. The save step carries `&& success()`, because an explicit `if:` replaces the implicit `success()` guard and a partial cache under a deterministic key would poison every later run
- `workers: 2`
- `@playwright/test` pinned exactly, so a floating `^1.61.0` can't make the next regression untraceable

The build stays uncached deliberately: prerendering fetches `api.kestra.io`, so the output isn't a pure function of the source tree and a content-hash key would be wrong as well as usually missing.

## Results

Verified by dispatching the workflow on this branch four times.

| | run 11 | run 12 | run 13 | run 14 |
|---|---|---|---|---|
| version | before | bounds only | bounds + scheme check | bounds + loopback check |
| outcome | cancelled after a 16m45s hang | failed, 20/54 | **success, 54/54** | failed, 20/54 |
| job total | 26m52s | 22m08s | **11m33s** | 22m25s |
| test step | 19m23s, killed | 15m03s | **4m43s** | 14m36s |

This PR's head carries the run 13 version of the CSP fix, plus the review fixes above. Those touch only failure paths and two response headers that don't affect rendering, so they have not been re-verified by a dispatch.

Run 13 breakdown: setup 1m53s, build 4m41s, tests 4m43s, upload and commit 16s. The build is now 40% of the run and the only place a further speedup can come from. Note run 14 landed on a slower runner (checkout 1m17s vs 28s), so treat cross-run setup timings as ±1min.

Run 13's baselines were sanity-checked rather than assumed: WebKit snapshots moved a few percent in both directions (Home tablet 1596581 → 1500995 bytes) and the smallest file was 207KB, so they were genuine re-renders rather than blank pages. Run 13 pushed that refresh as `d2bd782`; it has since been dropped so this PR stays to the fix. Dispatch the workflow on `main` after merge and it will open a snapshot-refresh PR you can review on its own.

## Two things to decide separately

Run 13's push came from `github-actions[bot]` with the default `GITHUB_TOKEN`, which by design does not re-trigger workflows. `Lint & Astro Check`, `Publish to Cloudflare Workers` and `Lighthouse Benchmark` were green on the commit before it and never ran on `d2bd782`. If any of those are required checks, this workflow will block every PR whose branch it pushes to, and the push needs a PAT or App token. That behaviour predates this PR and only starts biting now that the workflow reaches its push step, so I have left it alone rather than guess at your token setup.

Unrelated to this diff but visible on it: the latest Lighthouse run scores `/blogs` at **0** with every Core Web Vital at exactly `0.00s`, five of which the bot labels 🟢 improved. Zeros across every metric are a failed run, not a fast page. `/blogs` has measured LCP 12.16s, 12.53s, 12.54s and 12.56s across the four prior runs, an order of magnitude worse than any other page, so it was already close to Lighthouse's patience. Nothing in this diff can stop a page loading, and the previous head scored it 75. Worth its own issue, along with the reporting bug that makes a failed page read as the biggest win on the board.

Also unresolved: tablet and mobile baselines existed from April, so `upgrade-insecure-requests` was introduced after that. I have not identified which commit; the shallow clone in my environment cannot see it.

https://claude.ai/code/session_015wYvDRdnWEvA8fF4htNKhd


elevatebart:
`Update Linux Snapshots` is red on this head ([run 12](https://github.com/kestra-io/docs/actions/runs/34348952569/job/102457205532)), dispatched deliberately to test the change. 20 passed, 34 failed, test step 15m03s, job 22m08s.

The failure is pre-existing, not introduced here: run 10 failed and run 11 hung on the same thing. What this PR changed is that it now fails visibly instead of sitting silent for 17 minutes until someone cancels. Clean exit, no orphaned browser processes, both artifacts uploaded, and the commit steps correctly skipped so no baselines were written.

What the run establishes:

- All 18 `desktop` (chromium) tests pass against the same `wrangler dev` server.
- Every `tablet` and `mobile` (WebKit) test fails at `page.goto`, 45s, `waiting until "load"`.
- Exactly one page succeeds in WebKit, and it succeeds in **both** projects: `/plugins/core/debug/io.kestra.plugin.core.debug.return`, the lightest page in the sample. So this is deterministic, not a race.
- No orphaned `WPEWebProcess` at cleanup this time, so WebKit is not deadlocked. It is too slow to reach `load` on anything but the lightest page. Run 11's "hang" was the same slowness with no timeout to stop it.

Also settled: `waitUntil: "networkidle"` was not the cause. Swapping it for `load` + `document.fonts.ready` changed nothing for WebKit.

Not pushing a fix yet, because I cannot read the trace from this environment: the `playwright-report` artifact lives on `productionresultssa7.blob.core.windows.net`, which the egress proxy here blocks. The traces are `retain-on-failure` and present in that artifact on the run, so they are readable locally. Without them I would be guessing at which request stalls, and a speculative push costs a CI cycle.

The two directions worth testing, in order of how much they'd tell us:

1. Serve the prerendered `dist` with a static server for the visual tests instead of `wrangler dev`. `wrangler.jsonc` sets `run_worker_first: true`, so every asset request currently goes through a single-threaded workerd, which fits per-page cost scaling with subresource count. Risk: the worker also sets CSP and handles redirects, so a static server changes what is rendered.
2. Raise the WebKit navigation budget (120s) on the two WebKit projects only, to confirm slow rather than stuck. Cheap to prove, but at ~34 pages it puts the run well past the 20-minute `globalTimeout`, so it is a diagnostic, not a fix.

The timeout and artifact changes in this PR stand on their own and are what made the above diagnosable at all. Happy to split the WebKit root cause into its own PR if that's cleaner.

---

iitzIrFan (changes requested):
## Review: Request Changes

The root-cause analysis is excellent and the overall approach is correct. Two issues need to be fixed before merging; two more are worth discussing.

---

### Must fix

**1. `Save Playwright browsers` step runs even when the browser install fails**

In GitHub Actions, an explicit `if:` expression **replaces** the implicit `success()` guard rather than ANDing with it. So:

\`\`\`yaml
- name: Save Playwright browsers
  if: steps.browser_cache.outputs.cache-hit != 'true'   # ← no success() guard
\`\`\`

If `npx playwright install --with-deps chromium webkit` fails mid-download (network error, disk pressure), this step still fires and writes a partial `~/.cache/ms-playwright` under the deterministic key. GitHub cache entries are write-once per key, so every subsequent run restores the broken entry, skips the install, runs only `install-deps`, and fails with "browser executable not found". The cache stays poisoned until `package-lock.json` changes.

Fix:
\`\`\`yaml
if: steps.browser_cache.outputs.cache-hit != 'true' && success()
\`\`\`

---

**2. Two `import.meta.env.DEV` guards in `worker.ts` were not migrated (lines 44 and 63)**

The PR's commit message correctly explains that `DEV` is always `false` in `wrangler dev`. The CSP fix is right, but the same dead-code pattern remains in the same function:

- **Line 44**: `x-frame-options` is always `DENY` under `wrangler dev` (should be `SAMEORIGIN`)
- **Line 63**: `Strict-Transport-Security: max-age=31536000` is emitted over HTTP (ignored by browsers, but inconsistent with the fix's intent)

Both should use `!isInsecureOrigin` for consistency with the pattern introduced in this PR.

---

### Worth discussing

**3. `document.fonts.ready` doesn't cover fonts injected after `load`** (`tests/visual-regression.spec.ts`)

`FontFaceSet.ready` resolves for fonts already in the set at evaluation time. A page with lazily-loaded stylesheets can inject `@font-face` rules after this resolves, producing fallback-font screenshots that differ from the baseline. The 500 ms static wait may absorb this on fast runners but isn't guaranteed. If spurious diffs start appearing after merge, this is the likely cause.

**4. `retries: 0` is now unconditional** (`playwright.config.ts`)

The rationale ("nothing to be flaky about") holds for deterministic content but not for transient browser-process crashes or OOM on the runner. A single crash now requires a manual re-dispatch. `retries: process.env.CI ? 1 : 0` would preserve the improvement while absorbing runner instability without multiplying hang time.

elevatebart:
Thanks — all four verified against the code before touching anything. Fixed in `9494f8a`, rebased onto your `main` merge.

### 1. Cache-save guard — fixed

Confirmed, and the poisoning consequence is the part that makes it matter. `&& success()` added, with a comment saying why so nobody "simplifies" it back.

I checked the other conditional steps in that workflow while I was in there. `Install Playwright browsers` and `Install Playwright system dependencies` have the same missing guard but are harmless (they'd just fail again on an already-failing job, writing nothing persistent), and `Upload Linux snapshots artifact` / `Upload Playwright report` use `always()` / `failure()` deliberately. Left those alone rather than widen the diff; say the word if you want `success()` on the two install steps for consistency.

### 2. Remaining `DEV` guards — fixed

Both confirmed. Your line numbers were 44 and 63, they're at 51 and 69 on this head, same code:

- `x-frame-options` always emitted `DENY`
- `strict-transport-security` was always sent, including over http

Both now key off `isInsecureOrigin`. I also grepped the whole file to be sure there wasn't a third one hiding: `import.meta.env` now appears only inside a comment.

### 3. `document.fonts.ready` and post-`load` font injection — you're right, not changed

`FontFaceSet.ready` returns a promise for the fonts pending *at that moment*; a stylesheet that injects `@font-face` afterwards starts a fresh loading cycle that the already-resolved promise knows nothing about. So the 500 ms wait is genuinely load-bearing here, which is not a good place to be.

I've left it, because the current pipeline produces correct baselines and I'd rather not add speculative complexity to a passing test on top of an already-broad PR. Recorded here as the first suspect if spurious diffs appear. The cheap hardening, if you want it now, is to re-await after the settle so anything injected during those 500 ms is covered:

```js
await p.goto(page.path, { waitUntil: "load" })
await p.evaluate(async () => { await document.fonts.ready })
await p.waitForTimeout(500)
await p.evaluate(async () => { await document.fonts.ready })
```

That still doesn't cover injection after 500 ms, so it's an improvement rather than a guarantee. Happy to push it.

### 4. `retries` — done, with one consequence worth naming

`retries: process.env.CI ? 1 : 0`. Your reasoning about crashes and OOM beats mine about determinism; a wedged process is now covered by `globalTimeout` rather than by having no retries.

The interaction to be aware of: on a *systemic* failure like runs 12 and 14 (34 tests × 45 s), one retry pushes the test step past the 20 minute `globalTimeout`, so it gets cut off mid-list instead of reporting every failure. I think that's an acceptable trade — the traces still upload, and a trace is what actually identified the CSP bug, not the failure list — but it does mean the next systemic breakage will produce a partial picture. Raising `globalTimeout` to 30 (with `timeout-minutes` to match) would buy it back if you'd rather not lose that.

Validated before pushing: 295 unit tests pass, oxlint clean, workflow YAML parses, `playwright test --list` still enumerates 54 tests. I have not re-dispatched the workflow on this head — the changed lines only affect failure paths and two response headers that don't influence rendering, but given this PR has already produced two wrong diagnoses from me, I'd rather run it once than assert it's fine. Want me to?

---

iitzIrFan (approved):
All three blockers resolved in the follow-up commit: `success()` guard on the cache-save step, both residual `import.meta.env.DEV` guards migrated to `isInsecureOrigin`, and `retries: process.env.CI ? 1 : 0` restored for runner crash absorption. The `document.fonts.ready` discussion point is noted and the 500 ms static wait is a reasonable mitigation for now. Good to merge.

PR #5562, opened 2026-09-09
ci(lighthouse): link report pages to the deployed preview, and retry unscored runs

elevatebart:
### Problem

The page links in both Lighthouse report tables pointed at `http://localhost:8787`, the `wrangler dev` server that only exists inside the CI runner. Anyone clicking one from the PR comment gets a 404 unless they happen to be running the site on that exact port, so the links were decoration.

Chasing that turned up a second issue, visible in this PR's own first report: `/pricing` was listed with a performance score of `0 🔻 -98` and every metric at `0.00 🟢`. The page had failed to load, Lighthouse returned an LHR with null category scores anyway, and `extractResults` flattened those nulls to 0. The run counted as a success, so the retry wrapper never saw it and a broken page was published as a real measurement.

### Changes

**Report links**

- `scripts/lighthouse-benchmark.mjs` already had an unused `BASE_URL_OUTPUT` env var for the link base. The workflow now sets it, so Lighthouse still measures `localhost:8787` while the links point at the Cloudflare Workers preview for the branch (`https://kestra.io` on `main`).
- Error rows linked a bare `/path`; they now get the base URL like every other row.
- The header read `Tested: <url> on <date>`, which stopped being true once the measured and linked hosts diverged. It now reads `Tested on <date> · links point to <url>`.
- The alias/URL derivation lived in `deploy.yml` and was about to be copy-pasted into `lighthouse.yml`, with `deploy.yml` rebuilding the URL from the slug in three separate `format()` expressions. It now lives in `.github/actions/preview-url`, which returns both `slug` and `url`. Output verified byte-identical to the old pipeline on `Feature/My_Branch-Name` and on this branch, and the `Truncate URL` job on this PR produced the same alias as before.

**Unscored runs**

- `runLighthouse` now rejects an LHR carrying a `runtimeError` or any null category score, which routes those runs through the existing 3-attempt retry. The discriminator is `score == null`, so a page that genuinely scores 0 with real metrics still reports as 0.
- A page that stays unscored across all attempts becomes an `❌ error` row and is listed with its cause under the scores table.
- Delta lookups skip baseline rows that carry an `error`, so the zeros stored for a failed page on `main` no longer read as a `+98` win on the next PR.

Tested against a stubbed `lighthouse` module: recovery on a later attempt, giving up after three, and the rendering of both cases. On the live run `/pricing` came back at 98 with real metrics.

### Notes for the reviewer

- `truncate-url-slug` keeps its job id and display name so any required-check config keeps matching. It gains a `sparse-checkout: .github/actions` step, since a local composite action needs the repo on disk, which costs the deploy gate a few seconds.
- The links are only live if the Deploy workflow's preview upload for the same branch succeeded. The two workflows are independent, so a failed deploy means 404s again, just at a more useful address.
- Worth knowing but out of scope here: score deltas are noisier than the ±10pt threshold. Home moved 91 → 70 and About Us 98 → 85 between two runs of near-identical code on this branch, so the arrows will keep crying wolf until the threshold rises or the job takes a median of several runs.
- This PR's own comment is the test: the links in the two tables should land on this branch's preview.

https://claude.ai/code/session_01MhpwwzyR3rLvWD3uZCMoo5


iitzIrFan (approved):
LGTM 🙌

PR #5572, opened 2026-09-09
test(lighthouse): stop the benchmark scoring the runner instead of the site

elevatebart:
Home, Docs Landing and Blueprints Landing kept moving 20+ points between consecutive `main` runs with nothing related in the diff.

## What was actually happening

The swings correlate whole-run, not per-page. Two runs 90 minutes apart, on different runners:

| run (UTC) | runner | Home | About Us | Docs Landing | Blog Index | Blueprints |
|---|---|---|---|---|---|---|
| 13:02 | `...659358` | 66 | 84 | 83 | 74 | 72 |
| 14:32 | `...659730` | 91 | 99 | 95 | 75 | 75 |

Every page in the slower run is worse by about the same factor. Lighthouse scales the observed trace but does not normalise for host CPU, so a slow runner inflates TBT and LCP directly. Those three pages show it because they sit mid-curve where the metric-to-score mapping is steepest, and they are the most TBT-heavy in the sample. `/pricing` at 98 clips at the ceiling and hides the same effect.

Three aggravating factors on top: one run per page, a baseline artifact recorded days earlier on a different machine, and `/blueprints` being the only `prerender = false` page in the sample, fanning out ~15 live `api.kestra.io` calls per request with no cache under Miniflare.

## Changes

**`scripts/lighthouse-benchmark.mjs`**

- Per-page run counts: Home 5, `/docs` 5, `/blueprints` 3, `/about-us` 3, `/docs/workflow-components/flow` 3, the rest 1. That is 32 page runs for 18 pages. The reported row is the run whose performance score is the median, so scores and Core Web Vitals still come from a single trace rather than a per-field average no run produced. `MULTI_RUN_COUNT` overrides.
- `benchmarkIndex` recorded per page, printed as `cpu=` on each console line and in the report header next to the baseline's. The ▲/▼ deltas are suppressed when the two runners differ by more than 10%, which is the main way the report used to invent regressions.
- The per-run scores for repeated pages go into `lighthouse-results.json`, so the spread stays inspectable after the fact.
- Warm-up on `/privacy-policy` before measuring, three requests, results dropped. It is the barest page in the tree, so it drags the workerd compile and shared layout assets through `run_worker_first` without pre-warming anything in the sample.
- The six `prerender = false` pages are measured first, the prerendered ones take the tail, and the report sorts back to page-sample order so rows stay comparable run to run. See the caveat below: this helped less than expected.
- Every Lighthouse result is validated before use. A `runtimeError` other than `NO_ERROR`, or a `first-contentful-paint` audit with no `numericValue`, throws so the page is retried, and an exhausted retry renders `❌ error`. This came out of a real run where Pricing published performance 0 with every metric at 0.00 while its a11y (92), best practices (73) and SEO (100) were normal: those categories audit without the trace, so the page had loaded and only the metrics were missing, and the old code turned an absent `numericValue` into a confident zero and a 🔻 -99.

**`scripts/api-fixture-server.mjs`** (new)

Record-and-replay proxy the benchmark builds against (`API_URL` points at it). `FIXTURE_PATHS` decides what is recorded, defaulting to `/v1/blueprints*` plus the exact paths `/v1/plugins` and `/v1/plugins/subgroups`, which is what the blueprints page's `fetchToolIndex` and `fetchCoreToolIndex` ask for. Recorded paths are served from `tests/fixtures/api`, fetched upstream and written there on a miss, and everything else is proxied untouched, so the rest of the build keeps talking to the real API and the fixture set stays bounded. Deliberately not the whole `/v1/plugins` tree: that is hundreds of build-time schema fetches for the prerendered plugin docs.

**`.github/workflows/lighthouse.yml`**

Starts the fixture server before the build, drops `global_fetch_strictly_public` from `wrangler dev` (it blocks the worker from reaching loopback, and only governs same-zone routing, which does not exist locally), uploads the recorded fixtures as the `api-fixtures` artifact, and fails the job if no blueprint call reached the fixture server. That last check runs after the report is posted, so a failure never costs the comment.

## What CI has confirmed

The loopback question is settled: with the compat flag dropped the worker reaches the fixture server, `hits 149, records 273, passthrough 1222, errors 0`, and the check step passes.

Four runs on four runners (CPU index 2503, 2508, 2320, 2335). The last column is after the multi-run list widened and the SSR pages moved to the front:

| Page | 15:46 | 16:28 | 16:44 | 22:34 | runs now |
|---|---|---|---|---|---|
| Home | 92 | 95 | 95 | 95 | 5 |
| Docs Landing | 89 | 97 | 93 | 97 | 5 |
| About Us | 98 | 84 | 84 | 98 | 3 |
| Flow | 85 | 89 | 92 | 92 | 3 |
| Blueprints Landing | 87 | 87 | 88 | 82 | 3 |

Home holds within 3 points across four runners, which was the goal. Docs Landing, About Us and Flow all improved once repeated.

Blueprints is the honest miss. It dropped to 80 on the run where the job grew to 32 page loads, and moving it from last to first recovered only 2 of the 7 points, so "the tail of the job is slower" does not explain it. The likelier reading is the reverse: an SSR page measured first pays its own cold render, compiling its route and reading its fixtures inside the traced request, where running late it had been warmed by every earlier render. Its FCP tracks that: 0.84 s last in a short job, 1.20 s last in a long one, 1.01 s first. The next thing to try is a discarded warm-up request per SSR route, which removes the cold render wherever the page sits.

## Follow-up needed

The committed fixture set covers the top-level calls only. A full run needs 303 files, so each run still records the missing 273 and uploads them as the `api-fixtures` artifact. Download it, drop the JSON into `tests/fixtures/api/`, and commit: `records` then drops to 0 and the page is genuinely frozen. Refresh instructions are in the README there.

## Cost and caveats

- Roughly 5 to 6 minutes added to a 13 minute job, most of it the repeated runs.
- Fixtures freeze the page content: a layout change that only appears at today's blueprint count stays invisible until they are refreshed.
- The pass-through is silent by design, so a newly added blueprint endpoint records itself on the next run, but that run measured a live call.
- The `main` baseline predates this change so it carries no CPU index, which is why the report says deltas are not CPU-checked. That resolves on the first merge to `main`.

## Testing

`node --check` and oxlint clean on both scripts. `buildMarkdown`, `compareBenchmarkIndex`, `median` and `runsFor` exercised against synthetic same-CPU, slow-CPU and no-baseline inputs; the result guard against a good run, a `NO_ERROR` code, a runtime error, the Pricing shape and a missing audit; the run plan and the reordering checked against the real page sample, so a mistyped path can neither silently fall back to one run nor quietly drop a page. The fixture server tested against a stub upstream: record, replay, passthrough, replay with the upstream killed, path containment for `/v1/blueprints/../../etc/passwd`, and the stats endpoint.

The remaining lever, not in this PR: the benchmark build leaves `PREVIEW` unset, so PostHog, GTM and HubSpot load over the real internet in every trace. Also visible in the report and unrelated to it: Blog Index LCP is 14 s.

https://claude.ai/code/session_01P5E2r1GCXnoUKpdqv389F9


iitzIrFan (commented):
merge conflicts Bart!

elevatebart:
Merged `main` in, conflicts resolved in `5101fc6`.

Only `scripts/lighthouse-benchmark.mjs` actually conflicted, and one of the hunks was worth a careful look rather than a pick-a-side: `main` added `assertScored` for pages that come back unscored, and this branch had added the same class of guard for pages that come back scored but with no trace metrics (the run where Pricing published `0` with LCP, FCP and TBT all at `0.00` while its a11y, best practices and SEO were normal). Those catch different failures, so they are now one function: runtime error, then any null category score, then a missing `first-contentful-paint` numeric value.

The rest resolved cleanly in both directions. `BASE_URL_OUTPUT` and the preview-url step from `main` are in, so report links point at the deployed preview; the baseline lookup now skips both errored rows and CPU-mismatched baselines; and the header keeps `main`'s wording plus this branch's runner CPU line.

Verified after merging: `node --check` and oxlint clean, the workflow still parses, the merged guard accepts a good run and rejects all three failure shapes, and the report still renders rows in page-sample order with public links.

---

iitzIrFan (approved):
**Approve** — nice diagnosis, and the merge-commit run confirms it end to end: fixture server reachable from the worker (`hits 149, records 273, errors 0`), Home/Docs/About-Us steady, CPU index in the header. I also ran the fixture server locally against a stub: record → replay → passthrough behave as described, and `/v1/blueprints/../../etc/passwd` lands inside the fixture dir.

Non-blocking, but I'd like these tracked:

1. **Blueprints isn't frozen yet.** 273 of 303 fixtures are still recorded live on every run, and the merge-commit run shows it (`perf=86, median of 72/86/87`). Can you commit the `api-fixtures` artifact from this run as a fast follow-up so `records` drops to 0?
2. **"Check the fixture server was used" can't fail.** The build's own `/use-cases` fetch hits `/v1/blueprints*` before Lighthouse starts, so `hits + records` is never 0 while the build succeeds. Snapshot `/__fixtures/stats` after the build and diff it after the benchmark if we want it to prove the SSR path.
3. **Single source of truth for pages.** `MULTI_RUN_PATHS` and `SSR_FIRST_PATHS` re-key the page list by path. A `runs` / `ssr` field on the entries in `tests/fixtures/page-sample.mjs` would keep the script and the visual-regression spec reading one list.
4. **Compat flags are now duplicated** between `wrangler.jsonc` and the workflow; a comment in `wrangler.jsonc` pointing at `lighthouse.yml` would save someone a confusing afternoon.
5. Nit: the proxy forwards GET only (method/body dropped). Fine for the benchmark, but worth a line in the header comment.

Heads-up for the team: fixtures are 3.7 MB now and roughly double once the set is complete. Acceptable to me for a CI-only fixture dir that oxfmt/prettier already ignore, but flagging in case anyone prefers an artifact-cache approach later.

elevatebart:
Thanks for the review, and for re-running the fixture server yourself. Four of the five are in `93d22e7`.

**2, the check that could not fail.** You are right, and it is the one that mattered: the build calls the blueprint API before Lighthouse starts, so `hits + records` was never 0 while the build succeeded. The counters are now snapshotted straight after the build and the final step requires the total to grow during the benchmark. It also prints the delta, so the log says how many calls the measured pages actually made.

**3, single source of truth.** `MULTI_RUN_PATHS` and `SSR_FIRST_PATHS` are gone; the entries in `tests/fixtures/page-sample.mjs` carry `runs` and `ssr` and the script reads them. That also removes the failure mode where a rename in one list silently degrades the other. The spec only touches `path` and `label`, so the extra fields are inert there. Verified the plan is unchanged at 32 runs over the same 18 pages, with one deliberate difference: within the SSR group the order is now the sample's, so Blueprints runs fifth rather than first, on a worker four renders warmer. Given the spread you spotted, that is probably the better place for it.

**4 and 5.** `wrangler.jsonc` now says the Lighthouse workflow drops `global_fetch_strictly_public` and why; the fixture server's header notes it forwards GET only, method and body dropped.

**1, the fixtures.** Agreed, and your `median of 72/86/87` is the sharpest evidence yet: the 72 is the first of the three runs, which is the one paying the cold render and recording the missing 273 fixtures live. Committing the artifact should collapse that spread as well as dropping `records` to 0. I cannot download the run artifact from where I am working, so that one needs someone with the browser open: grab `api-fixtures` from the latest run, drop it into `tests/fixtures/api/`, commit.

On the 3.7 MB: it roughly doubles when complete, and it is 4% larger than it needs to be because these are compact-serialised already. Worth revisiting as an artifact cache if it keeps growing, but committed fixtures are the only version that makes the page deterministic on a fork PR with no cache to restore from.

---

iitzIrFan (approved):
**Approve.** Both follow-up commits look right.

- The fixture check now measures what it claims: counters snapshotted after the build, and the benchmark has to add to them. That was the one that mattered.
- `runs` / `ssr` on the page sample is the right home for that metadata. I re-derived the plan from the new file: same 32 runs over 18 pages, same six SSR routes.
- Per-route SSR warm-up is a sensible, cheap answer to the cold render, and the cancelled run on 93d22e7 already showed Blueprints tightening to 81/83/84.

Only the fixture-set commit remains, and it needs a browser. I'll grab the `api-fixtures` artifact from the run on this head once it's green and push it to the branch, then `records` should read 0 on the next run. Merge as soon as CI finishes on 2096ddc.

iitzIrFan:
## Fixture set committed in [165bcf7](https://github.com/kestra-io/docs/pull/5572/commits/165bcf75649204715e738509d6c20a176522f861)

For future reference, here is what the fixture commit is and how it was produced.

**Source.** The `api-fixtures` artifact from [run 34451628852](https://github.com/kestra-io/docs/actions/runs/34451628852) on `2096ddc`, the first green run of the diff-based fixture check. Downloaded with `gh run download`, JSON dropped into `tests/fixtures/api/`, committed on top of `ea8a569` (Plugins Landing x3). Nothing else in the PR was touched.

**Contents.** 272 new files, bringing the set to 302. The 30 already committed came back byte-identical, so the recording is deterministic for the same upstream state. All 302 are valid `{ path, status: 200, contentType, body }` records under `/v1/blueprints*`, `/v1/plugins` and `/v1/plugins/subgroups`, named by the server's own `slug.hash8.json` scheme.

**Size.** 10.2 MB raw, more than the ~7 MB estimated above. Three files carry 7.8 MB of it:

| file | bytes |
|---|---|
| `plugins-subgroups.b55aa003.json` | 3,764,400 |
| `blueprints-plugin-io-kestra-plugin-core-version-latest.54ca6a23.json` | 2,791,893 |
| `plugins.e137dbf5.json` | 1,288,916 |

The remaining 269 files are under 135 KB each. Git packs the whole directory to roughly 3.3 MB, so the repo cost is modest. If the set keeps growing, an artifact cache is the alternative discussed above; committed fixtures were kept because they are the only version that is deterministic on a fork PR with no cache to restore.

**Expected effect.** `records` should read 0 on the next run and the Blueprints trace stops paying a live API round trip inside the measured request. To refresh: delete the JSON files, let the workflow run, download `api-fixtures`, commit (see `tests/fixtures/api/README.md`).

iitzIrFan:
## Confirmed: `records` is 0 on `165bcf7`

[Run 34453372865](https://github.com/kestra-io/docs/actions/runs/34453372865) on the fixture commit, final fixture-server stats:

| hits | records | passthrough | errors |
|---|---|---|---|
| 424 | 0 | 1316 | 0 |

Every `/v1/blueprints*`, `/v1/plugins` and `/v1/plugins/subgroups` call during the build and the benchmark was served from the committed set; nothing went upstream on those paths. Blueprints Landing came back `83 / 83 / 84` (median 83), a one-point spread, against `72 / 86 / 87` before the SSR warm-up and the fixtures landed.

All seven checks are green on the head and the approval stands. Nothing open from the review side; good to merge. The first `main` run after merge records the runner CPU index into the baseline, so PR reports get CPU-checked deltas from then on.

elevatebart:
Thanks for confirming `records: 0`. Heads-up that the head moved after your comment: `b1858b4`.

Your `83 / 83 / 84` is the tell. The spread is exactly what we wanted, but the level dropped from 86-88, and on a *faster* runner (CPU 2934, FCP 1.06 s against 0.84 s on the CPU-2500 runs). A uniform cost applied to every render fits that shape, and it turned out to be mine: `readFixture` parsed the file and re-stringified the body on every request, which nobody noticed while the big payloads were still being fetched live.

Measured against the committed set:

| path | size | before | after |
|---|---|---|---|
| `/v1/plugins/subgroups` | 3.6 MB | 100 ms | 13 ms |
| `/v1/plugins` | 1.2 MB | 22 ms | 5 ms |

The blueprints SSR path awaits both sequentially per render, so it was carrying ~125 ms of fixture-server overhead inside each measured request. `b1858b4` caches the serialised response per path; body verified byte-identical at 3,764,314 bytes, and a re-record drops the path's cache entry so recording still works.

Expect Blueprints back around 86 with the same one-point spread, which would make the whole exercise land: stable *and* honest. If CI on `b1858b4` disagrees, that is worth knowing before this merges.

---

iitzIrFan:
## Blueprints recovered on `b1858b4`

[Run 34454964403](https://github.com/kestra-io/docs/actions/runs/34454964403) on the replay-cache commit, against the fixture-commit run before it:

| commit | Blueprints perf | per-run | FCP | runner CPU | records |
|---|---|---|---|---|---|
| `165bcf7` fixtures | 83 | 83 / 83 / 84 | 1.06 s | 2934 | 0 |
| `b1858b4` cache | **86** | 86 / 87 / 86 | **0.93 s** | 2469 | 0 |

Back to 86 on a slower runner, one-point spread, first paint down 0.13 s. Fixture stats unchanged at `hits 424, records 0, errors 0`, and the diff-based check counted the same 50 blueprint calls during the benchmark, so the freeze held.

Two things from the same run:

- **Plugins Landing** at `runs: 3` came back `95 / 98 / 98` (median 98). The 368 ms TBT spike that prompted the bump reads as a one-off, not a regression.
- Blueprints' Speed Index carries a 🟢 in the posted report, the first improvement the CPU-checked comparison has flagged on that page.

All seven checks green on the head. Nothing outstanding from the review side; merge order is this one first, then #5584.

PR #5584, opened 2026-09-10
ci(lighthouse): run the sample on four runners off one build

elevatebart:
Stacked on #5572, which is where the fixture server, the warm-up and the median-of-3 come from. Merge that first.

## Why

The benchmark step took 9m56 as a single job on run 3750. Median-of-3 takes the sample to 34 page runs, so it was heading for half an hour on a docs PR.

## What

Three jobs: build once, measure in a 4-way matrix, merge and comment.

The shards get the build as a zstd tarball artifact instead of building their own. Four builds would cost four times the Astro minutes and, worse, put four different builds under measurement. `dist` is thousands of mostly-image files, so it is tarred before upload with `compression-level: 0`; the artifact actions walking the tree file by file is what makes that slow, not the bytes.

`SHARD_TOTAL` comes from `strategy.job-total`, so the matrix length is the only knob.

### Slicing is by measuring time, not page count

A page costs one run, or `runs` of them for the ones measured by median. Longest-processing-time first lands at 9/9/8/8 runs per shard. Round-robin by index would have put 15 of the 34 on one shard, and that shard would have set the wall clock.

### The CPU gate is now per page

`benchmarkIndex` is a property of the runner and there are four of them now, so one number for the whole run no longer describes it. Each `PageResult` already carried its own, so a page's deltas are hidden when its runner is more than 10% off the baseline's. On a single runner this is exactly what #5572 does today.

Tested against a synthetic slow shard: only that shard's four pages lose their deltas, and the header says how many.

### Script split

`scripts/lighthouse-report.mjs` is new and owns everything after the measuring: merge the shards, mark any page no shard reported rather than dropping the row, render the Markdown. `lighthouse-benchmark.mjs` is back to measuring one slice. `METRIC_DEFS`, the thresholds and the sharding live in `scripts/lighthouse-shared.mjs`.

### Smaller things

- the fixture-server check runs on the shards that measured `/blueprints`, which is where the SSR calls happen; the others never touch it
- the `api-fixtures` artifact uploads only when something was recorded, so it means "commit these" rather than appearing on every run
- the report job sparse-checks out and skips `npm install`: it needs the scripts and the page sample, not 800 MB of content or `node_modules`
- it keeps the name `Lighthouse Benchmark` so the existing status check still resolves

## Numbers

Roughly 17 min wall clock against ~32 serial, for about 46 runner minutes. Building in every shard would be quicker still (no serialized build job in front, ~13 min) at four times the build minutes and four builds under measurement, which defeats the point.

## What is not verified

The `dist` size and the shards' `wrangler dev` against an unpacked `dist` need CI to say. The 17-minute estimate assumes the download costs about a minute; the `Pack the build` step will report the real size on the first run. If it is much bigger than the ~5.1k assets in the deploy log suggest, building per shard is the better shape and this becomes the wrong trade.

Report merge, the missing-shard rows, the per-page CPU gate and the 9/9/8/8 split were verified locally against synthetic shard output.

https://claude.ai/code/session_015GEhSD6GqR4y8jL869sp9s

---
_Generated by [Claude Code](https://claude.ai/code/session_015GEhSD6GqR4y8jL869sp9s)_

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -59,8 +59,21 @@ jobs:
 
       # dist is thousands of files, most of them images. One tarball moves in a
       # fraction of the time the artifact actions take over the tree itself.
+      #
+      # .wrangler/deploy/config.json travels with it: that redirect is what
+      # points wrangler at the generated dist/server/wrangler.json. Without it
+      # wrangler falls back to wrangler.jsonc and bundles the worker from src/,
+      # where the adapter's virtual modules do not resolve, so the dev server
+      # never serves. It is gitignored, so a shard only has it if it is packed.
       - name: Pack the build
-        run: tar -I 'zstd -1 -T0' -cf dist.tar.zst dist
+        run: |
+          test -f .wrangler/deploy/config.json || {
+            echo "::error::.wrangler/deploy/config.json missing after the build."
+            echo "::error::wrangler dev would bundle src/ instead of the build."
+            exit 1
+          }
+          cat .wrangler/deploy/config.json
+          tar -I 'zstd -1 -T0' -cf dist.tar.zst dist .wrangler/deploy/config.json
 
       - name: Upload the build
         uses: actions/upload-artifact@v7
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -18,6 +18,7 @@ jobs:
     name: Build site
     if: startsWith(github.repository, 'kestra-io/')
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     permissions:
       contents: read
 
@@ -92,6 +93,7 @@ jobs:
     name: Lighthouse shard ${{ matrix.shard }}
     needs: build
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     permissions:
       contents: read
     strategy:
@@ -149,13 +151,19 @@ jobs:
         # the one in wrangler.jsonc): it blocks the worker from reaching the
         # fixture server on loopback, and it only affects same-zone routing,
         # which does not exist locally.
-        run: npx wrangler dev --local --compatibility-flags nodejs_compat &
+        # Output is captured: the process outlives this step, so whatever it
+        # prints while the next one waits is otherwise lost.
+        run: npx wrangler dev --local --compatibility-flags nodejs_compat > wrangler-dev.log 2>&1 &
 
       - name: Wait for dev server to be ready
+        # The dev proxy accepts connections while the worker is still compiling
+        # and holds the request until it is ready, so an unbounded curl blocks
+        # instead of retrying: without --max-time one hung probe is the job.
         run: |
           echo "Waiting for http://localhost:8787…"
           for i in $(seq 1 24); do
-            STATUS=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:8787 || true)
+            STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
+              --connect-timeout 2 --max-time 10 http://localhost:8787 || true)
             if [ "$STATUS" = "200" ]; then
               echo "Server is ready (HTTP 200)."
               exit 0
@@ -163,7 +171,11 @@ jobs:
             echo "  Attempt $i/24: HTTP $STATUS — retrying in 5 s…"
             sleep 5
           done
-          echo "Warning: server did not return 200 after 2 min; running Lighthouse anyway."
+          echo "::error::Dev server never returned 200. Measuring anyway would"
+          echo "::error::score a dead server, so this shard stops here."
+          echo "--- wrangler dev output ---"
+          cat wrangler-dev.log || echo "(no output captured)"
+          exit 1
 
       - name: Run Lighthouse benchmark
         run: node scripts/lighthouse-benchmark.mjs
@@ -233,6 +245,7 @@ jobs:
     needs: [build, lighthouse]
     if: ${{ !cancelled() && needs.build.result == 'success' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     permissions:
       contents: write
       pull-requests: write
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -11,13 +11,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lighthouse:
-    name: Lighthouse Benchmark
+  # The site is built once and handed to the shards as a tarball. Building in
+  # each of them instead would cost four times the runner minutes, and the
+  # measured pages have to come from one identical build to be comparable.
+  build:
+    name: Build site
     if: startsWith(github.repository, 'kestra-io/')
     runs-on: ubuntu-latest
     permissions:
-      contents: write
-      pull-requests: write
+      contents: read
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


elevatebart:
Green here means very little: `welcome` is the only check that ran.

`lighthouse.yml`, `astro-check.yml` and `deploy.yml` all trigger on `pull_request: branches: [main]`, and this PR targets `elevate-bot/loving-hawking-qjmcm1`. So the sharded workflow this PR rewrites does not run on the PR that changes it. The parts that need a real runner to confirm (the `dist` size, `wrangler dev` against an unpacked `dist`, whether the download per shard actually costs about a minute) get their first exercise when this merges into #5572 and that PR builds against `main`.

Base branch merged in at 5b75fe3 to pick up b1858b4; no overlap with the files here.

---

iitzIrFan (changes requested):
**Request changes** — one blocking issue in the `report` job; the rest of the PR reads well (shard split, per-page CPU gate and the merge all checked out against synthetic input, see below).

### The report job's sparse checkout drops `.nvmrc`

```yaml
sparse-checkout: |
  .github/actions
  scripts
  tests/fixtures/page-sample.mjs
sparse-checkout-cone-mode: false
```

In non-cone mode `actions/checkout` appends exactly these three patterns to `info/sparse-checkout` and nothing else, so no root file is included. The next step is `setup-node` with `node-version-file: .nvmrc`, and `getNodeVersionFromFile` throws when the file does not exist. The job fails before it merges a single shard, on every run.

I reproduced it against `origin/main` with the same three patterns written the way the action writes them:

| pattern list | `.nvmrc` | `.github/actions/preview-url` | `scripts` | `page-sample.mjs` | `src/` leaked |
|---|---|---|---|---|---|
| as in the PR | **missing** | present | present | present | yes (`src/scripts/*` matches the unanchored `scripts`) |
| anchored + `.nvmrc` | present | present | present | present | no |

Suggested list:

```yaml
sparse-checkout: |
  /.github/actions/
  /scripts/
  /tests/fixtures/page-sample.mjs
  /.nvmrc
sparse-checkout-cone-mode: false
```

The leading slashes are optional but stop `src/scripts` riding along. Nothing else in the job needs a root file: no `npm install`, and both `lighthouse-report.mjs` and `lighthouse-shared.mjs` import only from `../tests/fixtures/page-sample.mjs`.

### Verified locally

- `shardPages(4)` on the current sample: 9 / 9 / 8 / 8 runs, every page in exactly one shard, `shardPages(1)` returns all 18.
- `lighthouse-report.mjs` against two synthetic shards (CPU 2500 and 1800) with one page omitted and a CPU-2500 baseline: the missing page renders as `no shard reported this page`, the header says `deltas hidden on 8 page(s)`, and only the nine comparable pages carry ▲/▼.
- `node --check` and oxlint clean on the three scripts.

Since `lighthouse.yml` only triggers on PRs into `main`, this would surface on the first run after #5572 merges. Worth fixing here so that run is a real test of the sharding rather than a checkout failure.

elevatebart:
Right on both counts, fixed in 6efce51.

I reproduced it before changing anything, driving git the way `actions/checkout` does (set `core.sparseCheckout`, append the patterns to `info/sparse-checkout`, then check out) against a tree with the same paths:

| patterns | `.nvmrc` | `src/scripts` leaked |
|---|---|---|
| as merged | missing | yes |
| anchored + `/.nvmrc` | present | no |

Your list, verified by parsing it back out of the YAML rather than retyping it: `.github/actions/preview-url/action.yml`, `.nvmrc`, both report scripts and the page sample, nothing else.

Nothing else in the job wants a root file. `package.json` is not one of them: the scripts are `.mjs`, so they parse as ESM without `"type": "module"`.

---

iitzIrFan (approved):
**Approve.** `6efce51` is exactly the list I verified against `origin/main`: `.nvmrc` and `.github/actions/preview-url` present, scripts and page sample present, no `src/scripts` leak. Parsed back out of the YAML, the four anchored patterns are the only ones in the step.

The rest stands from the earlier pass: 9/9/8/8 shard split with every page assigned once, the merge marking unreported pages instead of dropping them, and the per-page CPU gate hiding deltas only on the slow shard's pages.

Sequence still matters: merge #5572 first, then this base flips to `main` and the workflow gets its first real run here. That run is the one to watch for the `dist` tarball size and the download cost per shard, which nothing local can confirm.

iitzIrFan:
Run 3759's hang has a cause one step before the one in `22de608`: wrangler's bundle **failed**, it wasn't compiling. Shard 3's start step captured it at 08:55:00, two seconds after launch:

```
Build failed with 4 errors:
  Could not resolve "virtual:astro-cloudflare:config"   (@astrojs/cloudflare/dist/utils/cf.js)
  Could not resolve "virtual:astro-cloudflare:config"   (@astrojs/cloudflare/dist/utils/handler.js)
  Could not resolve "astro:static-paths"
  ...
```

Those are the adapter's virtual modules, so wrangler was bundling `src/middlewares/worker.ts` from the repo's `wrangler.jsonc`, not the built worker. The build writes `dist/server/wrangler.json` (`main: entry.mjs`, `assets.directory: ../client`) and `@cloudflare/vite-plugin` writes the redirect that makes wrangler read it: `.wrangler/deploy/config.json` → `{"configPath":"../../dist/server/wrangler.json"}`. That file is outside `dist`, gitignored, and not in the tarball. The single-job workflow never hit this because build and `wrangler dev` shared a runner.

One-line fix in the build job:

```yaml
- name: Pack the build
  run: tar -I 'zstd -1 -T0' -cf dist.tar.zst dist .wrangler/deploy/config.json
```

The redirect is relative, so unpacking at the repo root (as the shards do) lands it correctly.

Keep everything in `22de608`: the bounded curl and the captured wrangler log are what will make the current run fail in 2 minutes with these errors printed instead of hanging. Two other numbers from 3759 worth having in the PR: the `lighthouse-dist` artifact is **1.26 GB** (765 MB of unoptimised images under `_astro` because `NO_IMAGE_OPTIM=true`), and download times per shard were 14 s, 59 s, 80 s, 127 s. Unpack is 2–3 s.

elevatebart:
That's the cause, thank you. Fixed in ea3bbc3, packing `.wrangler/deploy/config.json` alongside `dist`.

Confirmed the mechanism here before applying it: `@cloudflare/vite-plugin` is at 1.42.1 in the lockfile, pulled in by `@astrojs/cloudflare` (`^1.39.0`), and `.gitignore:35` is `.wrangler`, so a fresh checkout has neither the redirect nor the generated config. Also verified the round trip, since the redirect is relative and the shards unpack at the repo root:

```
unpacked in a fresh workspace:
  .wrangler/deploy/config.json
  dist/server/wrangler.json
configPath  : ../../dist/server/wrangler.json
resolves to : dist/server/wrangler.json   exists: True
```

The build now fails loudly if the redirect is absent rather than handing the shards a tarball that cannot be served.

Your artifact figure is right and mine was wrong: the API reports `lighthouse-dist` at 1,323,510,595 bytes. I had read the 2s pack and 14s upload as evidence `dist` was small, which it is not — 1.23 GiB uploads in 14s at 90 MB/s.

That reopens the build-once question, and I'd rather settle it on this run's numbers than argue it now. At 1.23 GiB the download costs a shard up to 2m07 on top of a 5m33 serialised build job, against roughly 4 min to build in each shard with no download at all. If the download stays that variable, building per shard is probably the faster shape, at 4x the build minutes. Worth a follow-up either way; this PR should first prove the sharding works.

---

iitzIrFan:
## First green sharded run: 12m04 wall clock

[Run 34459905263](https://github.com/kestra-io/docs/actions/runs/34459905263) on `ea3bbc3`, 09:18:27 → 09:30:31.

| stage | time |
|---|---|
| build job | 5.4 min (Astro build 3.8 min, pack 2 s, upload 14 s) |
| shards, parallel (longest) | 5.5 min |
| report job | 22 s |
| **total** | **12.1 min** |
| runner minutes | ~26 |

Against the single-job design's 16 min for the same 34 page runs this morning ([run 34454964403](https://github.com/kestra-io/docs/actions/runs/34454964403)): ~4 min saved on wall clock at ~1.6× the runner minutes.

Per shard:

| shard | download | dev-server wait | benchmark | runs | fixture check |
|---|---|---|---|---|---|
| 0 | 36 s | 5 s | 192 s | 9 | n/a |
| 1 | 15 s | 5 s | 171 s | 9 | 11 calls, records 0 |
| 2 | 67 s | 5 s | 136 s | 8 | n/a |
| 3 | 35 s | 5 s | 165 s | 8 | 31 calls, records 0 |

The wait step that hung 14 min on 3759 returned 200 in 5 s on all four, so `ea3bbc3` was the fix. Report merged 4 shards, 18 pages, none missing. Blueprints 84 (84/84/85) on CPU 2383.

### The posted report's baseline is stale, and it isn't this PR

The header says `Compared against main baseline from 2026-09-04 … (baseline: unknown)`, so the 🟢 +31 on Home is against a six-day-old score with no CPU gate. The download step (`dawidd6/action-download-artifact@v6`, identical to `main`) walks the workflow's runs newest-first and takes the first `success`. At 09:30:25 it logged `(found) Run ID: 33873994679, 2026-09-04` — GitHub's `workflows/lighthouse.yml/runs?branch=main` endpoint was returning the 09-04 run as the newest at that moment, skipping everything since. The same query returns today's 09:15 run now, so it was a transient indexing lag on GitHub's side; the same step on #5572's runs at 07:52 and 08:30 correctly found the 06:58 baseline. None of the action's options (`check_artifacts`, `search_artifacts`) would have helped, since they iterate the same run list.

Net effect: this run's deltas are noise. The next run compares against today's `main` baseline (which carries per-page CPU indices since `e938ece`) and the per-page gate becomes live. Nothing to change here; noting it so nobody reads the +31 as real.

### Still open, as a follow-up

`lighthouse-dist` is 1.23 GiB. Download cost this run was 15–67 s per shard, last run 14–127 s. Bart's build-once-vs-build-per-shard question stands; the numbers above are the input for it.

PR #5602, opened 2026-09-11
feat(customers): redesign the customer stories page and story cards

### LINKED ISSUE #5569: feat(customers): redesign customers listing & story detail pages
Implement the updated design for the `/customers` landing page and individual story detail pages per the new Figma spec.

**Figma:** [WEBSITE – Customers page redesign](https://www.figma.com/design/7YIFRlwzARHPMOAZTkRF7w/WEBSITE?node-id=28851-24992&m=dev)

**Interactive card component**: https://www.figma.com/make/Xdf0pQaDat4z2dwpRoM9ih/Interactive-Card-Component?t=h3Nt0MpqxuFCmAKE-1

---

### 1. Customers index page (`/customers`)

**Hero section**
- New headline copy: **"Thousands of teams. Millions of flows. Zero drama."**
  (currently: *"Trusted worldwide as the default workflow orchestration platform."*)
- Updated subheadline / supporting text to match new copy
- Possible logo-strip row above or replacing part of the featured carousel

**Featured carousel** (`FeaturedCarousel.vue`)
- Review card layout against new Figma card component (logo icon placement, overlay, KPI strip)
- Confirm autoplay / nav dots / arrow behaviour unchanged

**Story cards grid** (`StoriesList.vue`, `Card.vue`)
- Updated card component design: company logo icon top-left, company name, category badge, hero thumbnail
- Review spacing, typography, and hover states against Figma component panel

### 2. Story detail page (`/customers/[slug]`)

- Cleaner article layout: company logo + title/headline more prominent at the top
- **4 KPI stats** displayed as large numbers directly below the title (currently nested further down)
- Content sections: *The problem → What Kestra Needed → Results → Kestra at [Company]*
- **New "Similar Kestra Stories" section** at the bottom of each detail page — shows 3–4 related story cards (filter by industry or use-case, fall back to featured order)
- "See what you can build with Kestra" CTA above the footer (already present, confirm copy/placement)

### 3. New/updated components

Per Figma's **COMPONENTS** panel:
- Featured card (large) — updated layout
- Standard story card — new compact design
- Logo icon variants (icon-only, light/dark)

---

## Affected files

| File | Change |
|------|--------|
| `src/pages/customers/index.astro` | Hero copy, layout wiring |
| `src/pages/customers/[slug].astro` | Detail page layout, "Similar Stories" section |
| `src/components/stories/StoryGallery.astro` | Hero headline, sub-copy |
| `src/components/stories/FeaturedCarousel.vue` | Card layout tweaks |
| `src/components/stories/Card.vue` | New card design |
| `src/components/stories/StoriesList.vue` | Grid / filter bar adjustments |
| `src/components/stories/StoryDetail.vue` | KPI placement, new "Similar Stories" slot |

---

## Out of scope

- Content / copy changes to individual customer stories (`.yaml` frontmatter)
- Filter logic changes (industry / use-case / task dropdowns)
- Any changes to the `/blog` or other non-customer pages

---

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5569

---


Techboidesign (approved):

MartinRst:
LGTM

flcarre (changes requested):
Design-system compliance review of the customers redesign. Reviewed all 32 files at `fd6cb76`.

Nice work on the structural cleanup — extracting `src/utils/stories.ts`, killing the duplicated `toStory` mappers in both Astro pages, the `relatedEntries` scoring instead of `randomSortFunction`, and the `StoryDetail` `related: any` → real type are all clear wins. The `TaskIcon` `theme` prop is the right way to solve the dark-card icon problem.

Four things I'd want fixed before this merges. Two are functional, not stylistic.

---

## Blocking

### 1. Keyboard focus permanently desyncs the carousel rail

`src/components/stories/FeaturedCarousel.vue`

The rail is positioned *purely* by `transform: translate3d()` — `setOffset()` never touches scroll. But `.fc-track { overflow: hidden }` is still a programmatically scrollable box: when a copy-1 card that is laid out off-screen receives keyboard focus, the browser sets `track.scrollLeft` to reveal it. Nothing in the component ever reads or resets `scrollLeft`, so from the first `Tab` into the carousel onward the container scroll and the transform offset are desynced, and the drift animates from a shifted origin. It never recovers.

Fix: `overflow: clip` on `.fc-track` (clips without creating a scroll container), or handle `scroll`/`focusin` by resetting `scrollLeft = 0` and folding the delta into `offset`.

### 2. The carousel lost every accessible control

The previous version had prev/next buttons (`aria-label="Previous story"` / `"Next story"`) plus per-slide dot buttons (`aria-label="Go to story N"`). All removed. The replacement is drag-only + auto-drift with:

- no `keydown` handler
- no `role` / `aria-label` on the region
- no focus style on `.fc-card` (and the default outline gets clipped by the `overflow: hidden` in #1)

And for `prefers-reduced-motion` users, `startDrift()` returns early, so the rail is frozen at its initial offset with **no** way to advance except a pointer drag. This is a public marketing page — it needs at least prev/next controls back, or arrow-key handling on a labelled region.

### 3. Flip cards hide half their content on touch

`src/components/stories/Card.vue:212-214`

`:hover` and `:focus-visible` are the only triggers for `rotateY(180deg)`, and there's no `@media (hover: none)` fallback. For the five `HIGHLIGHTED_IDS` cards, the stats face — KPIs and "Plugins Used" — is unreachable on touch devices; a tap just navigates. Worth deciding what the touch presentation is: auto-flip on intersection, a tap-to-flip affordance, or simply render the stats face for `(hover: none)`.

### 4. ~40 raw hex/rgba colors, several retyped from the palette

`Card.vue`, `FeaturedCarousel.vue`, `StoryGallery.astro`. These are exact palette values written as hex:

| Added | Line(s) | Already exists as |
|---|---|---|
| `#0e0e10` / `rgba(14,14,16,…)` | `FeaturedCarousel.vue:286,323-325`, `Card.vue:439` | `$base-grey-800` |
| `#a796ff` | `Card.vue:500`, `FeaturedCarousel.vue:401` | `$base-primary-300` |
| `#7f7f86` | `FeaturedCarousel.vue:437` | `$base-grey-400` |
| `rgba(99, 27, 255, 0.12)` | `Card.vue:190` | `rgba($base-primary-500, 0.12)` |
| `#fff` | 8 sites | `$base-grey-white` |

And these are off-palette entirely — **five different near-blacks for one visual role**:

- `#080a0c` — `Card.vue:272`
- `#08090a` — `Card.vue:415`, `FeaturedCarousel.vue:370`
- `#0a0b0d` — `StoryGallery.astro:39`
- `#111115` — `Card.vue:411`, `FeaturedCarousel.vue:427`

Plus off-palette `#adadbf` (`Card.vue:507`, `FeaturedCarousel.vue:408`), `#2e2e3c` (`FeaturedCarousel.vue:426`), and `rgba(57, 82, 242, 0.1)` (`Card.vue:253`) — a *blue* shadow on a purple brand.

One caveat on the fix, so this doesn't go the wrong way: these are intentionally always-dark surfaces, so **don't** reach for `--ks-*` or `--ks-*-inverse` — both flip with the theme (`--ks-content-primary-inverse` is `#ffffff` in light and `#131316` in dark, so an inverse token would give you dark text on a dark card). The right vehicle here is the theme-independent `$base-*` palette vars from `_color-palette.scss`, which `astro.config.mjs` already injects globally via `additionalData` and which the repo already uses this way.

---

## Non-blocking

**No visual regression coverage for the redesigned pages.** `tests/visual-regression.spec.ts` snapshots Home, Pricing, Enterprise, Blog, Get-Started and others at desktop/tablet/mobile. There's no `/customers` case and none added, so a ~1400-line redesign of two pages ships uncovered.

**Any resize snaps the rail back to the start.** `ro = new ResizeObserver(measure)` and `measure()` ends with `setOffset(loopWidth - bleed)`, so an orientation change discards the reader's position.

**`.filter-clear` has no focus style** (`StoriesList.vue`) — new button with `background: none; border: none` and a `:hover` rule only.

**`GRID_ORDER` hardcodes all 29 story ids** (`src/utils/stories.ts`). A newly added customer story silently lands at the end of the grid until someone edits the array. The `rank` → id fallback means it degrades rather than breaks, but this deserves a marker so the next person knows to update it.

**Nits**

- ~15 narrative comments added across `FeaturedCarousel.vue`, `Card.vue`, `Sidebar.vue`, `StoryGallery.astro`, `stories.ts`.
- `--fc-bleed: calc((100vw - 100cqw) / 2)` — `100vw` includes the scrollbar, so the full-bleed strip is ~7px asymmetric on desktop. Harmless, `.stories-hero { overflow: hidden }` clips it.
- 18+ `loading="eager"` hero images (3 rail copies × 6 featured). Same URLs so only 6 fetches, but 18 decodes on a landing page — worth `lazy` for `copy > 1`.
- `More.vue` changed `.arrow-icon` → `:deep(.arrow-icon)`, reaching into `Link.vue`'s internals. It does fix a previously dead rule; a prop or CSS var on `Link` would be cleaner.

---

## Verified clean — no need to chase these

- `SeeHow.astro` **does** accept `cta` and a `footer` slot, and defaults `cta` off when `footer` is present. The new CTA wiring is correct.
- `headerTheme="lightText"` is a valid `layout.astro` value.
- `$font-family-monospace` resolves (nine other components use it).
- All four newly-used tokens — `--ks-background-purple-light`, `--ks-border-active`, `--ks-background-input`, `--ks-background-primary` — exist in **both** light and dark themes.
- Every id in `FEATURED_ORDER` (6), `GRID_ORDER` (29) and `HIGHLIGHTED_IDS` (5) resolves to a real story directory, and `GRID_ORDER` covers all 29. All highlighted and featured stories have `quote`/`quotePerson`. Riverside has no `logo:` but does have `logoIcon`, which the `logo ?? logoIcon` fallback picks up.
- No dangling references to the deleted `KpiRotator.vue`, `StoryRow.vue`, `public/stories/grid.png`, `public/stories/gradient.png`, or the removed `totalStories` prop.
- The `var(--top-bar-height, 67px)` negative-margin hero pattern matches `home/Hero.astro`, `get-started/Hero.astro` and `twozero/Hero.astro` exactly.

iitzIrFan:
@flcarre thanks for the thorough review. Everything is addressed across `ad5c4a7`, `b0d54ac` and `07843fb`, all verified on the running dev server.

## Blocking

1. **Carousel rail desync** — `.fc-track` is now `overflow: clip`, so focusing an off-screen card can no longer scroll the track out of sync with the transform. Tabbing through the rail keeps `scrollLeft` at 0, and `onFocusIn` pulls a focused card back to the content edge via the transform instead.
2. **Accessible controls** — the region is labelled (`aria-label="Featured customer stories"`, `aria-roledescription="carousel"`), ArrowLeft/ArrowRight step the rail, cards have a visible `:focus-visible` ring, and only the exposed copy of the rail is in the tab order / a11y tree. For the visible control I went back to the Figma (WEBSITE `28923:44262`), which has no prev/next arrows but a centred stepper. That is what shipped in `07843fb`: one real `<button>` per story (`Go to story N: <company>`, `aria-current` on the active one), a 64px primary-gradient pill for the active story and 10px dots, 32px under the cards. Clicking a dot glides the rail the short way round the loop; the active dot tracks the card at the content edge during drift, drag and keyboard stepping. Reduced-motion users get an instant jump, so the rail is never stuck on a pointer drag.
3. **Flip cards on touch** — `@media (hover: none)` renders the stats face outright for the five highlighted cards, so KPIs and Plugins Used are reachable on touch. Checked under mobile emulation.
4. **Raw colours** — Card, FeaturedCarousel and StoryGallery now use the `$base-*` palette vars from `_color-palette.scss` (no `--ks-*` on the always-dark surfaces, as you suggested). The near-black variants collapsed onto `$base-grey-800` / `$base-grey-Black`, the blue shadow is now `rgba($base-primary-500, …)`, and the one remaining literal is the hero's second glow, which is documented as a deliberate off-palette accent. The stray `#fff`/`#000` on the story Sidebar logo box moved to the palette too. The stepper's dot colour in Figma (`#444459`) is off-palette as well, so it is built as `rgba($base-primary-200, 0.28)` with a note.

## Non-blocking

- **Visual regression** — `/customers` is in `PAGES` with a new `reducedMotion` flag, and the spec emulates `prefers-reduced-motion: reduce` for it so the auto-drifting rail produces a stable baseline. `PAGES`/`VISUAL_ONLY_PAGES` now share a `SamplePage` JSDoc typedef, which also fixes the `astro check` ts(2339) that the first version of that change introduced. Note the Linux baselines for the customers pages do not exist yet on `main` either; they need a run of the `update-linux-snapshots` workflow on this branch.
- **Resize snapping** — `measure()` now preserves the reader's progress through the loop instead of resetting to the start.
- **`.filter-clear` focus style** — added.
- **`GRID_ORDER` marker** — comment added explaining the `rank` → id fallback for stories not yet in the array.
- **Nits** — narrative comments trimmed, hidden rail copies load their images lazily (only the exposed copy is `eager`), trailing whitespace removed. I left `More.vue`'s `:deep(.arrow-icon)` as is for now; happy to move it to a prop on `Link` in a follow-up if you'd prefer.

PR #5603, opened 2026-09-11
ci(deploy): fail the build when HTML links a /_astro/ asset that dist lacks

elevatebart:
## Summary

Yesterday's outage was a page restored from a stale incremental-build cache linking a `layout.*.css` hash the new build no longer emitted. Cloudflare cache purges could not fix it, and it took ~50 minutes to find the Actions cache. This adds a guard so that shape of bug fails the build instead of shipping.

## Changes

- `scripts/check-dist-asset-links.mjs`: scans every `dist/**/*.html` (falling back to `dist/client`, where the Cloudflare adapter writes static output) for `/_astro/` references and fails with a list of missing assets and the pages linking them. Exports `findDanglingAssetLinks` for testing.
- `deploy.yml`: runs the check after `Build` and before either wrangler step, on preview and production alike.
- `src/utils/checkDistAssetLinks.test.ts`: three unit tests against a temp `dist`.

## First catch

The check's first real run on this PR found a pre-existing broken avatar on `/blogs/2026-04-28-how-we-release-kestra-with-kestra`: `racevedo-round.png` and `racevedo-sm.png` were byte-identical, so Vite emitted one `_astro` file for both. `Team.astro` runs the round one through `<Image>`, and after generating the transform Astro deletes the original as "not referenced outside of image processing". That original was the file the blog post's plain `` pointed at. Fixed here by re-encoding `racevedo-sm.png` (same pixels, different bytes) so the two imports get separate assets. Production was not exposed: the cloudflare image service transforms at the edge and never runs the deletion pass, so this only bit preview builds.

## Not in this PR

The `components/layouts/**` glob I meant to remove is already gone: the version of #5599 that merged replaced both globs with a single `~/**/*.scss` glob.

The remaining hole in Astro's dependency hash (Sass partials, `additionalData`, CSS chunk regrouping) is filed upstream as withastro/astro#17974 with a repro. This check is the safety net until that lands.

## Verification

- `npx vitest run src/utils/checkDistAssetLinks.test.ts`: 3/3
- `npx astro check`: 0 errors
- `oxlint` and `oxfmt --check` on the new files: clean
- Manual runs against fake `dist` and `dist/client` layouts: exit 1 with dangling refs listed, exit 0 once they resolve, exit 1 when no `_astro` directory is found.
- Avatar re-encode: decoded pixels identical before and after (sharp), md5 now differs from the round file.

https://claude.ai/code/session_01Fb2ZqWKJGHKhTvn7BDYrnR


brian-mulier-p (approved):
LGTM, CI is failing though

PR #5608, opened 2026-09-11
fix(docs): render archived-version components from the codebase, not mirrors

brian-mulier-p:
Versioned doc pages (`/docs/{major.minor}/...`) rendered MDC components from hand-written HTML copies kept in `renderVersionedDoc.ts`. Every new component needed a bespoke branch there, and anything without one was silently dropped.

`/docs/1.3/api-reference/enterprise` rendered nothing because of this: `<ApiDocEE />` is the last, childless node on that page, so the trailing-residue trim removed it before it could render. `next-link`, `api-doc`, `whats-new` and `child-table-of-contents` were dropped the same way.

Now:
- directives dispatch through the live remark `componentMap`, so a directive registered there works on archived versions with no extra wiring;
- real components are handed to `docs-versioned.astro` and resolved by file name over `src/components/{content,docs,common}/*.{astro,vue}`, lazily — adding a component there needs no renderer change;
- overrides remain only for components that read the current content collection, which an archived page must not use;
- a tag that resolves to nothing renders its children and is logged, instead of disappearing.

Archived API references also serve their own spec: RapiDoc's spec is a static asset of the live site, so `/docs/1.3/api-reference/*` documented today's API. The docs indexing job already uploads the repo's `public/` per version, so `/api/openapi/{version}/{oss,ee}.yml` proxies that version's own spec from the docs API. Versions with no published spec return an error rather than falling back to the current spec.

Components that read the current content collection (child cards, table of contents, `WhatsNew`) stay excluded from that glob: rendering them on an archived page would show today's docs tree, and importing `getCollection` pulled Astro's whole content data layer into this worker, pushing it past Cloudflare's 64 MB limit. A test pins the exclusion list to the files that actually import `astro:content`, so a new one is caught here instead of failing deploy.

Reported in Slack by Hans-Peter GRAHSL.


brian-mulier-p:
@copilot resolve the merge conflicts in this pull request

elevatebart (approved):
Looks good, nothing to add

src/utils/renderVersionedDoc.ts
```diff
@@ -71,45 +79,91 @@ function parseButtons(raw: unknown): HomePageButton[] {
...
+// Exists in the codebase, deliberately not rendered here: it lists the latest
+// blog posts off the current content collection, which an archived page has no
+// business showing — and which would drag the whole content data layer into the
+// docs worker. Known, so it stays out of the drift signal.
+const CURRENT_CONTENT_ONLY = new Set(["whats-new"])
+
+/** A component tag resolves to something renderable, one way or another. */
+function isKnownComponent(tag: string, ctx: RenderCtx): boolean {
+    return (
+        OVERRIDDEN_COMPONENTS.has(tag) ||
```
elevatebart, line 122: `CURRENT_CONTENT_ONLY` is only consulted in `componentHtml`'s default branch, not here. A page whose last node is `<WhatsNew />` is trimmed as trailing residue, never reaches `componentHtml`, and then gets flagged by the `referencedTags` loop. Ran it on this branch:

```
markdown: "## News\n\n<WhatsNew />"   -> unknownComponents: ["whats-new"]
markdown: same + a paragraph after it -> unknownComponents: []
```

Same shape as the `guides-child-card` hole the old `continue` papered over. Rendering is unaffected (it renders nothing either way), it is only a false drift warning in the worker logs, and `whats-new` does appear in the archived corpus.

```suggestion
        OVERRIDDEN_COMPONENTS.has(tag) ||
        CURRENT_CONTENT_ONLY.has(tag) ||
```

---

src/utils/renderVersionedDoc.ts
```diff
@@ -169,11 +199,89 @@ function componentHtml(
...
+function targetDocKey(props: Record<string, unknown>, ctx: RenderCtx): string {
+    // The MDC parser kebab-cases camelCase attribute names, so a 0.19-era
+    // `{pageUrl="..."}` directive lands as `page-url`, not `pageUrl` — check
+    // both so the intended directory is honored.
+    const target = props.pageUrl ?? props["page-url"] ?? props.directory
+    if (typeof target !== "string") return ctx.pageKey
+    return `docs/${target.replace(/^\/?docs\/?/, "").replace(/\/+$/, "")}`.replace(/\/$/, "")
+}
+
+/** Nested link list of a node's children, mirroring ChildTableOfContents.astro. */
```
elevatebart: The list this emits is flat, one `<ul>` of `directDocChildren`, while `ChildTableOfContents.astro` emits a second level (`<ul data-level="1">`) gated by the `max` prop, which is dropped here. So an archived ToC silently loses its grandchildren and `data-level="1"` styling. Either recurse one level and honor `max`, or reword the comment so the next reader doesn't take "nested" at face value.

---

src/pages/docs-versioned.astro
```diff
@@ -8,13 +8,60 @@ import ArchivedBanner from "~/components/layout/ArchivedBanner.astro"
...
+// globbing them pulls Astro's content data layer (50 MB) into this worker and
+// blows Cloudflare's 64 MB limit. They must not render on an archived page
+// anyway — they'd list the current docs tree — so the renderer overrides them
+// or skips them outright.
+// The negative patterns are repeated literally in both calls on purpose: Vite
+// only reads statically analysable patterns here, and silently keeps the
+// positive one when handed anything else. The glob test pins them to the files
+// that actually import astro:content.
+const vueLoaders = import.meta.glob<{ default: AstroComponent }>([
+    "/src/components/{content,docs,common}/*.vue",
```
elevatebart: Three directories is wider than the corpus needs, and the width is what forces the exclusion list. Every tag the tests exercise (`api-doc`, `api-doc-ee`, `home-page-buttons`, `support-links`, `card-logos`, `download-logo-pack`, `child-table-of-contents`) lives in `content/`; of the four exclusions, three are only there because `docs/` and `common/` got pulled in.

The side effect is that any file name under those directories becomes a component tag. `link` is not in `HTML_TAGS` (it is in `VOID_TAGS`, so the intent was to treat it as HTML), which means raw `<link>` in an archived page now resolves to `common/Link.astro`:

```
"<link rel=\"stylesheet\" href=\"/x.css\">"
-> components=[{"tag":"link","props":{"rel":"stylesheet","href":"/x.css"}}]
```

That renders the site's arrow-link `<a>` pointing at a stylesheet. Rare input, but the same collision is open for `Tabs`, `Modal`, `Copy`, `Share`, `Button`. Narrowing to `/src/components/content/*` plus an explicit list of anything else archived pages need would drop three exclusions, shrink this route's dependency graph (no more rapidoc riding along), and close the collision class. Also worth adding `link` to `HTML_TAGS` either way.

Second-order: component instantiation happens in the template, outside the `try/catch` around `renderVersionedDocBody`, so a component that throws in SSR (missing required prop, `document` at setup) takes the whole archived page to a 500 rather than degrading to children. A narrower glob bounds that too.

---

src/pages/api/openapi/[version]/[edition].yml.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (edition !== "oss" && edition !== "ee") {
+        return new Response("Unknown spec edition", { status: 404 })
+    }
+
+    const source = versionedSpecSourceUrl(API_URL, params.version ?? "", edition)
+    if (!source) {
+        return new Response("Unknown docs version", { status: 404 })
+    }
+
+    const upstream = await fetch(source)
```
Comment (elevatebart) [inline src/pages/api/openapi/[version]/[edition].yml.ts]: Two things on the proxy:

1. Unguarded `fetch`: a rejected promise (DNS, connect reset, worker subrequest limit) throws out of the handler and the reader gets a 500 HTML error page inside RapiDoc instead of the 502 the code below is careful to produce. A `try/catch` returning 502 keeps the failure shape consistent.
2. `Cache-Control` only tells the browser. The EE spec is 1.2 MB and an archived spec is immutable by definition, so every cold load re-proxies the full body through the worker. `caches.default` (or `fetch(source, { cf: { cacheEverything: true, cacheTtl: 86400 } })`) would make the second reader cheap, and `immutable` on the header is free.

I could not verify the upstream assumption from here (egress to `api.kestra.io` is blocked in my sandbox), so the one thing worth confirming manually: `{api}/docs/kestra-ee.yml/versions/1.3.0` actually returns the 1.3 spec rather than a 404, for the oldest version you intend to keep serving. If the indexing job only started uploading `public/` recently, older versions get the error path, which is the intended behavior but a visible regression from "wrong spec" to "no spec".

---

src/utils/renderVersionedDoc.test.ts
```diff
@@ -995,3 +1030,281 @@ title: T
...
+            expect(componentFileKeys).toContain(componentKey(tag))
+        },
+    )
+})
+
+describe("versioned-docs component glob", () => {
+    const srcDir = resolve(dirname(fileURLToPath(import.meta.url)), "..")
+    const page = readFileSync(resolve(srcDir, "pages/docs-versioned.astro"), "utf8")
+    const excluded = new Set(
+        [...page.matchAll(/"!(\/src\/components\/[^"]+)"/g)].map((match) => match[1]),
```
elevatebart: Nit: this collects the negatives from the whole file into one set, so it passes when the exclusions are dropped from one of the two globs, which is exactly the "repeated literally in both calls on purpose" invariant the page comment relies on. Matching per glob call (or asserting the two lists are equal) would pin it.

---

src/utils/renderVersionedDoc.ts
```diff
@@ -169,11 +199,89 @@ function componentHtml(
...
+ * registered in `componentMap` needs no case here to work on archived versions.
+ */
+function directiveHtml(tag: string, node: MdcNode, ctx: RenderCtx): string {
+    ctx.seenComponents.add(tag)
+    const data: LoweredNode["data"] = {}
+    // The handlers mutate `children`, so hand them a copy: `node` is the real
+    // tree, and a re-render of the same memoized parse must start clean.
+    const lowered: LoweredNode = { type: "element", children: [...(node.children ?? [])] }
+    try {
+        directiveHandlers[tag](data, node.props ?? {}, lowered)
```
elevatebart, line 289: Nit, not blocking: handlers that replace `node.children` wholesale drop the body, which breaks the "`inner` is always appended" rule the file states a few lines up. `badge` is the one that does it:

```
"::badge{version=\"0.14\" editions=\"EE\"}\nbody sentence\n::"
-> the badge pills, "body sentence" gone (main kept it via badgeHtml(props) + inner)
```

Harmless against today's corpus (every `:::badge` in the docs is immediately closed with an empty body, and an unclosed one parses as a leaf), so this is a note rather than a request. If you want belt and braces, serialize the pre-handler children when a handler replaced them and the originals were non-empty.

---

elevatebart (commented):
Reviewed the whole diff and ran the suite on the branch (26 files, 395 tests, all green) plus a few probes of my own against the renderer.

The direction is right: dispatching through the live `componentMap` and resolving the rest by file name removes the per-component switch that was the actual bug, and the trailing-residue fix (a component that still exists is never residue) is the correct root cause for the blank 1.3 API reference. Serving each version its own spec closes a second, quieter wrongness that nobody had filed yet.

What I found, none of it blocking:

- **Confirmed, one line**: `isKnownComponent` skips `CURRENT_CONTENT_ONLY`, so a trailing `<WhatsNew />` still lands in `unknownComponents`. Logs only, but it is the same false-drift hole the old `guides-child-card` `continue` existed for.
- **Fidelity**: `tableOfContentsHtml` is flat and ignores `max`, while the real component nests one level. The comment says nested.
- **Scope**: the glob spanning `{content,docs,common}` is what creates the exclusion list, and it makes every file name in those directories a live component tag (raw `<link>` now resolves to `common/Link.astro`, verified). Narrowing to `content/` plus an explicit list would drop three exclusions and the collision class.
- **Proxy**: unguarded `fetch`, and an immutable 1.2 MB body re-proxied on every cold load.
- Two nits inline (glob-pin test, `badge` dropping a non-empty body).

Nothing here needs to hold the merge. The `whats-new` line and the ToC comment are cheap enough to fold in; the glob scope is worth a conversation, not a rewrite of this PR.

---

PR #5609, opened 2026-09-11
feat(changelog): redesign changelog page as a timeline

### LINKED ISSUE #5596: feat(changelog): redesign changelog page as a timeline
Redesign the changelog page to follow a timeline-based layout, replacing the current grid of empty boxes that each link to secondary pages.

## Motivation

The current changelog page requires users to click into individual secondary pages to see content. A timeline layout surfaces content inline, making it easier to scan and consume updates at a glance.

## Design

Figma: [WEBSITE → node 30017-5278](https://www.figma.com/design/7YIFRlwzARHPMOAZTkRF7w/WEBSITE?node-id=30017-5278&t=vlSj8fluRgkcHv8K-11)

Key changes:
- Replace the box/card grid with a vertical timeline
- Inline changelog entries — no secondary page clicks required
- Inspired by [supabase.com/changelog](http://supabase.com/changelog) and [linear.app/changelog](http://linear.app/changelog)

## Related

- Design discussion: [kestra-io/design#23 (comment)](https://github.com/kestra-io/design/issues/23#issuecomment-5605653129)

iitzIrFan:
fixes: https://github.com/kestra-io/docs/issues/5596

---

URL: /docs/changelog 

---


Techboidesign (approved):

MilosPaunovic:
Are we okay with this much space between `scope` and the commit hash?
image


Also, I find it really hard to navigate like this where we have all version releases just sitting there, `2.0.1`, then `1.3.38` then `1.0.60`... As most of the people actually exploring the changelog would actually like to focus on a single verison, say `v2.0.x` and then scroll through only releases of that - it's more of a design note so pinging @Techboidesign. I get that we can type the version prefix in the text search and filter it out, but maybe consider some pill-like quick filters also?

MilosPaunovic (commented):
Reviewed the diff and the files, and ran the branch locally with `astro dev` off `1f50a930a`. `CI` is green (12/12), the 15 new unit tests pass, `oxlint` is clean on the changed paths, and the page does what https://github.com/kestra-io/docs/issues/5596 asks for. Six things I would want changed, then nits.

## 1. Enterprise Edition change links point at the wrong repository

`src/components/changelog/ChangelogChangeGroup.vue:42` and `src/utils/changelog/markdown.ts:3` hardcode `https://github.com/kestra-io/kestra/issues/${pr}`. Under the `Enterprise Edition` heading the `(#N)` id is a `kestra-io/kestra-ee` number, so every `EE` entry links a reader to an unrelated `OSS` issue. Verified on the running page, `v1.3.17`:

- rendered: `clear kill switch form when opening create dialog (#7697)` links to https://github.com/kestra-io/kestra/issues/7697
- that one is `chore(ui): properly sanitize markdown content before rendering`
- the real one is `kestra-io/kestra-ee` `#7697`, `fix(cluster): clear kill switch form when opening create dialog`

Same bug in the generated `/docs/changelog.md` (`#### Bug Fixes` under `### Enterprise Edition`). `src/utils/changelog/parseRelease.ts:208` already special-cases the `EE` edition to drop the `SHA`; the `PR` number needs the same treatment. Since `kestra-io/kestra-ee` is private, dropping the link and keeping the plain `(#N)` text is the consistent fix. Note `src/utils/changelog/parseRelease.test.ts:103` asserts `pr: 7697` on the `EE` change, so that expectation moves with the fix.

## 2. The 503 branch in `changelog.md.ts` can never fire

`src/pages/docs/changelog.md.ts` has no `export const prerender = false`, and `astro.config.mjs` sets no `output`, so the default `static` applies (28 other routes opt into `SSR` explicitly). Astro's static build writes the response body to `dist/docs/changelog.md` for any non-3xx status and throws away the status plus the `cache-control: no-store` and `retry-after` headers (`astro/dist/core/build/generate.js:275-280`).

So the failure mode is not a 503: `fetchMajorReleases` swallows every error and returns `[]`, and the build then ships a `200` whose entire content is `Release notes temporarily unavailable`, cached like any other static asset until the next deploy, and `Copy as Markdown` / `Open in ChatGPT` / `Open in Claude` hand that string to the user. Either add `export const prerender = false`, or make the empty case fail the build loudly.

## 3. Searching collapses every group, hiding the match

`src/components/changelog/ChangelogTimeline.vue:226` returns `[]` from `openGroupsFor` whenever a search or a Major/Minor filter is active. Searching `clear kill switch form` narrows 100 releases to one change and then shows only a collapsed `Bug Fixes  1 change` row, one more click away from the thing you searched for. The search already filters the changes inside each group, so opening the groups that survived the filter costs nothing and makes the result visible.

## 4. `PR` numbers are rendered but not searchable

`searchedEntries` matches `message`, `scope` and `sha`, not `change.pr`. Verified on the page: `a3ab8b5` gives 1 result, `surface the AI Copilot` gives 1 result, `19393` gives 0 results while `(#19393)` is on screen right next to it. One extra `matches(String(change.pr), query)` closes the gap.

## 5. Dates are formatted in the viewer's timezone inside a prerendered island

`src/components/changelog/ChangelogEntry.vue:65-72` (`toLocaleDateString` plus `getFullYear`) and `src/utils/changelog/markdown.ts:5` pass no `timeZone`. The page is prerendered (build server on `UTC`) and then hydrated by Vue with the viewer's local timezone, so the date flips and Vue logs a hydration mismatch. Measured against the real data: 35 of the 100 releases currently served have a `published_at` whose `UTC` date differs from the Tokyo date. `v1.3.37` at `2026-09-02T18:16:07Z` is `Sep 2` for the build and `Sep 3` in Tokyo. `timeZone: "UTC"` plus `getUTCFullYear()` fixes both spots. This one is inherited from `ChangelogList.vue`, but the date is now the timeline's anchor element, so it is worth doing here.

## 6. Unrelated JSDoc deletion in `markdown-actions/types.ts`

All eight doc comments on `MarkdownActionContext` and `MarkdownActionDefinition` were removed while adding a single `lazyMarkdown` field. That is shared code the changelog only borrows, so worth restoring and documenting `lazyMarkdown` alongside them.

## Nits

- `changelog.md.ts` has stray double blank lines after the imports and after `const releases`.
- The dropdown trigger now reads `Copy as Markdown` and so does its first menu item (`MARKDOWN_ACTIONS[0].label`); the docs default `Copy Page` avoided that collision.
- Only the trailing `(#N)` is linked, so entries with two ids read oddly: `add dateFilter support to EE execution search (#7673) ([#7747](...))`, in both the `UI` and the markdown.
- `src/utils/changelog/parseRelease.ts:141`, `cleaned.split(/\s+-\s+\*/)[0]`, is the one line whose intent is not self-evident. It is covered by the `RFC 9457` test, so a one-line comment naming the italic-body case would keep it from being "simplified" later.
- `.entry-dot` uses `--ks-border-active` (purple) for `.major`, while `.badge-major` and the `Major` filter chip are green, the same concept in two colours.
- `changelog.md.ts` adds a fourth unmemoised `releases?per_page=150` call per build (page, `.md` route, `[tag]` pages, sitemap). `$fetchCached`'s `cf.cacheTtl` does nothing under `prerenderEnvironment: "node"`, and unauthenticated GitHub allows 60 requests/hour per `IP`.
- Not this `PR`, but relevant to its premise: GitHub caps `per_page` at 100, so `fetchMajorReleases(150)` only ever returns 100 releases, and the served `/docs/changelog.md` contains exactly 100.

## What I verified locally

Dark and light render, search, the Major/Minor filter, `?q=` and `?filter=` sync, group expand/collapse, `Load 20 more releases`, and `/docs/changelog.md` returning `200 text/markdown` (261 KB, 100 releases). The sticky scroll-spy dot I only reviewed in code, since the browser pane went hidden partway through and scrolling does not apply while it is.

iitzIrFan:
Thanks for the thorough pass, Milos, especially for tracing the #7697 mismatch to kestra-ee and for catching that the 503 could never actually fire on a prerendered route. Both would have shipped quietly.

Pushed b7afa2b9 with everything from the list:

- Enterprise ids stay as plain text now (same treatment as the SHAs), and the GitHub URL lives in one issueUrl() helper shared by the page and the markdown.
- changelog.md.ts is prerender = false, so the 503 and its headers are real. That also drops the build-time GitHub call you counted.
- A search that narrows a release opens the groups it kept; title matches and Major/Minor results stay collapsed, which was the design ask.
- PR numbers are searchable.
- Dates format in UTC on both sides, with a test for the late-evening case.
- The JSDoc in types.ts is back, with lazyMarkdown documented next to it.

Nits: blank lines, the Copy Page label (I removed the label prop altogether), the italic-body comment, and the major dot now matching the badge green are all in. I also tightened the scope/hash spacing from your screenshot: the change rows use a subgrid so columns size to content per group.

Two I left: the double-id linking would mean linkifying every #N in the message text on both renderers for a rare case, and per_page as you noted is its own change. The version quick-filter idea I'll leave with @Techboidesign; happy to add pills once there's a design call.

---

MilosPaunovic (commented):
`src/pages/docs/changelog.md.ts` is `prerender = false` now, so the route re-parses 100 release bodies and rebuilds ~261 KB per request, and the 200 carries only `Content-Type`. It is the only generated `SSR` endpoint here without a cache header: `src/pages/api/github-releases.ts` is the closest precedent (`prerender = false`, GitHub release data, `Cache-Control: public, max-age=86400`) and `src/pages/llms-full.txt.ts` uses `public, max-age=3600`. It matters for the failure path too, since `fetchMajorReleases` calls `api.github.com` unauthenticated at 60 requests/hour and `cf.cacheTtl` only caches the upstream fetch, so a rate-limited worker 503s per request. The `no-store` and `retry-after` on the 503 are right as they are.

iitzIrFan:
Pushed 8351464e with Cache-Control: public, max-age=3600 on the 200, matching llms-full.txt.ts; 503 unchanged. That should fix it.

MilosPaunovic (approved):
All three rounds are addressed, and the `Cache-Control` is verified on the 200 locally alongside the untouched 503. Approving.

One thing for later, not for this `PR`: `run_worker_first: true` plus the extension short-circuit in `src/middlewares/worker.ts` means `/docs/changelog.md` never reaches the Cache API, so `max-age` helps browsers and the AI fetchers but every cold request still re-parses 100 release bodies in the Worker. The short-circuit's comment assumes `.md` routes are "a cheap fetch + string reshape" - true everywhere except this one. Worth its own issue against `worker.ts` if the `TTFB` ever shows up.

PR #5616, opened 2026-09-14
test: verify astro incremental-build CSS fix on a preview (do not merge)

elevatebart:
Throwaway branch to exercise `experimental.incrementalBuild` against the fix candidate for [withastro/astro#17974](https://github.com/withastro/astro/issues/17974), without putting it anywhere near production. **Not for merge.**

## What is pinned

`astro` points at `https://pkg.pr.new/astro@cb815d8`, the continuous release of `withastro/astro@factory/fix-17974`. That commit captures compiled CSS in a `transform` hook and hashes it instead of reading the raw entry file from disk, so Sass partials pulled in via `@use` and `additionalData` reach the dependency hash. Its base is main after 7.3.2, so it is a forward pin, not a rollback to the older `incremental-graph-hashing` snapshot.

The URL is inlined in both `overrides` entries instead of `$astro`. A failed expansion there would nest a second astro under `astro-expressive-code`, which is a poor way to learn something about cache invalidation.

`package-lock.json` is untouched on purpose: CI runs `npm install`, which re-resolves it in the runner.

## How to read the result

1. **This build is cold.** A new branch has an empty `astro-build-preview-*` cache and the astro version changed, so every page renders. It only proves the tarball installs and builds.
2. **The next push changes one Sass partial and nothing else.** That is the actual test.

On that second run:

- Build log: pages that pull `_variable.scss` should be absent from the `(restored)` / `(cached)` lines.
- The existing "Check for dangling /_astro/ links" step is pass/fail.
- Preview alias URL for the eyeball check.

## Known gap

The pinned commit covers the Sass partial and `additionalData` causes from the issue. It does not address CSS chunk regrouping, where a shared chunk is renamed across builds with no source change. [withastro/astro#17976](https://github.com/withastro/astro/pull/17976) covers that one by folding each page's resolved stylesheets into the dependency hash. A partial pass here is the expected outcome.

pkg.pr.new prunes old builds, so this pin has a shelf life. Close the branch once the upstream PR lands.

https://claude.ai/code/session_01GmBCjUUoZRS4jbVk4gXjUk

---
_Generated by [Claude Code](https://claude.ai/code/session_01GmBCjUUoZRS4jbVk4gXjUk)_


elevatebart:
`Lighthouse shard 1` failed on `b95f0c4` at the **Install Lighthouse** step (`npm install --no-save lighthouse chrome-launcher`), before any page was measured. Shards 0, 2 and 3 on the same commit installed fine and carried on, and the diff at that commit is one line of body copy in a docs markdown file, so it is not something this branch can cause.

Not re-running it: the rerun endpoint returns 403 for this session. Adding a retry to the step would mean editing `lighthouse.yml`, which is outside what this throwaway branch is for. The next push here re-triggers the workflow anyway, which serves as the re-run.

---

elevatebart:
**Publish to Cloudflare Workers** is red on `b95f0c4`, and that red is the result this branch was opened to get, not a defect to fix.

```
1 /_astro/ reference(s) point at files missing from dist:
  /_astro/edarras-sm.CIX2jR8x.png  <- 15 page(s): blogs/2023-10-05-announcing-kestra-funding-...
```

`b95f0c4` changes one line of body copy in a docs markdown file. No scss is touched, so `layoutDigest()` (#5599) stays quiet, pages restore from the warm cache, and 15 restored blog pages end up linking an image this build never emitted. Build was 185s against ~300s cold, so the restore path did most of its job, then shipped dead references.

Notable: the missing asset is an **image**, not a stylesheet. [withastro/astro#17976](https://github.com/withastro/astro/pull/17976) and the `cb815d8` candidate both address CSS hashing and neither covers this. Astro has a carry-over path for restored pages' static images (`cache.previousStaticImages` in `core/build/generate.js`), so this looks like a bug in that rather than a missing feature.

Run log for the record:

| Commit | astro | Build | Guard |
|---|---|---|---|
| `382c667` cold | `pkg.pr.new` cb815d8 | 300s | pass |
| `ecb4841` warm, sass partial | cb815d8 | 263s | inconclusive |
| `b0895a8` seed | ^7.3.1 | 316s | pass, full rebuild |
| `01d9c1c` warm, sass partial | ^7.3.1 | 288s | inconclusive |
| `b95f0c4` warm, markdown | ^7.3.1 | 185s | **fail** |

Both sass-partial runs are inconclusive by construction: `layoutDigest()` folds a digest of every `.scss` under `src` into each page's `cacheKey`, so any stylesheet edit invalidates every content page and nothing stale can survive to be caught.

Leaving this red. Not for merge.

---

elevatebart:
`Lint & Astro Check` failed on `b09cece` at the **Set up reviewdog** step, which downloads the reviewdog binary. `Run astro check` and `Run lint` were skipped as a consequence, so nothing in this branch was actually linted. Unit tests passed (315).

Same shape as the earlier `Lighthouse shard 1` failure: a tool installer dying on a runner, before anything in the diff is exercised. Not re-running it, the rerun endpoint returns 403 for this session; the next push re-triggers the workflow.

Worth noting for whoever reads this later: a reviewdog setup failure turns into a **skipped** lint rather than a failed one, so a genuine lint error on this commit would be invisible right now.

---

elevatebart:
Closing. This was a throwaway test bed for `experimental.incrementalBuild`, never for merge. The branch has been stripped back so its tree is identical to `main`.

### What it established

`.md.ts` endpoints restored while every `.astro` page route did not, on builds where nothing changed, so a warm build cost the same as a cold one (~300s).

The cause was **cache lineage, not Astro and not the page code**. Every run saved a ~726 MB entry against a 10 GB repo-wide quota shared with all open PRs, so entries were evicted about as fast as they were written and a run's restore fell back to an older entry rather than its predecessor. Against a stale entry `dependencyHash` mismatches for every page route, while endpoints keep hitting because their inputs are content digests that barely move.

Two commits on main dismantled it: #5621 cut entries to 22 MB, and #5626 prunes superseded entries so each ref keeps exactly one. Current measurement on this branch: **restored=2317 of 2578**, build 175s.

### Hypotheses tested and discarded

| Suspect | Result |
|---|---|
| Cached passthrough images (`node_modules/.astro/assets`) | not the cause: full 728 MB cache restored, probes still 20/20 |
| #5621's source changes | not the cause: yesterday's `src/` with today's workflow, still 20/20 |
| Astro version (`pkg.pr.new` vs 7.3.1) | not a variable, identical pin both sides |
| A route's own stylesheet | no effect |
| Chunk-name churn | dead: the probe routes reference no hashed asset at all |

A minimal Astro 7.3.1 reproduction did **not** reproduce, under the Cloudflare adapter with `prerenderEnvironment: "node"`, with vue/mdx/expressive-code/icon/fonts, or across a fresh `npm install` with the cache restored over it. That is what stopped an incorrect upstream issue being filed.

### Worth knowing for later

A conflicted PR produces **no workflow runs at all**, silently. This branch hit that twice and each time it looked like CI had died.

Separately, the ~140 static pages (`use-cases/*`, index and marketing pages) can never be restored: `cacheKey` only exists on objects returned from `getStaticPaths`, and a route with a concrete pathname never calls it. That is a real Astro gap and is being written up separately.

---

PR #5621, opened 2026-09-14
perf(build): stop re-fetching the API and re-spawning git for every prerendered page

elevatebart:
## Why

`experimental.incrementalBuild` does its job: in the deploy for a one-line markdown change (run 6325) 2233 pages were restored in ~7s and only the changed page re-rendered. The build still took 3m08s because of two things the flag cannot touch:

- Astro rebuilds the three Vite/Rolldown bundles from scratch every time (1m20s to 1m40s depending on the runner). That is the floor and this PR does not move it.
- The 263 routes without a `cacheKey` spent 75s rendering, and almost none of it was HTML. `/plugins/subgroups` (3.6 MB) was downloaded about 20 times, once per marketing page or component calling `$fetchApiCached`, at ~2.2s each. The sitemaps ran `git log -1` once per content file: ~800 spawns, 28s, and with the depth-1 checkout every one of them returned the deploy commit, so ~1000 URLs claimed to change on every deploy.

## What

- `$fetchCached` / `$fetchApiCached` memoize GET responses for the duration of a production build. Guarded by `import.meta.env.SSR && PROD` and `navigator.userAgent !== "Cloudflare-Workers"`, so the worker keeps its 1h edge cache and the browser bundle is untouched. Failures are not memoized; payloads are shared and read-only (`fetchFullCatalog` copies before appending).
- `gitLastModified` runs one `git log --no-renames --format=%cI --name-only` and answers from a map shared by all eight sitemaps. Output matches the old per-file command for all tracked files against the full history, 0 mismatches. `--no-renames` matters: rename detection reads blobs, and on a blobless clone that downloads them (4m21s measured).
- The deploy job keeps its depth-1 checkout and then fetches this commit's history without blobs (`git fetch --unshallow --filter=blob:none --no-tags origin "$GITHUB_SHA"`, ~9 MB, 4s to 11s on the runner, skipped on a complete clone), so the sitemap dates become real: 39 distinct days across the docs tree instead of one. `actions/checkout` with `fetch-depth: 0` + `filter` was tried first and took 22s to 1m51s because it fetches the 900 MB working tree on demand afterwards.
- Blog author avatars go through `getImage({ width: 96 })`, 2x the rendered 48px. Preview builds use the passthrough image service, so every transform is a local copy and Astro prunes originals no rendered page referenced; blog pages spread the raw `ImageMetadata` onto ``, so restored posts by authors absent from the listing pages lost their avatar (withastro/astro#17993, reproduced in runs 34878955136 then 34879396089). Transforms are recorded per page and replayed for restored pages. In production this moves avatars onto Cloudflare Image Resizing as 96px webp (1.4 kB vs 4.6 kB per avatar), a deliberate change.
- If the dangling-links check still fails (astro#17974, #17993), the job discards the incremental manifest, rebuilds from scratch and re-checks strictly, with a warning annotation, and pings Slack on main (non-fatal). Exercised end to end on the throwaway #5622: check fails, rebuild 2m46s, strict re-check passes, deploy green with the annotation.
- `node_modules/.astro/assets` (1683 passthrough copies, ~700 MB per PR push against a 10 GB quota, regenerated in seconds) is deleted in a final `always()` step before the cache save: 22.0 MB saved instead of 728 MB. Negated `path` patterns cannot do this because actions/cache hands tar the directory itself.
- `actions/setup-node` caches the npm download directory. The docs contributor guide documents the `updated` front matter field, which is now the first source for a page's `lastmod`.

## Measured (preview deploys of this PR, production main for the baseline)

| | main (8 runs) | cold, no cache | warm, 1 markdown file | warm, nothing changed |
|---|---|---|---|---|
| Checkout (+ history fetch) | 21–33s | 1m46s | 55s | 36s + 11s |
| Build step | 3m08s–4m23s, median 3m20s | 3m29s | 2m02s | 1m42s |
| of which prerender | 1m19s | 1m01s (all ~2500 pages) | 14s (233 rendered, 2263 restored) | 15s |
| Deploy | 25s | 1m00s | not reached (check failed, see above) | 22s |
| Job total | 5m00s–6m45s, median 5m08s | 7m19s | 3m51s | 3m47s |

## Verification

- 328 unit tests pass (13 new, covering the memo guards, the batched git parsing and its argv)
- `astro check`: 0 errors; oxlint clean on the changed files
- Preview deploys green on every head; `/sitemap/docs.xml` renders in 7ms (was 17.8s)

https://claude.ai/code/session_01MwrJi7dfhw6o1HC97SYgAD


iitzIrFan (changes requested):
Nice piece of work overall — the prerender phase really does drop from 1m43s to 14.75s, and I verified the risky parts rather than taking the description's word for them. Two things need to change before this can land.

## Verified, so please don't re-litigate these

I checked the parts I expected to push back on, and they hold up:

- **Batched `git log` parity.** I ran the old per-file command and the new parser over **all 5217 tracked files** against 6220 commits: **0 mismatches, 0 missing**, including the 6 paths containing spaces. Stronger than the 3341-file figure in the commit message.
- **`--no-renames`.** Built the map with and without rename detection: identical key counts (12897) and **0 date differences**. It is purely a perf flag here, no semantic change.
- **The workerd guard.** Ran this repo's own `@cloudflare/workerd-darwin-arm64` binary with `compatibility_date = "2026-03-01"` + `nodejs_compat`: `navigator.userAgent === "Cloudflare-Workers"` at module scope and in-request. The exclusion is sound.
- **Memoization semantics.** Key construction, non-GET bypass, and the failure-eviction are correct — the `.catch` is the first-registered reaction on `pending`, so there's no window where it could delete a successor entry.
- **The problem is real.** Production `/sitemap/docs.xml` right now has **565 of ~571 URLs sharing the deploy commit's timestamp**. Good catch.

---

## 1. The Actions cache exclusion is a no-op (blocking)

`.github/workflows/deploy.yml:84-86`

Measured from the saved-cache sizes in the logs:

| Run | Exclusion present? | Cache saved |
|---|---|---|
| `34879638759` (#5608 preview) | no | **726,655,340 B** |
| `34882831570` (this PR, preview) | yes | **728,038,986 B** |

So the ~693 MB is still being uploaded. `@actions/glob` expands the positive `node_modules/.astro` to all descendants, but the negative pattern matches only the directory entry itself:

```yaml
path: |
  node_modules/.astro
  !node_modules/.astro/assets/**
```

This isn't just a missed optimisation — changing `path` changes actions/cache's version hash, so **every existing `astro-build-*` cache is invalidated on merge** (one guaranteed cold rebuild) and the workflow keeps a comment claiming a 670 MB saving that the next maintainer will trust. Please fix the glob and re-measure, or drop `fc73d3ad8` from the PR.

## 2. Production avatar URLs change, and the commit says they don't (blocking)

`src/pages/blogs/[...slug].astro:58`

`getImage({ src: original })` with no options isn't preview-only. `baseService.validateOptions` fills `format: "webp"` when unset, and since the Cloudflare service isn't a local service, `getURL` emits:

```
/cdn-cgi/image/onerror=redirect,format=webp/_astro/<avatar>.<hash>.png
```

Every blog author avatar moves onto Cloudflare Image Resizing. "Production uses the Cloudflare image service and never prunes" is true about pruning but reads as "production is unaffected", and it isn't.

Worth noting too: `BlogDetails.vue:13-17` and `BlogMobileHeader.vue:10-14` put `width="48"`/`"40"` *after* the `v-bind`, so no width reaches `getImage` — the new hop resizes nothing and still delivers the 250×250 original.

Either resolution is fine by me:

- **Keep production output stable:** pass `{ format: original.format }`.
- **Optimise deliberately:** pass `width: 96` / `80` (2× the rendered size) and say so in the commit message.

---

## Strongly recommended

**3. `src/utils/blueprints/catalog.ts:21`** — now that the payload is memoized, `first.results` is the cache's own array and this `push` mutates it; each call appends ~800 more entries (the `seen` dedup happens to hide it in the output). Unreachable today only because all three callers are `prerender = false`, but that's one flag flip away. One line: `const results = [...(first?.results ?? [])]`. A note in `memoizeGet`'s comment that cached payloads must be treated as read-only would help too — I audited every other prerendered consumer and this is the only one that mutates, but it's an easy trap to fall into now.

**4. `.github/workflows/deploy.yml:104-110`** — the fallback has never actually run; this PR's own build logged `All /_astro/ references resolve to files in dist`. The mechanics look right to me (`bash -e`, the paths match `INCREMENTAL_CACHE_FILE`/`INCREMENTAL_OUTPUT_DIR`, and the post-step saves a clean cache so it self-heals), but I'd want it exercised once on a throwaway branch before it sits on the production deploy path.

**5. `.github/workflows/deploy.yml:99`** — `continue-on-error: true` turns a production-safety gate into a warning annotation on a green job. If the upstream Astro gaps recur, main silently double-builds forever and nobody notices. The job already has `SLACK_WEBHOOK_URL` — could we ping on the fallback firing, or keep it strict on `refs/heads/main`?

## Nits, take or leave

- `src/utils/sitemap.test.ts` — nothing asserts the git argv, so `--no-renames` and `core.quotePath=false` can both be deleted with every test still green. One `expect(args).toContain("--no-renames")` closes it.
- `.github/workflows/deploy.yml:62` — `--unshallow` hard-fails on a complete repository; fine with the current checkout defaults, but it breaks the day someone sets `fetch-depth: 0`. Also, this fetched 116 remote branches in the observed run — only HEAD's history is read.
- `src/pages/blogs/[...slug].astro:50` — `authors ?? (author ? [author] : [])` diverges from `useBlogAuthors.js`, which falls back to `[author]` when `authors` is an empty array. No post hits this today (I checked all 92 author image refs), but the old code was immune because it passed the whole map.

iitzIrFan:
Re-reviewed `012e11c7d`. Five of the six items are fixed and verified. Before anything else:

## My glob advice was wrong — sorry

I told you to use `!node_modules/.astro/assets/**`. That was never going to work, and I should have checked the mechanism instead of guessing at it.

| Run | `path` exclusion | Cache saved |
|---|---|---|
| `34879638759` (#5608) | none | 726,655,340 B |
| `34882831570` | `!…/assets` | 728,038,986 B |
| `34937504426` (after the fix) | `!…/assets/**` | **728,075,504 B** |

The actual cause is in `@actions/cache`, `lib/internal/cacheUtils.js:57-65`:

```js
const globber = yield glob.create(patterns.join('\n'), {
    implicitDescendants: false
});
```

With `implicitDescendants: false`, `node_modules/.astro` resolves to **exactly one entry — the directory itself**. `tar` does all the recursion afterwards. The glob never enumerates anything inside that directory, so **no `!` pattern can exclude its contents**, with or without `/**`.

Two fixes that do work. I'd take the second — it keeps working when Astro adds a new cache artifact, and it means never editing `path` again:

```yaml
# Option A — enumerate the children, so tar only recurses what is listed
path: |
  node_modules/.astro/data-store.json
  node_modules/.astro/incremental-build.json
  node_modules/.astro/dist
  node_modules/.astro/fonts
```

```yaml
# Option B — delete it before the post-cache step runs (last among the main steps)
- name: Drop the passthrough image cache before it is saved
  if: always()
  run: rm -rf node_modules/.astro/assets
```

Related: run `34937504426` restored **0 pages** and took 3m24s against the previous 1m39s, because editing `path` changes actions/cache's version hash. That cold rebuild is a cost per `path` edit, which is the other argument for Option B.

## One new issue from the fix commit

**`.github/workflows/deploy.yml:108-112` — a Slack outage can block a production deploy.**

```bash
| curl -sS -X POST -H 'Content-type: application/json' --data @- "$SLACK_WEBHOOK_URL"
```

No `|| true`. A revoked webhook or a Slack 5xx makes curl exit non-zero and the step fail. An `if:` containing no status function gets an implicit `success()`, so `Rebuild without the incremental cache` **and** both deploy steps are skipped — a recoverable stale-cache situation turns into a failed deploy on main, which is the only place this step runs. `|| true` or `continue-on-error: true` closes it.

Note this won't show up in the drill either: the step is gated on `github.ref == 'refs/heads/main'`, so it's skipped on PR branches.

## Verified fixed

| Item | Result |
|---|---|
| Avatar sizing | Rendered `w=96&h=96&f=webp`; height derives correctly from the 250×250 source. **1,384 B vs 4,632 B** — ~70% smaller. The `/cdn-cgi/image` hop now earns its place, and the commit message says so. |
| `allAuthors` fallback | Matches `useBlogAuthors.js:3-9`. |
| `catalog.ts` | `[...(first?.results ?? [])]` closes the mutation hazard; the `memoizeGet` comment now states payloads are shared and read-only. |
| History fetch | **Verified in CI**: the shallow guard plus `origin "$GITHUB_SHA"` fetched **0 branches, down from 116**. The `||` fallback did not fire. |
| Sitemap argv assertions | `--no-renames` and `core.quotePath=false` asserted, filtered on `args.includes("log")` so `rev-parse` isn't miscounted. |
| Local checks | vitest 21 files / 328 tests · `astro check` 0/0/0 · oxlint clean · blog pages 200 with avatars resolving. |

## On the drill (#5622)

Nicely constructed — reverting the avatar spread plus a one-line markdown change so a content-only push can follow. But run `34937525624` is **step 1, the cold build**. The restore-from-cache push that actually triggers the prune hasn't happened yet, so `Rebuild without the incremental cache` still has never executed. I'd like to see step 2 green before this merges.

## Left to do

1. The cache exclusion, for real this time (Option A or B above).
2. `|| true` on the Slack curl.
3. Drill step 2.

Everything else from the first round is done. Thanks for the quick turnaround, and again, sorry for the bad steer on the glob.

elevatebart:
Second round done, head is `594ceda3`.

**1. Cache exclusion, for real.** Took Option B: a final `if: always()` step removes `node_modules/.astro/assets` before the post-job save, `path` is back to `node_modules/.astro` and never needs editing again. For the record, Option A worked too and went in as `5bd589e7` on the way:

| Run | Approach | Cache saved |
|---|---|---|
| `34937504426` | `!…/assets/**` | 728,075,504 B |
| `34938165613` | explicit list (data-store, manifest, dist) | 21,508,462 B |
| `34938393613` | `rm -rf …/assets` before save | **22,040,016 B** |

The 0.5 MB difference is the fonts cache the explicit list would have dropped. Because `path` is back to its original value, `34938393613` also restored the pre-existing cache (warm build, 2m04s), so merging will not cost a cold rebuild after all.

**2. Slack curl.** Falls through to `echo "::warning::Slack notification failed; continuing"`, so a webhook outage can no longer skip the rebuild or the deploy steps. Agreed it will not show in any preview run, it is main-only by design.

**3. Drill step 2** ([#5622](https://github.com/kestra-io/docs/pull/5622), [run 34938135935](https://github.com/kestra-io/docs/actions/runs/34938135935)): restore hit, warm build 2m04s, `Check for dangling /_astro/ links` failed under `continue-on-error`, `Rebuild without the incremental cache` ran (2m46s, prerender 55.8s from scratch), strict re-check passed with `All /_astro/ references resolve to files in dist`, preview deployed, job green with the `Incremental cache discarded` annotation. Drill PR closed; its branch stayed behind because this session's credentials cannot delete remote branches, feel free to remove it.

Thanks for measuring the saves and for the `implicitDescendants` pointer, that was the missing piece.

---

iitzIrFan (approved):

PR #5624, opened 2026-09-15
ci: cache node_modules across workflows, compile SCSS with sass-embedded, split the Bootstrap prelude

elevatebart:
### Gain

Cloudflare deploy preview (`Publish to Cloudflare Workers`), PR that changes markdown only:

| | cold cache (first push of a PR) | warm cache (every push after) |
|---|---|---|
| before this PR | about 5m40s (estimated) | about 3m58s (estimated) |
| node_modules cache (commits 1 and 2) | 5m29s ([#6372](https://github.com/kestra-io/docs/actions/runs/34942319197), measured) | 3m24s ([#6373](https://github.com/kestra-io/docs/actions/runs/34942570095), measured) |
| plus sass-embedded (commit 3) | 4m14s ([#6377 attempt 1](https://github.com/kestra-io/docs/actions/runs/34945901692/attempts/1), measured) | 2m48s ([#6377 attempt 2](https://github.com/kestra-io/docs/actions/runs/34945901692), measured) |

Warm deploy 3m58s to 2m48s, cold 5m40s to 4m14s. The `Build` step alone: 1m40s to 1m05s warm, 3m13s to 1m44s on a full build. The prelude split (#5627, merged into this branch) adds no build time on top of sass-embedded, measured flat within noise locally and on CI ([#6393](https://github.com/kestra-io/docs/actions/runs/34951228847), 1m35s for a near-full render); it is a hygiene and CSS-size change, see below.

Cold: the node_modules cache and the incremental build cache both miss, once per PR (new merge ref) or per lockfile bump. Warm: both hit. The "before" cells are the measured runs plus what the install used to cost: 41s `npm install` against 7s on a hit and 30s on a miss (`npm ci` + save).

For scale: the same one-file markdown deploy took 4m59s on main on Sep 14 ([#6325](https://github.com/kestra-io/docs/actions/runs/34845686131)), before #5621 and this PR.

Other PR workflows: Lint & Astro Check 2m18s to 1m52s (install 42s to 9s), Lighthouse full build 3m26s to 2m16s and its install 41s to 9s, each of the four Lighthouse shards about 26s shorter. Full table in the [measurements comment](https://github.com/kestra-io/docs/pull/5624#issuecomment-5676633550); bundle profile, the sass-embedded measurement and the prelude inventory in the comments after it.

### Why

`npm install` ran on every job that touches the tree: the deploy job, lint and astro check, the lighthouse build and each of its four shards. On the deploy job that is 41s with a warm `~/.npm`, spent extracting 980 packages and running the native postinstalls (sharp, workerd, esbuild). Since #5621 the build itself is about 2 minutes on a markdown-only change, so the install became a visible slice of the job.

Profiling that build showed the bundle spending 47% of its main-thread time in the dart2js `sass`: the `additionalData` prelude imports bootstrap's functions, variables, maps, mixins and utilities into each of the 350 `lang="scss"` blocks, once per Vite environment, about 1050 evaluations per build.

### What

- New composite action `.github/actions/node-modules`: restore `node_modules` from the Actions cache, `npm ci` on a miss, save right after the install. Used by `deploy.yml`, `astro-check.yml` and both jobs of `lighthouse.yml`.
- Key: `node-modules-<os>-<arch>-<ImageOS>-<ImageVersion>-<hash(.nvmrc, package-lock.json)>`. No `restore-keys`: a partial match would hand the build a native binary from another Node or runner image, and a cold install beats that.
- `npm ci` replaces `npm install` on a miss. The lockfile is in sync (`npm ci --dry-run` passes on this tree).
- The save happens before any build writes `node_modules/.astro`, so the incremental build cache keeps its own per-ref entry and neither cache captures the other. No exclusion pattern needed, which is just as well since they do not work on a matched directory (see the passthrough image note in `deploy.yml`).
- The lighthouse shards still add lighthouse and chrome-launcher with `npm install --no-save` afterwards; that lands after the save, so the cached tree stays the exact lockfile.
- The runner image is read in a shell step with an `unknown` fallback, so the key does not rely on those variables being visible to the `env` context.
- `sass-embedded` in `dependencies` next to `sass`, both `^1.104.1` and locked to the same Dart Sass version; bump them together, the byte-identical CSS between the two compilers (358 of 358 units, verified in review) holds only while they match. Vite's `loadSassPackage` resolves the embedded one first, so no config change. Local measurement: prerender build 112.8s to 52.5s, SSR build 35.5s to 17.5s, client build 37.3s to 18.5s, bundle -52%, GC 22s to 5s. It is a platform binary, which the cache key above already covers. Sass runs at build time only; the worker ships the compiled CSS.
- Prelude split (#5627): `_variable.scss` holds project values plus pinned Bootstrap literals, a verbatim copy of Bootstrap's breakpoint mixins and three one-line shims; Bootstrap's preamble moves to `_bootstrap-config.scss`, imported once by `vendor.scss`; the `:root { --bs-* }` rule moves to the end of `app.scss`. No component file touched. Emitted CSS is byte-identical apart from the removed duplicate `:root` copies (about 45 KB raw, 6.7 KB gzipped across the site, four small stylesheets now inline under Astro's 4 KB threshold). One declaration in `app.scss` changes on purpose: `--bs-link-color-rgb` goes from `var(--ks-content-link)` to `var(--ks-content-color-highlight)`, which is the value the prelude's per-component copies were already winning with in every bundle (they came after `app.scss` in the cascade), so the effective value is unchanged and `app.scss` now states it. Verification and the pixel comparison of the regenerated snapshots are in #5627 and the comments.
- The Linux snapshot refresh captured from this branch (08ec73c) is reverted in 1ce8ce8. main's own capture of the same 54 files is #5628, where the refresh belongs.

### Expected

- First run on a given lockfile, Node and image: miss, `npm ci`, save.
- Every following run, in any of the three workflows: restore instead of install. PR runs read main's entry, so a lockfile bump pays the install once.
- A lockfile change also invalidates Astro's incremental manifest (it hashes the lockfile), so that run is cold on both caches; the run after it is warm again.

https://claude.ai/code/session_015nu7pdX9cohW17FCYWfRpM


elevatebart:
### Measurements

Install step per job. "Hit" is the whole composite step: image lookup, restore, skipped install and skipped save.

| job | before (`npm install`) | miss (`npm ci` + save) | hit (restore) |
|---|---|---|---|
| Deploy, Publish to Cloudflare Workers | 41s ([#6325](https://github.com/kestra-io/docs/actions/runs/34845686131), [#6371](https://github.com/kestra-io/docs/actions/runs/34940667915)) | 30s = 25s + 5s ([#6372](https://github.com/kestra-io/docs/actions/runs/34942319197)) | 7s ([#6373](https://github.com/kestra-io/docs/actions/runs/34942570095)) |
| Lint & Astro Check | 42s ([run 1580](https://github.com/kestra-io/docs/actions/runs/34942319237)) | n/a | 9s ([run 1581](https://github.com/kestra-io/docs/actions/runs/34942570173)) |
| Lighthouse, Build site | 41s ([run 3946](https://github.com/kestra-io/docs/actions/runs/34940667820), main) | n/a | 9s ([run 3948](https://github.com/kestra-io/docs/actions/runs/34942570042)) |
| Lighthouse, shards 0 to 3 | 38s to 42s (run 3946) | n/a | 6s to 8s (run 3948) |

Notes:

- Lint & Astro Check is the one job whose other steps are identical between the two runs: 2m18s to 1m52s. Deploy totals are not comparable here, the first PR build had no incremental cache for the new merge ref.
- The miss path is also cheaper than before: `npm ci` skips the resolution `npm install` does (25s vs 41s), and the save adds 5s.
- On the shards, the ad hoc `npm install --no-save lighthouse chrome-launcher` went from 10s to 17s on top of a restored tree (npm reconciles the full tree from disk instead of the one it just built). Net per shard is still about 50s to 24s.
- One cache entry per lockfile, Node version and runner image. After merge, PR runs read main's entry, so a lockfile bump pays the install once.

---

elevatebart:
### Bundle profile

Setup: a 4 vCPU sandbox, Node 26.3.1, Astro 7.3.1, rolldown 1.2.8, `sass` 1.104 (dart2js). `ASTRO_TIMER_PATH` for the phase split, `node --cpu-prof` on `astro build` for the attribution, API served by the lighthouse fixture server since api.kestra.io is not reachable from there (the generate phase therefore aborted on `/v1/plugins/metadata` after 1585 pages; the bundle phases are unaffected). Absolute numbers are about 2x CI, the split is what matters.

Phase timer:

| phase | cold (no content cache) | warm |
|---|---|---|
| Sync content layer | 63s | 3s |
| Prerender build | 139s | 113s |
| SSR build | 34s | 35s |
| Client build | 41s | 37s |
| bundle, the three builds | 214s | 186s |

CPU profile of the bundle window, main thread self time, 185.5s sampled:

| share | self time | where |
|---|---|---|
| 47% | 87.6s | sass (`sass.dart.js`) |
| 12% | 22.6s | idle, main thread waiting on rolldown's native threads |
| 12% | 22.0s | garbage collector |
| 7.5% | 14.0s | rolldown JS bindings, `emitFile` alone 7.6s |
| 5.7% | 10.5s | (program) |
| 3% | 5.7s | svgo |
| under 1.2% each | | astro, vite, @vue/compiler-*, @babel/parser, shiki, postcss, @astrojs/compiler-rs |

Same shape in every environment: sass is 46% of the prerender build, 50% of the SSR build, 47% of the client build.

### Why sass

`vite.css.preprocessorOptions.scss.additionalData` prepends `@use "/src/assets/styles/variable" as *` to every style block, and `_variable.scss` imports bootstrap's `functions`, `variables`, `variables-dark`, `maps`, `mixins` and `utilities`. There are 350 `lang="scss"` blocks (210 in `.astro`, 140 in `.vue`), each compiled once per environment, so bootstrap's SCSS preamble is parsed and evaluated about 1050 times per build. The `:root { --bs-* }` rule inside the prelude is also emitted into every component's CSS.

### Levers

1. Add `sass-embedded` as a devDependency. Vite resolves it ahead of `sass` when present: the Dart VM over a protocol instead of dart2js on the main thread. No config change. If sass drops from 88s to about 35s here, the bundle goes from 186s to about 130s; scaled to CI's 1m42s that is roughly 1m10s. Needs measuring, the ratio varies with how much of the time is the protocol round-trips.
2. `vite.css.preprocessorMaxWorkers: true`. Vite 8 still ships the option; it runs preprocessors in worker threads, which takes sass off the main thread on the 4 vCPU runners. Overlaps with 1, measure both ways.
3. Cut the prelude. `@use` bootstrap's `functions`, `variables` and `mixins` only where a block needs them, or precompute the variables the components actually read into a small partial without `maps` and `utilities`, the map-merge heavy part. Also removes the duplicated `:root` rule. Larger change: 350 files or the prelude's semantics.
4. Smaller: GC is 12% with a 2.4 GB heap peak in the prerender build, so `NODE_OPTIONS=--max-old-space-size=4096` in the deploy job is a cheap experiment. svgo at 3% comes from `experimental.svgOptimizer` running on every SVG import per environment.

Rolldown's own Rust work only shows up as the 12% idle, so this JS-side view bounds what can be cut: about 75% of the bundle's main-thread time is sass, GC and rolldown glue.

<details><summary>Top functions, bundle window, self time</summary>

```
22.6s  (idle)
22.0s  (garbage collector)
10.5s  (program)
 7.6s  emitFile  rolldown/dist/shared/bindingify-input-options.mjs
 4.6s  _Universe_evalInEnvironment  sass/sass.dart.js
 4.1s  (anonymous)  sass/sass.dart.js:4978
 2.6s  peekChar$1  sass/sass.dart.js
 2.0s  instanceType  sass/sass.dart.js
 2.0s  _wrapJsFunctionForAsync  sass/sass.dart.js
 1.8s  read  (fs)
 1.8s  _completeWithValue$1  sass/sass.dart.js
 1.8s  _Future__propagateToListeners  sass/sass.dart.js
 1.7s  _eval$1  sass/sass.dart.js
 1.6s  get$hashCode  sass/sass.dart.js
 1.6s  openFileHandle  (fs)
 1.5s  readFileHandle  node:internal/fs/promises
 1.5s  readChar$0  sass/sass.dart.js
 1.4s  _asyncAwait  sass/sass.dart.js
 1.4s  $index  sass/sass.dart.js
 1.4s  internalSet$2  sass/sass.dart.js
```

</details>

---

elevatebart:
### sass-embedded, measured

Same sandbox and setup as the profile above, `sass-embedded@1.104.0` installed with `--no-save` next to `sass@1.104.0`, nothing else changed. Vite's `loadSassPackage` picks the embedded one when it resolves.

| phase | `sass` (dart2js) | `sass-embedded` | delta |
|---|---|---|---|
| Sync content layer | 2.7s | 2.8s | 0 |
| Prerender build | 112.8s | 52.5s | -53% |
| SSR build | 35.5s | 17.5s | -51% |
| Client build | 37.3s | 18.5s | -50% |
| bundle, the three builds | 185.5s | 88.6s | -52% |
| Total build (generate aborts at the same page in both) | 233.6s | 137.0s | -41% |

Both runs under `--cpu-prof`, so the overhead is the same on both sides. Scaled to CI's 1m42s bundle, the deploy job's `Build` step should land around 1m for a markdown-only change, so the warm-cache deploy from the table at the top would be roughly 2m35s instead of 3m24s.

CPU profile of the new bundle window, 88.6s sampled:

| share | self time | where |
|---|---|---|
| 26% | 23.3s | idle: rolldown's Rust threads and the Dart VM process |
| 15.5% | 13.7s | rolldown JS bindings, `emitFile` 8.1s |
| 13.5% | 11.9s | sync fs calls (read, open, close, fstat) |
| 11.6% | 10.3s | (program) |
| 6.2% | 5.5s | svgo, now the largest JS plugin |
| 6.1% | 5.4s | garbage collector, down from 22.0s: the sass heap was the GC pressure |
| 2.5% | 2.2s | sass-embedded client, protobuf, rxjs |

Heap peak in the prerender build dropped from 2.4 GB to 1.0 GB.

Proposal: a separate one-line PR adding `sass-embedded` as a devDependency, `sass` kept until dev mode is checked. It is a platform binary (`sass-embedded-linux-x64` and friends as optional deps), which the node_modules cache key from this PR already accounts for. Lever 3 from the profile comment, trimming the bootstrap prelude, is still worth doing afterwards: it shrinks the remaining sass work and the CSS output, but it is a bigger change.

---

elevatebart:
### Prelude inventory, for the follow-up PR

What the 362 SCSS sources (210 `.astro` blocks, 140 `.vue` blocks, 12 standalone `.scss` files) take from Bootstrap through the `additionalData` prelude. Identifiers defined in `_variable.scss` itself are excluded; the intersection is with what `functions`, `variables`, `variables-dark`, `maps` and `mixins/*` define.

| Bootstrap identifier | kind | sources | note |
|---|---|---|---|
| `media-breakpoint-up` | mixin | 149 | |
| `media-breakpoint-down` | mixin | 140 | |
| `media-breakpoint-between` | mixin | 12 | |
| `media-breakpoint-only` | mixin | 1 | |
| `$spacer` | variable | 39 (37 component blocks, 2 global sheets) | Bootstrap value `1rem` |
| `$line-height-base` | variable | 2 | `1.5` |
| `$primary`, `$gray-500`, `$pink`, `$zindex-fixed`, `$input-padding-x` | variable | 1 each | `BusinessValue.astro`, `VerticalsIntro.astro`, `Search.vue`, `LpHeader.astro`, `preview-access/Features.vue` |
| `$prefix`, `$zindex-modal` | variable | 1 each | `app.scss` only, a global sheet |
| `font-size` (rfs) | mixin | 2 | `SocialsList.vue`, `NavToc.vue` |
| `border-radius` | mixin | 1 | `NavSideBar.vue` |
| `subtract()` | function | 1 | `NavSideBar.vue`, `subtract(100vh, 9rem)` |

`$colors` matched 7 files but every hit is an `@each $id, $colors in ...` loop variable shadowing the Bootstrap map, so it is not a dependency.

Prelude-defined overrides that components read and that stay as literals: `$font-size-sm` (111), `$white` (50), `$font-size-lg` (43), `$border-radius-lg` (41), `$font-family-monospace` (19), `$font-size-base` (11), the `$h*-font-size` set, `$border-radius` (8), `$black` (6), `$box-shadow-*`, `$grid-gutter-width`. `$grid-breakpoints` and `$container-max-widths` are read directly only by `app.scss`.

144 of the 362 sources use no Bootstrap identifier at all.

### What the split needs

- A local `media-breakpoint-up/down/between/only` set over the literal `$grid-breakpoints` map. Bootstrap's own is `breakpoint-min`, `breakpoint-max`, `breakpoint-next`, `breakpoint-infix` plus the four mixins, about 60 lines, MIT, and copying it keeps the `max-width: N - 0.02px` semantics byte for byte.
- Literals for `$spacer: 1rem` and `$line-height-base: 1.5`, plus the five one-offs either as literals in the prelude (no component touched) or inlined in their five files.
- `font-size` (rfs), `border-radius` and `subtract()`: three components, replace with `font-size:`, `border-radius:` and `calc(100vh - 9rem)`, or ship three one-line shims in the prelude to keep components untouched.
- The global sheets (`vendor`, `app`, `layout`, `markdown`, `_buttons`, the two themes) compile once and import Bootstrap explicitly, with the project overrides ahead of `bootstrap/scss/variables` as today and the `$spacers`/`$font-sizes` `map-merge` between `variables-dark` and `maps`. The `:root { --bs-* }` rule moves there too instead of being emitted into every component's CSS.

### Verification

Build before and after, diff `dist/client/_astro/*.css` with hashes stripped: the only expected difference is the duplicated `:root { --bs-link-color ... }` block leaving the component CSS. Then the Playwright visual suite as the second gate.

---

iitzIrFan (approved):
Approving. The measurements hold up and the cache design is right.

Verified locally:

- **CSS output is unchanged under `sass-embedded`.** Compiled all 358 units — 7 global sheets plus the 351 `lang="scss"` blocks, each with the real `additionalData` prelude and the same `silenceDeprecations` list — under `sass@1.104.0` and `sass-embedded@1.104.0`: **358 identical, 0 differing, 0 errors**. Worth stating explicitly because `test:visual` only runs on `workflow_dispatch`, so nothing on a PR would have caught a regression here.
- **`npm ci --dry-run` passes** on this tree, so the miss path is sound.
- Key covers os/arch/ImageOS/ImageVersion + `.nvmrc` + lockfile with no `restore-keys`, the save lands before `node_modules/.astro` exists, and all three callers run `setup-node` before the composite action.

Non-blocking notes:

1. **`sass-embedded` is in `devDependencies` while `sass` is in `dependencies`.** Both are build-time only, and the repo already keeps build tools (`sharp`, `svgo`, `typescript`, `wrangler`) in `dependencies`. Any install with `--omit=dev` would silently fall back to dart2js sass — a 2x build-time regression with no error. Suggest moving it next to `sass`.

2. **Independent `^` ranges.** The byte-identical result above depends on both resolving to the same Dart Sass version. Fine as locked today; worth a note so a future bump moves both together.

3. **Cache quota.** The repo is at 10.92 GB / 30 entries, already past the 10 GB limit, so eviction is running. The 289 MB node_modules entry stays hot and survives, but the tail it trims is `astro-build` entries worth ~90s each against this one's ~35s. The actual driver is `astro-build-...-${github.sha}` creating a new ~694 MB entry per push (#5621 holds four). Separate PR, but worth doing before this cache has to compete for space.

4. **Fork PRs get their own cache scope** and can't read main's entry, so external contributors still pay the full install. Not a regression, just worth setting expectations on since community docs PRs are a large share of the traffic here.

5. Optional: `cache: npm` on `setup-node` in `deploy.yml` now only earns its keep on the miss path. Cheap to leave; cheap to drop.

The three commits are cleanly separable, so if `sass-embedded` misbehaves on someone's platform in dev mode, commit 3 reverts without touching the caching.

elevatebart:
Thanks for the 358-unit compile check, that is the verification this PR was missing.

Notes 1 and 2 are in 182d786: `sass-embedded` moved to `dependencies` next to `sass`, both on `^1.104.1`, and the lockfile resolves both to 1.104.1 (npm took `sass` from 1.104.0 to 1.104.1 with it, so the two Dart Sass versions stay equal). The commit message carries the bump-together rule since package.json cannot.

Notes 3 to 5, acknowledged and left as they are here:

- Cache quota: agreed it is the `astro-build-...-${{ github.sha }}` entries, one per push, that fill the 10 GB, and this PR's single hot entry is the wrong thing to blame or fix for it. Happy to open the separate PR: the sha suffix is what lets each run save at all, so the fix is either a purge step for a ref's older entries at the end of the job, or shrinking what the entry holds.
- Fork PRs: correct, they only see their own scope, so community docs PRs pay the install. Same as today.
- `cache: npm` on `setup-node` stays; it still shortens the miss path.

---

iitzIrFan (approved):
Re-approving on `182d786f4`; my earlier approval was recorded against `2cace12e3`.

`182d786f4` closes notes 1 and 2. Confirmed the lockfile resolves `sass` and `sass-embedded` both to 1.104.1, and the platform binaries with them (`sass-embedded-linux-x64`, `sass-embedded-darwin-arm64`).

Since my byte-identical result was measured at 1.104.0 and scoped to matching versions, I re-ran the same 358-unit compile — 7 global sheets plus the 351 `lang="scss"` blocks with the real prelude — at **1.104.1 on both sides**: **358 identical, 0 differing, 0 errors**. Unchanged.

All 12 checks green on this head. The run also confirms the cold-path prediction in the description: the lockfile bump invalidated the node_modules key and Astro's incremental manifest together, and Publish to Cloudflare Workers came in at 4m17s against the 4m14s cold estimate. The next push should be warm on both.

Notes 3 to 5 stand as discussed, nothing blocking. Happy to review the `astro-build` purge PR when you open it.

iitzIrFan (changes requested):
Requesting changes on one commit: please drop `08ec73ce5 test(visual): refresh linux snapshots`. Everything else checks out and I have verified the two SCSS commits below.

## `2cace12e3` + `182d786f4` — sass-embedded: no visual change

You ran the snapshot workflow on both `main` and this branch at 08:57, so I pulled both artifacts and compared them. All 54 snapshots have **identical dimensions** between the two builds — same width and same full-page height. A CSS change shifts page height; none did. 40 of 54 are byte-identical, `About-Us-desktop` differs by exactly 1 byte with zero differing pixels, and the densest difference band on `Cloud-desktop` shows identical layout, text and positions, differing only in glyph rasterization. With the 358/358 byte-identical CSS at 1.104.1, the compiler swap is settled.

## `19be3b6c4` (#5627) — the prelude split is correct

This arrived after my approval and my earlier result did not cover it: that compared two compilers on the same sources, and this changes the sources. So I compiled every unit against both preludes — 7 global sheets and 350 component blocks, same compiler both sides:

- **354 of 357 units change, and every single change is exclusively the duplicated `:root { --bs-* }` block leaving component CSS.** Zero differences of any other kind, zero compile errors. Exactly the outcome you predicted in the inventory comment.
- **No Bootstrap variable disappears.** `vendor.scss` still emits all 304 unique `--bs-*` with identical values, and it is imported by `src/components/layout.astro`, the only component that emits `<html>`, so coverage is global. The four that left `layout.scss` and `markdown.scss` (`--bs-link-color`, `--bs-link-hover-color`, `--bs-primary`, `--bs-link-color-rgb`) were duplicates of declarations `app.scss` still carries with their `--ks-*` mappings intact.
- **One value did change and it looks deliberate:** `app.scss`'s `--bs-link-color-rgb` goes from `var(--ks-content-link)` to `var(--ks-content-color-highlight)`. That matches what the old prelude's `:root` was injecting into every component block and therefore winning with, so the effective value is unchanged — you aligned `app.scss` to what actually applied. Nothing in `src/` reads the variable either way. Worth one line in the description so the next reader does not have to derive it.

Two process notes rather than objections. #5627 was merged into this branch with zero reviews, and the description here still lists the prelude split under **"Not in this PR ... it is a separate PR"** — that needs updating, since the PR now carries it. Separately, I could not find any importer for `src/assets/styles/layout.scss` or `markdown.scss` anywhere in `src/`; if they are dead, that is a nice follow-up cleanup.

## `08ec73ce5` — please drop this one

Applying the configured gate (pixelmatch at `threshold: 0.2` against `maxDiffPixelRatio: 0.002`) to those same two runs of functionally identical output:

| page | diff ratio | gate 0.200% |
|---|---|---|
| Cloud-desktop | 0.8181% | over |
| Get-Started-desktop | 0.6119% | over |
| VS-Page-sample--desktop | 0.2294% | over |
| Pricing-desktop | 0.1951% | marginal |
| Blueprints-Landing-desktop | 0.1872% | marginal |
| Home-desktop | 0.1856% | marginal |

Three pages fail against each other and four more sit within 10% of the threshold, so a baseline captured from one run does not make the suite green — the next run fails on those pages anyway. (Upper bounds: I did not implement pixelmatch's antialiasing detection, which would shave them.) `document.fonts.ready` and the 500 ms settle are already there and are clearly not enough. The fix is to stabilise what varies or widen the tolerance on those pages, not to rebaseline.

Four reasons to split it out:

1. 44 MB of binaries in git history cannot be undone after merge.
2. It is not needed for this PR's CI — the visual suite is `workflow_dispatch`-only, so nothing goes red without it.
3. A baseline regenerated from the branch under test can only hide a rendering change, never reveal one. Harmless this time because I checked, but the check should run the other way round.
4. It is already stale against its own PR: the snapshots were taken at 08:57 and the prelude commit landed at 09:12, so they do not reflect the branch's current CSS.

Drop it and I will re-approve on the new head. The refresh looks like real pre-existing drift worth landing — the last one was #4552 — just as its own PR, ideally after the reproducibility issue so the new baselines hold.

elevatebart:
Dropped: 1ce8ce8 reverts 08ec73c, so the 54 images are out of this PR's diff. The repo squash-merges (#5627 landed as one commit), so they never reach main's history from here either. #5628 keeps the refresh, and your reproducibility table is the argument for landing it after the capture noise is fixed rather than before.

Your pixelmatch numbers match what I measured with a plain per-pixel compare: 40 of 54 byte-identical, the rest desktop-only with mobile and tablet identical, which is why I read it as capture noise and not CSS.

On the two notes:

- `--bs-link-color-rgb` in `app.scss`: intentional, and now one line in the description. The old prelude emitted its `:root` block into every component's CSS, and those copies came after `app.scss`'s `:root` in every bundle (byte offsets in the layout bundle: `app.scss` at 268,597, the next copy at 280,755), so `var(--ks-content-color-highlight)` was already the effective value. The split keeps that value and drops the line it shadowed.
- `layout.scss` and `markdown.scss` are not dead: `app.scss` imports `layout` on line 5, and `layout.scss` imports `markdown` on line 3 for its `markdown-styles` mixin. Sass-side imports, so a grep for the file names in components finds nothing.

The description was updated when #5627 merged in; the "Not in this PR" paragraph is gone and the split has its own bullet.

---

elevatebart:
@iitzIrFan It's dropped

iitzIrFan (approved):

iitzIrFan (approved):
Re-approving on `1ce8ce8e3`. This supersedes my `CHANGES_REQUESTED` on `19be3b6c4` — the snapshot commit is out and the net diff is back to 10 files.

On the revert rather than a drop: I checked whether that still leaves the 44 MB in history, and it does not. The repo has `allow_merge_commit=false` with squash and rebase enabled, and main's history is linear, so a squash lands only the net diff and the snapshot blobs never reach main. Equivalent to dropping it for history purposes.

Nothing needed re-verifying: `src/` is byte-identical to `19be3b6c4`, the tree I compiled, so the prelude result carries over — 354 of 357 units change, every change exclusively the duplicated `:root { --bs-* }` block leaving component CSS, zero other differences, zero errors. Together with 358/358 byte-identical CSS across the compiler swap, all three changes are verified.

Thanks for rewriting the description too. The `--bs-link-color-rgb` note matches what I derived independently — the prelude's per-component copies came after `app.scss` in the cascade and were already winning, so the effective value is unchanged.

Checks: everything green except the four Lighthouse shards, still running when I posted. They are benchmark reporting rather than a correctness gate, so I am not holding the approval on them — worth a glance before merge all the same.

The snapshot refresh still looks like real pre-existing drift worth landing (the last one was #4552). When you do, the reproducibility problem is the thing to fix first: between two runs of functionally identical output, three pages exceeded the configured `maxDiffPixelRatio` and four more sat within 10% of it, so fresh baselines will not hold on their own.

PR #5623, opened 2026-09-15
docs: security matching rules

aj-emerich:
Stems from https://github.com/kestra-io/kestra/commit/7f34765b61625393f7c82d6ecebdc9a4965c1fb6

PR #5626, opened 2026-09-15
ci(cache): keep one incremental build entry per ref and drop a PR's on close

elevatebart:
### Why

Every deploy saves a new `astro-build-<variant>-<ref>-<sha>` entry and nothing removes the previous ones, so a ref with N pushes holds N entries until the 7-day expiry or an eviction under the repo's 10 GB quota. The eviction is what emptied main's entry on Sep 15: [#6371](https://github.com/kestra-io/docs/actions/runs/34940667915) found no `astro-build-production-main-` key although [#6335](https://github.com/kestra-io/docs/actions/runs/34849610432) had saved one the day before, and it rebuilt from scratch.

Sizes, from the post-job save lines: production entries on main are ~21 MB, preview entries ~22 MB since #5621 dropped the passthrough image cache before the save (22,102,340 and 22,025,763 bytes on the two #5624 runs). The ~694 MB entries under #5621's merge ref predate that commit's own fix and expire on their own after 7 days unused. The recurring big entries are setup-node's `~/.npm` cache at 331 MB per lockfile hash and, once #5624 lands, the node_modules entry at 289 MB, one live copy each.

### What

- `deploy.yml`: a last step deletes the older entries under the job's own key prefix, `astro-build-<variant>-<ref>-`, keeping the current commit's if a re-run already saved it. It runs on success only, so a failed job (which saves nothing) keeps its last good entry, and it is `continue-on-error` because a fork PR's token cannot delete caches. Needs `actions: write`, added at the workflow level next to `pull-requests: write`.
- New `cache-cleanup.yml`: on PR close, delete every entry under `refs/pull/N/merge`. Uses `pull_request_target` so the base repo's token handles fork PRs too; nothing from the PR is checked out or run.

Net effect: one ~22 MB entry per open PR and one per production ref, released when the PR closes, instead of one per push forever.

### Not in this PR

- `cache: npm` on `setup-node` in the deploy job (331 MB per lockfile hash) only helps the miss path once #5624's node_modules cache is in; dropping it is a quota-versus-one-slow-miss trade to make separately.
- A one-off purge of the legacy #5621 entries: `gh cache list --key astro-build-preview-5621/merge- --json id --jq '.[].id' | xargs -n1 gh cache delete`, if waiting a week is too long.

https://claude.ai/code/session_015nu7pdX9cohW17FCYWfRpM

---
_Generated by [Claude Code](https://claude.ai/code/session_015nu7pdX9cohW17FCYWfRpM)_


iitzIrFan (approved):
Approving. The design is sound and all 13 checks are green. Two notes, neither blocking, and one thing worth doing before merge.

First, a correction to my own note on #5624: you are right about the sizing and I was wrong. Breaking down the current 10.2 GB — only 10 of the 24 `astro-build` entries are large (692-694 MB, 6.9 GB together); the other 14 total ~0.3 GB at ~21 MB each. Every large one is a pre-#5621 run or from #5616, the "do not merge" test PR for this fix. #5623 makes the timeline plain: 692 MB at 07:15, 21 MB at 07:44, either side of #5621 landing at ~07:20. So the 6.9 GB self-resolves at the 7-day expiry and the steady state is ~3.4 GB. That reframes this as hygiene rather than relief — capping unbounded N-pushes-to-N-entries growth is still worth doing, just not urgent.

**The prune runs before the save, so a failed save leaves the ref with nothing.** `actions/cache@v4` saves in a post-job step (`post-if: "success()"` in its own `action.yml`, which also confirms your "a failed job saves nothing" reasoning). A normal step therefore always runs before it, so the order is: delete the old entry, then post-job writes the new one. `actions/cache` swallows save failures as warnings, and when that happens the ref ends up with **zero** entries and a fully cold next build — where today a failed save leaves the last good entry standing. #5624 already established the fix: split into `actions/cache/restore` plus an explicit `actions/cache/save`, then put the prune after the save.

**Please exercise the delete path once before merging.** This PR's deploy ran the step with `KEY_PREFIX=astro-build-preview-5626/merge-` and produced no output, because it was the first deploy on the ref and there was nothing to delete. Exactly one 21 MB entry exists for the ref now, which is the right end state but not proof the deletion works. One more push would give it a predecessor to remove. Worth doing because `continue-on-error: true` with no echo makes a broken delete look identical to that clean log — adding the `echo "deleted N"` you already have in `cache-cleanup.yml` would make a silent no-op visible.

Correct as written, for the record:

- **The `pull_request_target` is safe.** No checkout, no PR code executed, and the only PR-controlled value interpolated is `github.event.pull_request.number`, an integer, so there is no injection surface. Keep that comment prominent — the safety is contingent on nobody adding a checkout later.
- `gh cache list --key` is a prefix filter and the prefix embeds `github.ref_name`, so the prune is ref-scoped by construction.
- `select(.key != "<prefix><sha>")` preserves a re-run's saved entry, `xargs -r` handles empty, and the per-ref `concurrency` group with `cancel-in-progress: false` rules out two pushes racing the prune.

Nits:

1. `actions: write` is set at workflow level, so `truncate-url-slug` inherits it for a sparse checkout and a slug computation. Scope it to `deploy-workers`.
2. The lever you deferred is now the largest one: `node-cache` (setup-node's `~/.npm`) is 3 entries at ~310 MB, ~0.93 GB, roughly 3x the entire post-fix `astro-build` footprint, and after #5624 it only helps the miss path.

Independent of #5624 and touching different regions of `deploy.yml`, so these merge in either order.

elevatebart:
Both points taken, in 1a63de9:

- The restore is now `actions/cache/restore`, the save an explicit `actions/cache/save` on success, and the prune runs after it, so a swallowed save failure leaves the ref its previous entry instead of none. Your `post-if: "success()"` reading also confirms the "failed job saves nothing" half.
- The prune prints `deleted N superseded entries under <prefix>`, so a no-op is visible in the log.
- `actions: write` is scoped to `deploy-workers`; the workflow level is back to `pull-requests: write` only.

This push is also the delete-path exercise you asked for: the ref now has the 21 MB entry from the first deploy as a predecessor, so the prune on this run should report `deleted 1`. I will read that log line when the run lands and quote it here.

Thanks for the cache breakdown; 10 large legacy entries expiring on their own and a ~3.4 GB steady state is the picture I had from the save lines, good to have it confirmed from the listing. The `node-cache` lever (3 × 310 MB, miss-path only after #5624) is the next PR.

---

PR #5632, opened 2026-09-15
refactor(styles): remove bootstrap

elevatebart:
Nothing in `src` imports bootstrap any more, in SCSS or in JS, and
`bootstrap` and `@types/bootstrap` are gone from package.json.

### Gain

Measured against main at `bd614bf`, the base this branch was cut from; `main`
has since been merged in and the numbers are unaffected by it. "Page JS" is
every `.js` the browser actually fetches: the eager scripts, the island entries,
and their static import closures. `/about-us` is browser-measured (build served
locally, requests counted after network idle, 63 files); the routes that cannot
prerender in a sandbox are the same set reconstructed from the built chunks and
calibrated against that measurement.

| Page JS, gzipped | main | this PR | |
|---|---|---|---|
| `/about-us` | 363 KB | 339 KB | −6.6% |
| `/docs/<page>` | 376 KB | 352 KB | −6.4% |
| `/plugins` | 377 KB | 353 KB | −6.3% |
| `/plugins/<slug>` | 380 KB | 356 KB | −6.3% |

| Global stylesheet | main | this PR | |
|---|---|---|---|
| gzipped | 25,044 B | 8,126 B | −68% |
| minified | 181,479 B | 43,816 B | −76% |

The JS delta is the same 24,506 B gzipped (81,608 B raw) on every route, because
it is one global script: `scripts/bootstrap.js` imported Modal, Collapse and
Tooltip, Tooltip drags in popper, and the lot shipped whether or not anything
opened. Its replacement, `src/scripts/modal.ts`, is 392 B. With ~17 KB of CSS
that is **about 41 KB gzipped off every page**.

**Correction to what this description said earlier.** It claimed −33% page JS on
`/about-us`. That was wrong: it summed only the chunks named in the HTML and
ignored their static import closures, which are most of the payload. The real
denominator is ~340 KB gzipped on every route, because `Search` is `client:load`
and statically imports `marked-shiki` plus twenty-odd shiki grammars. Against
that, bootstrap is 6-7%. Worth having; not the headline, and that shiki closure
is the bigger fish — filed as #5636.

### Build time: no, and this is a correction

An earlier version of this description claimed 44.9 ms → 5.4 ms on the SCSS
prelude. That was measured against main *before* #5624, which has since landed
and does the same prelude split. #5624 owns that win. Re-measured against
current main:

| | main | this PR |
|---|---|---|
| Prelude eval, ×~370 per build | 3.25 ms | 3.42 ms |
| `vendor.scss` compile, once per Vite environment | 420.1 ms | 88.3 ms |

So this PR is ~5% *slower* on the per-block prelude — the extra
`@import "breakpoints"` resolution — and saves ~330 ms per Vite environment on
the one-time global-sheet compile. Call it a second off a build that runs two
to four minutes: not material, and not a reason to take this change. Main's own
deploy runs swing 2m22s–6m25s on cache state alone, so a single CI comparison
would be noise in either direction.

Take this for the bundle and for owning the CSS. Not for build time.

### Why

The website's own CSS had grown to the point where bootstrap was mostly being
fought rather than used: `app.scss` carried 38 selectors whose only job was to
out-rank a bootstrap component, several with `!important`. The removal is
mostly deleting the thing being fought.

Some numbers that make the case better than prose. Of bootstrap's 1570
generated utility classes the markup uses 144. `navbar` shipped
`navbar-expand-*` for six breakpoints plus offcanvas and a toggler icon, where
the header only ever uses `navbar-expand-xl`. `forms` shipped form-check,
form-switch, form-range, floating labels, file inputs and the whole validation
suite against five classes in use. `buttons` shipped sixteen colour variants
and their outline pairs; four are used, and two of those were already being
repainted from the `--ks-*` tokens.

### What

**JS.** Modal, Collapse, Tooltip and Dropdown were the only reason the bundle
loaded `bootstrap` at runtime.

- Modal → `<dialog>` + `showModal()`. `data-bs-toggle="modal"` triggers become
  `data-modal-target="#id"`, handled by one delegated listener. Escape, focus
  trap and background inerting now come from the platform; the backdrop, the
  fade and the body scroll lock are CSS.
- Collapse → a `0fr`/`1fr` single-row grid below `lg`, which animates height in
  CSS; above `lg` the wrappers are `display: contents` so they leave the box
  model alone. The mobile nav panel closes with `display: none` plus
  `transition-behavior: allow-discrete`, which keeps the fade without costing
  layout for its 268 descendants on every page.
- Dropdown → `useDropdown`, a composable over vueuse's `onClickOutside` plus
  Escape. The header's mobile submenus now reuse the `showMenu`/`showMenuId`
  state the desktop mega-menu already had.
- Tooltip → popper directly, built on first hover. Popper was already a
  dependency (`SchemaToCode.vue`), and it is no longer global.

**CSS.** Sixteen partials ported into `src/assets/styles/components/`, keeping
only the selectors that render.

`_utilities.scss` is a map of the 144 classes in use driving a small generator,
which is both the whole implementation and the only place to edit. A utility
not listed there does not exist. The grid deliberately goes the other way and
keeps all twelve columns at every breakpoint rather than the twenty-two in use,
because `col-md-3` is the class someone adds without thinking.

`root` shrinks from ~120 custom properties to the 30 this repo reads, and every
`--bs-*` name in `src` becomes `--ks-*`.

### Four ways this breaks silently

Every one of these compiles, type-checks and renders without complaint. They
are the reason this took as many passes as it did, and they are worth knowing
before anyone attempts something similar.

**An unknown Sass function is not an error.** `max-height: subtract(100vh, 9rem)`
used bootstrap's `subtract()`. With bootstrap's functions gone, Sass passes it
through as a CSS function, the declaration is invalid at parse time and is
dropped. The docs sidebar lost its max-height and its scroll containment with
it. Caught by the visual suite, not by anything else.

**Children read variables declared on the component root, with no fallback, and
that is load-bearing.** The header's mega-menu uses `.dropdown-item` outside any
`.dropdown-menu`; `VideoDialog.astro`, `TutorialsGrid.vue` and
`common/Modal.vue` use `.modal-header`/`.modal-body` outside any `.modal`. There
bootstrap's variables resolved to nothing and every declaration fell back to its
initial value. Adding fallbacks — the obvious thing when porting — paints
padding and borders onto markup that never had them. Three instances: dropdown,
card, modal.

**Replacing a `display` toggle changes what the browser skips.** Collapsing the
mobile nav with `visibility: hidden` instead of `display: none` keeps the fade
but costs layout for the panel's 268 descendants on every page at mobile width.
Wrapping a panel in a grid adds boxes and suppresses margin collapsing, shifting
the docs sidebar ~8px. Both were invisible in computed styles taken out of
context.

**normalize is not a drop-in for reboot.** modern-normalize owns the
cross-browser layer, but bootstrap's opinionated half has to stay — heading
rhythm, list indentation, `svg { vertical-align: middle }`, fieldset/legend, and
form controls inheriting the body's typography. Dropping it moved 400+
measurements. 1200-odd markdown files rest on `p { margin-bottom: 1rem }` and
`ol, ul { padding-left: 2rem }`.

(And a smaller one: a trailing `, none` in a `background-image` can be a slot,
not a no-op. `.form-select` declares `var(--bg-img), var(--bg-icon, none)`, and
the second layer is where `.is-invalid` puts its error icon.)

### Deliberately not changed

Bootstrap's light palette is carried over verbatim wherever the site never
re-pointed it: a `.card` with no background of its own still paints white, on a
dark site. Hardcoding those values makes them visible in the diff instead of
hidden behind an import. Retheming them is a visual decision for a separate PR.

`.alert` and its variants are left alone too. They appear across the contact
form, the newsletter boxes, the partner form and the plugin beta notice, but
`vendor.scss` never imported `bootstrap/scss/alert`, so bootstrap never styled
them here either. The only rule they have ever had is `app.scss`'s
`.alert > *:last-child { margin-bottom: 0 }`, and it is unchanged. Giving them a
box now would be a new design, not a port.

### Verification

Three independent passes, because the first two each missed things the next one
caught.

**Computed styles.** `harness/` mounts the components whose CSS this owns next
to a styleguide of every class and element it touches, and records ~30 computed
properties plus the box size for 752 of them at desktop and mobile widths. Two
snapshots and a diff, twenty seconds, no network:

```sh
npx vite build --config harness/vite.config.mjs
npx vite preview --config harness/vite.config.mjs &
node harness/snapshot.mjs /tmp/before.json
# change something
node harness/snapshot.mjs /tmp/after.json && node harness/diff.mjs /tmp/before.json /tmp/after.json
```

Ran between every partial removal; caught four regressions that reading the
diff had not.

**Class coverage.** Diffing every class main's CSS defines against this branch's,
intersected with what the markup uses, found five more dropped rules the harness
had no probe for — `.form-control.is-invalid` (the contact form lost its red
border), `.h1`-`.h6` as classes, `.modal-header .btn-close`, the modal variable
scoping, and the `, none` layer. All five now probed.

**The visual suite** (`Update Linux Visual Snapshots`, dispatched three times as
fixes landed). Found three more the other two could not see, because they need
a real page: the `subtract()` bug, the sidebar box-model shift, and the mobile
nav layout cost. See the [comment below](https://github.com/kestra-io/docs/pull/5632#issuecomment-5680090422)
for the detail.

Final state: every snapshot back to its pre-PR rendering, largest remaining
delta −17 B, no layout difference on any page. Lighthouse on the fixed head with
the most evenly-matched runners of four runs: Docs Landing 81 → 94, LCP 2.48s →
1.15s, no score regression anywhere. Computed-style diff against main is
colour-only, on elements that are hidden or repainted. `harness/behaviour.mjs`
31/31, `astro check` clean, `Build site` and `Publish to Cloudflare Workers`
green.

The harness is not part of the site build and ships nothing. If you would
rather not carry it, `git revert e0b8c96` drops it cleanly.

### Merging main in

`main` was merged at `90b1530`. One textual conflict, in
`MarkdownActionsMenu.vue`: main widened the `v-if` to `markdownBody ||
lazyMarkdown` while this branch had swapped bootstrap's Dropdown for
`useDropdown`, which wants `ref="root"` and `:class="{ open }"` and no longer
wants the `dropdown` class. Both sides kept, minus that class.

One silent conflict git could not see: `ChangelogChangeGroup.vue`, new on main,
reads `--bs-font-monospace`, which this branch renamed to `--ks-font-monospace`.
Same trap as above — an undefined custom property invalidates the declaration
rather than failing, so the changelog's monospace type would have quietly fallen
back to the body font.

Then `2837918` dropped four `--bs-*` aliases in `app.scss` that nothing reads
any more: not this repo's CSS, not the markdown content, and not the five vendor
stylesheets `vendor.scss` imports.

### Landing

`bootstrap` stays in node_modules regardless, because `@kestra-io/topology`
declares it as a peer dependency. Its bundle never imports it — grepped — so
that is a one-line fix in that repo, not this one.

### Review suggestion

The refactor commits are each self-contained and the messages carry the
reasoning, so commit-by-commit is likely easier than the combined diff. The
parts most worth a second pair of eyes are `components/_utilities.scss` (the
trim is a judgement call about what future pages may want) and
`components/_base.scss` (the reboot/normalize split).

https://claude.ai/code/session_01733YbjY3AViGH7htzr8STV


elevatebart:
Ran the visual suite against this branch (`Update Linux Visual Snapshots`, three times as fixes landed). It found three regressions the computed-style harness could not, all now fixed. Recording them because each is a way removing bootstrap breaks something no compiler or type checker can see.

**1. `subtract()` silently dropped the docs sidebar's max-height.** `max-height: subtract(100vh, 9rem)` used bootstrap's Sass `subtract()`. With bootstrap's functions out of the prelude, Sass does not error on an unknown function — it passes it through as a CSS function, and the declaration is invalid at parse time and dropped. The sidebar lost its scroll containment and expanded to the full nav height, pushing the page and footer down: `/docs/contribute-to-kestra` grew 1552px -> 1843px. Fixed with `calc(100vh - 9rem)`, which is what `subtract()` compiled to.

**2. The collapse rewrite perturbed the sidebar's box model.** Replacing `.collapse` with a `0fr`/`1fr` grid adds a wrapper and an inner div, and `display: grid` suppresses the margin collapse the sidebar's spacing relied on — the sidebar sat ~8px low and clipped at the wrong point. Above `lg` the panel is always open and never animates, so both wrappers are now `display: contents` and leave the box model untouched. The grid stays below `lg`, where the transition runs.

**3. `visibility: hidden` on the mobile nav cost layout on every page.** I collapsed the panel that way so the fade could run. It does not skip layout, so the browser laid out the panel's 268 descendants at mobile width on every page — work `.collapse`'s `display: none` skipped. Now `display: none` plus `transition-behavior: allow-discrete`, which gets both. Descendants with layout boxes: 268 -> 0.

### Result

All snapshots are back to their pre-PR renderings. Every remaining difference is noise:

| Snapshot | vs pre-PR |
|---|---|
| Contribute to Kestra (desktop) | −17 B (0%) |
| Docs Landing (desktop) | −2 B (0%) |
| Get Started (desktop) | 1px taller, −1.7% bytes (animation frame) |
| Get Started mobile/tablet, Home mobile/tablet | +24 to −315 B |
| Cloud, Enterprise (desktop) | −1.6%/−2.1%, layout pixel-identical (logo carousel frame) |

No layout difference on any page.

Lighthouse on the fixed head, on the most evenly-matched runners of the four runs so far (shard CPU 2494/2489/2532/2477 against a 2499 baseline): Docs Landing performance 81 -> 94 and LCP 2.48s -> 1.15s, no score regression on any page, and TBT improved on Docs Landing, Pricing and Blog Index. Two small TBT flags remain (Flow 96ms, Plugins Landing 85ms) in a metric that has read 90 -> 144 -> 85 for Plugins Landing across runs of near-identical code.

### Worth knowing for whoever reviews

The suite is `workflow_dispatch`-only, so none of this runs on PRs. A change of this shape is exactly what it is for, and it earned its keep here — the harness in `harness/` agreed with main on all 752 probes while three of these were live, because it had no probe rendering a sticky sidebar, a viewport-height constraint, or a real docs page.

---

elevatebart:
Correcting the JS figures in the description above. Both were wrong, in the flattering direction.

**The size was inflated.** 102,721 B came from a vite lib build that did not actually minify — it came out larger than bootstrap's entire minified dist (80,496 B for `bootstrap.bundle.min.js`), which should have been the tell. Re-measured with esbuild, `import { Collapse, Modal, Tooltip } from "bootstrap"` bundles to **81,608 B minified, 24,506 B gzipped**. That is close to the full bundle because Tooltip pulls in popper.

**The percentage used the wrong denominator.** "−99.6%" was the global script file going 81,608 B → 392 B. True of that file, and misleading as a headline: it reads as "99% less JavaScript on a page", which is not what happens. Against the JS a page actually loads, transitive closure from the built output:

| `/about-us` | main | this PR | |
|---|---|---|---|
| Page JS, raw | 219,412 B | 137,804 B | −37% |
| Page JS, gzipped | 73,276 B | 48,770 B | −33% |

About a third of page JS. The share varies by route — plugins and blueprints pages carry much more island JS, so the same 24.5 KB is a smaller slice there; thin pages, larger.

Combined with the stylesheet (25,044 → 8,126 B gzipped), a page like `/about-us` drops roughly **41 KB gzipped**, of which ~24.5 KB is JS and ~17 KB CSS.

The stylesheet figures in the description are unaffected — those were measured by compiling `vendor.scss` directly and are unchanged.

---

elevatebart:
### Per-route JS, and a correction to the earlier numbers

The `−33% page JS` in the first version of the description was wrong, and wrong in the flattering direction again. It summed the chunks named in `about-us/index.html` (13 files, 137,804 B) and ignored their static import closures. A browser fetches the closure, not the entry list.

Measured properly: serve `dist/client` locally, load `/about-us` headless, count every `.js` response after network idle. 63 files, 1,250,221 B raw, 346,976 B gzipped. Reconstructing the same set from the built chunks by walking static imports gives 66 files / 1,234,476 B, so the reconstruction is within ~1% and can stand in for the routes that cannot prerender in this sandbox (`api.kestra.io` is blocked here, so only `/1-0` and `/about-us` build).

| gzipped | main | this PR | delta |
|---|---|---|---|
| `/about-us` (browser-measured) | 363 KB | 339 KB | −6.6% |
| `/docs/<page>` | 376 KB | 352 KB | −6.4% |
| `/plugins` | 377 KB | 353 KB | −6.3% |
| `/plugins/<slug>` | 380 KB | 356 KB | −6.3% |

Islands counted per route: docs = `NavSideBar` + `PrevNext` + `NavToc`; `/plugins` = `PluginsLists` + `Faq`; `/plugins/<slug>` = `PluginSidebar` + `RightSideBar`. `HelpfulVote` is `client:visible` below the fold and adds ~3 KB once scrolled.

So the answer to "what's the delta on plugins and docs pages" is: the same 24,506 B gzipped as everywhere else. It is one global script, so the absolute number never moves; only the denominator does, and those routes carry 60 KB gzipped of extra island JS, which is why the percentage is slightly smaller than on `/about-us`.

### The thing worth taking away

Every route ships ~340 KB of gzipped JS before any of this. `Search` is `client:load` in `layout.astro` and statically imports `MarkdownRenderer` → `marked-shiki` → twenty-odd shiki grammars: `module.4njZ_gBS.js` alone is 277 KB raw, `javascript.BipqpuGD.js` 175 KB, plus python, java, sql, groovy, shellscript and friends. That is a syntax highlighter for search results, downloaded on every page whether or not anyone opens search.

Making that closure lazy is worth several times what removing bootstrap is. Out of scope here; worth its own issue.

---

elevatebart:
Filed the Shiki closure as [#5636](https://github.com/kestra-io/docs/issues/5636), with the requirement that code stays readable in plain form while the highlighter loads.

One correction to the comment above: I named `module.4njZ_gBS.js` (277 KB raw) as part of the Shiki closure. It is not. It is posthog-js, loaded eagerly from `layout.astro`'s first inline script, and it belongs in its own bucket. The Shiki closure is 20 chunks, 664,687 B raw and 136,386 B gzipped, which is 53% of raw page JS and 39% gzipped. The point stands, on smaller numbers than I gave it.

---

src/assets/styles/components/_utilities.scss
```diff
@@ -0,0 +1,314 @@
...
+            danger: #dc3545,
+            white: $white,
+            muted: rgba(33, 37, 41, 0.75),
+            reset: inherit,
+        ),
+    ),
+    "background-color": (
+        property: background-color,
+        prefix: "bg",
+        values: (light: #f8f9fa, white: $white, transparent: transparent),
```
Piyush-r-bhaskar, line 248: This drops `bg-primary`/`bg-secondary`/`bg-success`/`bg-info`, but two render paths still emit them for edition badges:

- `src/markdown/remark/remark-custom-elements/badge.mjs:61` — `` `badge d-flex align-items-center bg-${info.color}` ``
- `src/utils/renderVersionedDoc.ts:90` — same string

`info.color` comes from `badgeMaps.mjs` and is one of those four. Nothing defines them any more (checked the compiled `vendor.scss` + `app.scss` and every component sheet). With `_badge.scss` hardcoding `color: $white` and `markdown.scss` losing its `--bs-badge-color` override, these render as white text on no background — invisible in light mode.

Live on six `:::badge{...}` directives today (docs-contributor-guide ×2, workflow-components/outputs, concepts/backfill, how-to-guides/python-dependencies, use-cases/approval-processes) plus whatever `renderVersionedDoc` serves.

I'd give `_badge.scss` its own `.badge.bg-*` rules against `--ks-*` tokens rather than restoring bootstrap's `#0d6efd` here — these are the only consumers.

src/assets/styles/components/_utilities.scss
```diff
@@ -0,0 +1,314 @@
...
+// lists those 144 and nothing else, which is the whole point: it is the one
+// place to look, and the one place to edit. Adding a class means adding its
+// value to `values` (and its breakpoint to `responsive`) here — a utility that
+// is not listed does not exist, and will simply not apply.
+//
+// `responsive: (md lg)` emits `<prefix>-md-<key>` and `<prefix>-lg-<key>` in
+// those breakpoints' min-width blocks, mirroring bootstrap's infix order so
+// the cascade is unchanged.
+
+$ks-utility-spacers: (
```
Piyush-r-bhaskar: `$ks-utility-spacers` is never referenced — every spacing utility inlines its own `values` map. Given this file's framing as the one place to look and edit, a decoy map of spacer values is the worst thing to leave in it. Wire the spacing utilities through it or drop it.

src/components/docs/NavSideBar.vue
```diff
@@ -209,19 +205,42 @@
                 }
             }
         }
+        // Height transition without JS: 0fr to 1fr on a single-row grid.
         .bd-menu-collapse {
+            display: grid;
+            grid-template-rows: 0fr;
+            transition: grid-template-rows 0.25s ease;
+
+            .bd-menu-collapse-inner {
```
Piyush-r-bhaskar, line 223: `overflow: hidden` at `0fr` clips the panel but doesn't take it out of the focus order or the a11y tree — bootstrap's `.collapse` was `display: none`. On mobile, tabbing past the toggle now walks every link in the collapsed sidebar, scrolling the clipped container as it goes.

Same in `NavToc.vue:408` and `BlogToc.vue:155`. You solved it correctly for the nav panel in `Header.vue` (`display: none` + `allow-discrete`); here `:inert="!menuExpanded"` on the inner wrapper would fix it and keep the height animation.

`harness/behaviour.mjs` checks height and `aria-expanded` but not focusability, which is why it passed.

src/components/docs/NavSideBar.vue
```diff
@@ -209,19 +205,42 @@
...
             @include media-breakpoint-down(lg) {
                 nav {
                     padding: calc($spacer / 2) $spacer;
                     border: $block-border;
                     box-shadow: $box-shadow-sm;
-                    @include border-radius(var(--bs-border-radius));
+                    border-radius: $border-radius;
                 }
             }
             @include media-breakpoint-up(lg) {
```
Piyush-r-bhaskar, line 240: Braces balance, but this block and its closer sit at the wrong depth, which is what makes it read as unbalanced. `NavToc.vue` and `BlogToc.vue` got this right.

src/components/plugins/schema/Tooltip.vue
```diff
@@ -1,61 +1,74 @@
...
+        v-bind="$attrs"
+        @mouseenter="show"
+        @mouseleave="hide"
+        @focusin="show"
+        @focusout="hide"
+    >
         <slot name="default" />
     </span>
-    <span class="d-none" ref="$tooltipContent">
+    <span v-show="visible" ref="content" class="ks-tooltip" role="tooltip">
```
Piyush-r-bhaskar: Bootstrap wired `aria-describedby` from trigger to tip on show; `role="tooltip"` with nothing pointing at it gives screen readers nothing. Give this an `id` and bind `:aria-describedby="visible ? id : undefined"` on the trigger span.

src/components/plugins/schema/Tooltip.vue
```diff
@@ -1,61 +1,74 @@
...
+    }
+
+    const hide = () => {
+        visible.value = false
+    }
 
-    onBeforeUnmount(async () => {
-        tooltip.value?.dispose();
-    })
+    onBeforeUnmount(() => popper?.destroy())
```
Piyush-r-bhaskar, line 76: `popper ??=` builds once and only destroys on unmount, and popper's `eventListeners` modifier is on by default — so every tooltip a user hovers leaves permanent scroll/resize listeners on its scroll parents. On a schema page with hundreds of these, scrolling gets steadily more expensive the more you hover; bootstrap destroyed the popper on hide.

Cheapest fix is toggling the modifier in `show`/`hide`:

```js
popper?.setOptions({ modifiers: [{ name: "eventListeners", enabled: false }] })
```

src/assets/styles/components/_root.scss
```diff
@@ -0,0 +1,36 @@
...
+    --ks-border-radius-lg: #{$border-radius-lg};
+    --ks-gray-300: #dee2e6;
+    --ks-gray-400: #ced4da;
+    --ks-gray-600: #6c757d;
+    --ks-secondary: #6c757d;
+    --ks-pink: #d63384;
+    --ks-primary: var(--ks-content-color-highlight);
+    --ks-link-color: var(--ks-content-link);
+    --ks-link-hover-color: var(--ks-content-color-highlight);
+    --ks-link-color-rgb: var(--ks-content-link);
```
Piyush-r-bhaskar: Declared again with the same value at `app.scss:55`.

Separately: `_base.scss:149` carries `color: rgba(var(--ks-link-color-rgb), var(--ks-link-opacity, 1))` forward, and this holds a hex rather than a triplet — so that declaration has always been invalid and dropped. Inherited from main, not introduced here, but a port whose rule is "keep only the selectors that render" is the moment to collapse it to `color: var(--ks-link-color)` and delete both `-rgb` vars.

Piyush-r-bhaskar (changes requested):
Went through this by compiling both stylesheets and diffing the selector sets rather than reading the trim. 2134 classes before, 546 after; of the 1624 dropped, only `nav-item`, `card-link` and `card-group` appear in markup and all three are safe (the `.nav-item` rules all needed a `.nav-tabs`/`.nav-fill` ancestor, the other two are defined in `app.scss` and `Stacked.astro`). Element selectors match apart from `b, strong, sub, sup`, which modern-normalize covers. No markdown content carries HTML classes, so that risk isn't there either. Nice work, and the corrections in the description are the right way to handle it.

Two things to fix before merge, both in paths no test covers:

- `bg-*` colour utilities are gone but still generated at render time for edition badges → invisible badges on six docs pages.
- The `0fr` collapses keep their contents in the tab order, where bootstrap's `.collapse` was `display: none`.

Rest is small. Happy to approve once those two land.

src/assets/styles/components/_utilities.scss
elevatebart: Fixed in `b5c61b7`, and thank you — this is a genuine miss. My class-coverage pass compared what the CSS defines against what the markup uses, and these classes are built at render time from `badgeMaps.mjs` rather than written in any template, so they were invisible to it.

Took your suggestion: `.badge.bg-*` for exactly the four values in that map, scoped to `.badge` since those two call sites are the only consumers, and paired with content tokens so both themes stay legible rather than relying on `.badge`'s hardcoded white:

| class | background | color |
|---|---|---|
| `bg-primary` | `--ks-background-button-primary` | `#fff` |
| `bg-secondary` | `--ks-background-secondary-inverse` | `--ks-content-secondary-inverse` |
| `bg-success` | `--ks-background-alert-success` | `--ks-content-alert-success` |
| `bg-info` | `--ks-background-alert-info` | `--ks-content-alert-info` |

The three pairs are ones the theme already defines together, so they invert correctly. `bg-primary` keeps white text because `--ks-background-button-primary` is `#631bff` in both themes.

Worth flagging that this is a colour change, not a restoration — main painted bootstrap's `#0d6efd` here. You asked for `--ks-*` over the hex and I agree, but it means the six `:::badge` pages will show a snapshot diff, and that diff is expected rather than a regression.

---

src/components/docs/NavSideBar.vue
elevatebart: Fixed in `b5c61b7` across all three. You're right that the height check was measuring the wrong thing.

One adjustment to the suggested binding: `:inert="!menuExpanded"` would have taken the **desktop** sidebar out of the a11y tree. Above `lg` the panel is always shown via `display: contents`, and `menuExpanded` stays `false` there — it only ever flips on mobile (`closeSidebar` even resets it below 992px). So the flag alone says "collapsed" on a desktop panel that is fully visible.

Gated on the viewport instead, at the same boundary as the SCSS:

```ts
const belowLg = useMediaQuery("(max-width: 991.98px)")
const collapsed = computed(() => belowLg.value && !menuExpanded.value)
```

`NavToc` and `BlogToc` have the same shape and got the same treatment.

On the harness gap: added a focusable-descendant count outside `[inert]`, asserted both collapsed and expanded. Checked it actually bites — reverting just the `inert` binding turns it red:

```
PASS  docs menu starts collapsed  :: h=0
FAIL  docs menu collapsed is not tabbable  :: focusable=3
```

35/35 with the fix in place.

---

src/components/plugins/schema/Tooltip.vue
elevatebart: Fixed in `b5c61b7`. Correct diagnosis — the listeners outlive the tooltip, and a schema page is the worst place for that.

Took the modifier toggle, with one wrinkle worth noting: `setOptions` merges the options object shallowly, so passing `modifiers` alone would have dropped the `offset` modifier and shifted every tooltip back onto its trigger. Both are restated each call:

```ts
const options = (listeners: boolean) => ({
    placement: props.placement,
    modifiers: [
        { name: "offset", options: { offset: [0, 6] } },
        { name: "eventListeners", enabled: listeners },
    ],
})
```

`show` applies `options(true)` (also on re-show, since the instance is reused), `hide` applies `options(false)`.

---

src/components/plugins/schema/Tooltip.vue
elevatebart: Fixed in `b5c61b7`. `useId()` for the id, bound only while visible so a hidden tip is never referenced:

```vue
<span ref="trigger" :aria-describedby="visible ? tooltipId : undefined" …>
<span v-show="visible" :id="tooltipId" role="tooltip" …>
```

---

src/assets/styles/components/_root.scss
elevatebart: Both gone in `b5c61b7`, and you're right about the rgba form — the var holds a hex, so that declaration has never applied.

One deliberate divergence: I deleted the declaration rather than collapsing it to `color: var(--ks-link-color)`. Repairing it would make it *newly* apply, giving every anchor the rule reaches a colour it has never had on main — anchors currently inheriting, or picking one up from a later rule, would change. That is a redesign hiding inside a cleanup, and not one this PR should smuggle in. So `a` keeps only `text-decoration: none`, which is the half that was rendering, and both `-rgb` vars plus the `app.scss` duplicate are deleted.

The compiled selector diff for the whole commit is the four badge rules added and `a:hover` gone, nothing else.

Happy to re-point links properly in a follow-up if that is the intent — it just wants a visual run behind it rather than riding along here.

---

src/assets/styles/components/_utilities.scss
elevatebart: Dropped in `b5c61b7`. Fair hit, and the framing argument is the right one — a file that claims to be the single place to edit cannot carry a map that edits nothing.

Chose deleting over wiring: routing the spacing utilities through it means each one lists keys and resolves values through the map, which splits "what exists" from "what it is worth" across two places and changes generated output for no visual gain. Deleting leaves each utility's values next to its prefix, which is what the file already does everywhere else.

---

src/components/docs/NavSideBar.vue
elevatebart: Re-indented in `b5c61b7` to match `NavToc` and `BlogToc`.

---

Piyush-r-bhaskar (approved):

elevatebart:
Thanks for the approval. **Please hold the merge for one more commit** — the branch currently carries 35 wrong snapshot baselines, and that is my doing.

### What happened

I refreshed the visual snapshots after your review. The run went green and rewrote 35 use-case baselines (`a460850`). Green meant nothing: `--update-snapshots=changed` rewrites whatever fails, so it recorded a broken rendering as the new truth.

Diffing those PNGs pixel-wise rather than trusting the tick: 823,710 differing pixels on `Use-Case-Modern-Data-Stack-desktop`, 12.7% of the image, max channel delta 255, in contiguous bands rather than scattered. Every alternating feature section had flipped left-to-right.

`features/shared/Alternation.astro` reverses each odd row with `order-md-0` / `order-md-1`. The utilities map declared `order` as `responsive: (lg xl)`, so the md variants were never generated and both columns fell back to the base `order-0` / `order-1` — image first, text second, at every width, on every use-case page. `ms-sm-3` in `FooterContact.astro` was missing the same way (`margin-start` had only `xl`).

Same failure mode as the badges you caught: a class the map does not cover does not error, it just does not apply.

Byte sizes are why it hides. Mirroring a layout preserves nearly every pixel, so the PNGs compress to roughly the same size — that page moved 2.6%, several moved *down*, and GitHub's binary view shows only the byte count.

### Fixed, and made checkable

`baa87a4` adds the two breakpoints. It also adds `harness/coverage.mjs`, which diffs every bootstrap-shaped class in the markup against the CSS this repo generates and exits non-zero on anything undefined. Against the previous commit it reports exactly these three; against the fix, clean. Ten remaining hits are v4 leftovers and typos (`form-group`, `col-md-13`, `text-left` and friends) that bootstrap 5.3.8 never defined either — verified against the package, listed separately and ignored so nobody re-litigates them.

A second refresh run is in flight to put the 35 baselines back. I will diff that one the same way and confirm here before this is mergeable.

### Still open

The `_root.scss` thread, where I removed the invalid `rgba(var(--ks-link-color-rgb), …)` rather than repairing it — repairing changes what every anchor renders, so that is your call rather than mine to slip in.

---

elevatebart:
Second refresh landed as `6fffb65`. The flip is gone and the branch is mergeable from my side.

### The fix worked

Measured against main's baselines, not against the run's exit code:

| page | before `baa87a4` | after |
|---|---|---|
| `Use-Case-Modern-Data-Stack-desktop` | 12.68% of pixels differ | 0.64% |
| `Use-Case-Provisioning-desktop` | 9.03% | 0.56% |

### What still differs from main, and why none of it is this PR

Thirty-nine use-case snapshots still differ. I diffed every one of them pixel-wise; they fall into three shapes.

**Main's baseline caught an illustration before it loaded.** `Use-Case-Databases-tablet` is the clearest: main's PNG has the purple glow and no icons, ours has the icons. That is a flaky baseline on main's side from when #5673 added these 42 routes, and our render is the correct one. This accounts for the high max-deltas (166-238) on the handful of pages that show them.

**Gradient dithering.** Soft blur edges on the section backgrounds, max channel delta 27 of 255, mostly under 16. Invisible.

**A 1px vertical offset partway down seven pages.** On `Use-Case-Automotive-desktop` the two are identical to row 6510 and then everything below sits exactly 1px lower, making the page 9763 → 9764. Content matches; one element above that point computes 1px taller. Same shape as the `Get Started` 1px noted in the earlier verification round.

Caveat on method: I confirmed the illustration case and the offset case on one page each and classified the rest by signature (max-delta profile plus the height pattern), rather than eyeballing all thirty-nine.

The remaining eight non-use-case snapshots (`Get-Started` ×3, `Home` ×2, `Enterprise`, `Docs-Landing`, `Contribute-to-Kestra`) are the deltas already documented [in the earlier comment](https://github.com/kestra-io/docs/pull/5632#issuecomment-5680090422).

### Also cleared

The `About Us` Lighthouse flag (−14) is gone. That run compared against an 18 Sep main baseline; the 08:06 run against a same-day baseline puts About Us at 85 with no flag, and adds `Plugins Landing` +11 and `Plugin Debug Page` +28. So it was baseline staleness, not a regression — I had flagged it as probably structural, and that was wrong.

Still open: the `_root.scss` thread on whether to repair the link colour rather than delete it. That one is yours.

---

PR #5637, opened 2026-09-15
fix(build): retry transient upstream fetches and stop decorative data from failing the build

### LINKED ISSUE #5631: Build-time fetches have no retry, and one unguarded throw aborts the whole build
## What happens

A single transient response from an upstream API fails the entire Deploy build.

Most recent instance: [run 34957626620](https://github.com/kestra-io/docs/actions/runs/34957626620) on #5446, 2026-09-15 10:25 UTC.

```
10:25:00 [ERROR] Error: Fetch error: 504  on url https://api.github.com/repos/kestra-io/kestra
10:25:00 [ERROR] [build] Caught error rendering /api/github: …
##[error]Process completed with exit code 1.
```

The same commit's `Build site` job in the Lighthouse workflow passed minutes earlier, and a plain re-run of the failed job went green with no code change — so this was purely GitHub's API blipping, nothing to do with the branch.

## Why one 504 is fatal

**No retry anywhere in the fetch layer.** `src/utils/fetch.ts` has zero retry or backoff: `internalFetch` throws on any non-2xx. 159 `$fetchCached` / `$fetchApiCached` call sites sit on top of it, reaching `api.kestra.io`, `api.github.com`, `raw.githubusercontent.com` and `registry.terraform.io`. The build memo at `fetch.ts:28` avoids *replaying* a transient failure across pages, but there is nothing to retry with.

**An inconsistent guard in the route that tripped.** `src/pages/api/github.ts` treats its two upstream calls differently:

- contributors — wrapped (`try` at `:20`, `catch` at `:25` → `defaultValues`, plus a 404 guard at `:32`)
- repo metadata at `:55` — no guard at all, throws straight through

`/api/github` is prerendered, so that throw aborts the run. It exists to feed `GithubStargazer.client.vue:15` — a star-count badge, fetched client-side. A decorative counter's upstream blip discards a complete build.

`DISABLE_GITHUB` would short-circuit it (`github.ts:16`) but is only set in `update-linux-snapshots.yml:65`, never in Deploy.

## Blast radius

Rare but total. 12 of the last 100 Deploy runs failed; exactly **1** was a fetch error. When it does hit, it discards a build of ~2,500 routes (production) or ~4,200 (preview) and reds the PR.

Long-lived PRs are the worst case — #5446 has run Deploy 16 times since 2026-08-30, so it rolls the dice ~8x more often than a same-day PR. The incremental cache makes this sting more, not less: a discarded build is cheap to redo only if the cache survived, and PR cache entries are currently being evicted under the 10 GB quota anyway (see below).

## Proposed fix

1. Bounded retry with backoff on 5xx/429 in `internalFetch`, GET/HEAD only so a form POST is never replayed. Covers all 159 call sites.
2. Guard `github.ts:55` the way `:20-27` already is — degrade to `defaultValues`, keeping the contributor count that did resolve.

Optionally: audit the other prerendered routes that fetch externally for the same unguarded pattern.

## Adjacent observation (separate from the above)

While tracing this, the Actions cache is at **8.86 GB / 10 GB**, with 5.5 GB in 8 stale `astro-build-preview-*` entries of ~694 MB each (PRs #5616, #5621, #5622, #5623) that predate the "Drop the passthrough image cache before it is saved" step. Every entry saved since is 21 MB. This already evicted #5625's own PR cache mid-PR, forcing a full rebuild. They should clear as those PRs push or close, but it is worth knowing about.

## Related

- #5387, #5392 (via #5446)
- #5624, #5626, #5621 — build/cache work in the same area

elevatebart:
Closes #5631.

One 504 from `api.github.com` discarded a 2,500-route Deploy build ([run 34957626620](https://github.com/kestra-io/docs/actions/runs/34957626620)). Two causes, both fixed here.

### 1. Retry in `internalFetch`

`src/utils/fetch.ts` had no retry, so every one of the ~159 `$fetchCached` / `$fetchApiCached` call sites was one blip away from a red build.

- 3 attempts (1 + 2 retries) on `408/425/429/500/502/503/504` and network errors
- GET/HEAD with no body only, so a form POST is never replayed
- backoff 250ms then 500ms, overridden by `Retry-After` when the upstream sends one, capped at 5s
- `4xx` still fails on the first attempt, and the existing error shape reaching callers is unchanged

Three rather than four because each extra attempt costs real wall-clock against a slow upstream, and the two statuses that actually rate-limit us send `Retry-After` anyway.

The per-build memo is unaffected: retries happen inside the memoized load, so the memo caches the settled outcome and still evicts on failure.

### 2. The unguarded prerendered routes

`/api/github`'s repo-metadata call threw straight through a prerendered route to feed a star-count badge. It now degrades to `defaultValues` while keeping the contributor count that did resolve, matching the guard its contributors call already had.

Then the sweep the issue asked for, applying one rule: **decorative sections degrade, route generation does not.**

Degraded (`$fetchApiCachedOptional`, empty fallback):
- blueprint carousels on the three use-case pages, `data`, `infra-automation`, `ai-automation`, `get-started` and the homepage
- blueprint tag lookups, related-blueprints on plugin pages, and the plugin logo walls in `PluginsBox` / `PluginsBlock`
- `BlueprintsGrid` hides itself with `:not(:has(.card-wrapper))` when it has no cards, so a degraded build ships nothing rather than a heading over an empty band

Deduplicated: seven sites re-rounded `/plugins/subgroups` by hand, unguarded, while `fetchTotalPluginsCount()` next door did the same thing with a guard. They all call it now. It returns a number instead of a string; all 20-odd consumers interpolate it into markup, so rendered output is identical, and `getSections()` already wanted a number. `home/Blueprints.astro` was on the uncached `$fetchApi`, so it also picks up the per-build memo.

Left alone deliberately: `getStaticPaths` and the doc, plugin and blueprint loaders. Degrading those would silently ship a site missing thousands of routes, which only beats a red build if nobody is looking. Every remaining build-time fetch util (`docSectionsFetch`, `catalog`, `fetchPluginPageData`, `taskUrls`, …) already has its own catch.

### Verification

- `337 tests passing` (+7 new, 1 rewritten). New `transient failures` block in `fetch.test.ts` covers recovery, exhaustion, network errors, the 4xx non-retry, the POST non-replay and `Retry-After` timing, all on fake timers so nothing actually sleeps. New `src/pages/api/_github.test.ts` covers the degraded path. Underscore-prefixed because anything else under `src/pages` becomes a route: the first push had it as `github.test.ts`, Astro built `/api/github.test`, and prerendering imported vitest and died. Caught by CI, fixed in 65fd884.
- The pre-existing memo test used a rejected promise for its first call, which retry would have swallowed. It uses a 404 now and still proves the memo does not replay a failure.
- `astro check`: 0 errors. `oxlint`: clean. It caught the one real break, `vs/[slug].astro` passing the count into `String.replace`.
- A full `astro build` was not runnable locally: the sandbox proxy blocks `api.kestra.io`, so every build-time fetch 403s at CONNECT. CI on `65fd884` is green across all 12 checks, and that build exercised the degrade path for real, GitHub rate-limited the runner with a 403 on the contributors call and `/api/github` rendered anyway rather than taking the build down.

Worth a reviewer's eye, since I could not open the preview from the sandbox: the blueprint carousels and plugin counts on `/get-started`, `/data`, `/overview` and the use-case pages should look exactly as they do on `main`. The `:has()` rule is inert while cards exist, but that is the one bit only a browser can confirm.

The 8.86 GB Actions cache noted at the bottom of the issue is untouched, it is a separate concern.

https://claude.ai/code/session_017uV66qGycp8aL6SXsUqyjV


iitzIrFan (changes requested):
The retry and the `/api/github` guard both do what they say, and I verified the part you couldn't from the sandbox: pointing a dev server at a dead `API_URL`, every converted page renders 200 instead of throwing, `/data` ships zero `card-wrapper` divs, and the compiled rule is `.blueprints-carousel[data-astro-cid-…]:not(:has(.card-wrapper))` — Astro leaves the inside of `:has()` unscoped, so it matches this component's own cards and the band does disappear. Against the live API the carousels and `2000+` counts on `/get-started`, `/data`, `/overview`, `/pricing` and `/use-cases/data-engineers` are identical to production. 337 tests and oxlint reproduce clean locally.

One gap in the sweep: `src/pages/use-cases/index.astro:18-24` still uses a bare `fetch()`, bypassing `internalFetch` entirely — no retry, no memo, no guard — and it feeds a `BlueprintsGrid`, the same decorative category as everything else you converted. In the dead-API run it was the only page to 500 while its siblings all returned 200, so a blip there still takes the build down. Could you move it onto `$fetchApiCachedOptional` like the sibling pages?

Also worth coordinating before either lands: #5446 declares `RETRY_BASE_DELAY_MS` at module scope in the same file. The two merge **without a conflict marker** and then fail to compile — merging them in a throwaway worktree, `src/utils/fetch.ts` auto-merged and oxlint reported `Identifier RETRY_BASE_DELAY_MS has already been declared` (lines 40 and 192). `pluginCount.ts` and `vs/[slug].astro` conflict loudly so those get caught; `fetch.ts` is the silent one. Whichever goes second should also drop `$fetchApiCachedWithRetry` (otherwise retries nest 3×3 = 9 attempts per dead URL) and reconcile `fetchTotalPluginsCount`'s return type — `number` here vs a formatted string there.

Non-blocking, happy to take these as follow-ups:

- A degraded build renders `0+ Plugins` / `0+ integrations` as visible body copy on `/`, `/pricing`, `/data` and `/get-started`. Pre-existing in `fetchTotalPluginsCount`, but seven sites that used to hard-fail now reach it — a floor constant would read better than `0+`.
- `home/Blueprints.astro` has no equivalent of the `:has()` guard, so a degraded homepage shows `0+ Blueprints, Ready to Build.` over an empty carousel.
- `Retry-After` sent as an HTTP-date parses to `NaN` and silently falls back to exponential backoff. Fine in practice, just noting it.
- No `AbortSignal`/timeout, so a *hung* upstream still stalls the build — retry doesn't cover that case.

elevatebart:
Thanks for running the dead-API check, that closed the one gap I flagged and couldn't test. The compiled-rule detail is the useful part: I'd written `:not(:has(.card-wrapper))` assuming the inside of `:has()` stays unscoped but had no way to confirm it, and it's what the second fix below relies on.

All three code points are in `e70c2db`, CI green across 12 checks.

### `use-cases/index.astro` — fixed

You're right, and the miss is embarrassing in a specific way: the sweep was driven by grepping `$fetch`, which cannot see a bare `fetch()`. Now on `$fetchApiCachedOptional` with the same `?? []` fallbacks as its siblings.

I re-swept for `(^|[^.$\w])fetch\(` afterwards. It was the only unguarded build-time one. The rest already degrade or aren't build-time:

- `home/PlatformOverview.astro:58` catches per icon and returns `null`
- `plugins/fetchPluginPageData.ts:160` and `plugins/repoReleases.ts:21` keep stale data on failure
- `hubspot-submit.ts`, `trackingProxy.ts` and the `plugins/[...slug].astro` poll are runtime, not prerender

Those three build-time ones still bypass `internalFetch`, so they get no retry. Worth a follow-up, but they degrade today so they can't red a build.

### `home/Blueprints.astro` — fixed

`section:not(:has(.blueprint))`, targeting the card class since the cards come from a child component. Works for the same unscoping reason you found. The `0+ Blueprints, Ready to Build.` heading goes with the band, so that page's half of the `0+` problem is gone too.

### `Retry-After` as HTTP-date — fixed

Both forms honoured now, and a date in the past falls back to backoff rather than resolving to a negative delay. Two tests added, 339 passing.

### Not taking, and why

**The `0+` floor.** Agreed it reads badly, but picking the number is a content decision I shouldn't make unilaterally, and it's visible marketing copy on `/` and `/pricing`. Happy to wire up a `PLUGIN_COUNT_FLOOR` if someone names the value.

**`AbortSignal`/timeout.** Correct that retry doesn't cover a hung upstream, and it's the more dangerous failure of the two since it has no upper bound. It's also a behaviour change for every one of the ~159 call sites and wants its own PR with a deliberate timeout budget, rather than riding along here.

### On #5446

Fine either way on order; whichever of us is second, I'll do the reconciliation if it's me.

One thing worth pinning down though. The identifier collision fails lint, so that half can't merge silently broken. The part CI would *not* catch is someone resolving it by renaming the constant and leaving `$fetchApiCachedWithRetry` in place: lint goes green and you quietly get 3×3 = 9 attempts per dead URL, which is exactly the wall-clock cost the 4→3 change here was meant to avoid. So the rename is the trap, not the duplicate declaration.

Same shape for `fetchTotalPluginsCount`: `number` here, formatted string there. `pluginCount.ts` conflicts loudly so it gets caught, but the seven call sites this PR moved onto it don't all conflict, and a `string` reaching `getSections()` is a type error rather than anything visible. `astro check` should catch that one.

---

iitzIrFan (approved):
Re-verified on `e70c2db`. `/use-cases` returns 200 under a dead `API_URL` where it was the only page 500ing before, and renders 26 cards against the live API. The homepage guard compiles to `section[data-astro-cid-…]:not(:has(.blueprint))` and works — degraded output has zero `class="blueprint"`, so the band and the `0+ Blueprints` heading go together; live it's 20 cards and `690+ Blueprints`. `Retry-After` handles both forms with past dates falling back. 339 tests, oxlint clean, 12/12 green.

I re-ran the bare-`fetch(` sweep independently and agree it's clear. `pluginIcon.ts:7` is also bare and throws, but both build-time callers wrap it, so it degrades.

Agreed on both declines — the floor value is a content call and the timeout wants its own PR.

And you're right that the rename is the trap, not the duplicate declaration; that's the better framing.

On ordering: this should go first. I've written up why on #5446 — that branch turns the two count endpoints into hard build dependencies (`/`, `/get-started` and `/data` all 500 under a dead API), which would partly undo the guard you just added to `home/Blueprints.astro`. I'll rework that side on top of this one, including dropping `$fetchApiCachedWithRetry` so the `RETRY_BASE_DELAY_MS` collision never reaches `main`.

PR #5639, opened 2026-09-15
perf: load Shiki lazily, per grammar, and render fences before it lands

### LINKED ISSUE #5636: Load Shiki lazily: 136 KB gzipped of grammars ships on every route
### Problem

Every route downloads Shiki's core, four themes and sixteen grammars before anything renders, whether or not a code block is ever shown.

Measured on the built output (`dist/client` served locally, `/about-us` loaded headless, every `.js` response counted after network idle):

| | raw | gzipped | share of page JS |
|---|---|---|---|
| Shiki closure (20 chunks) | 664,687 B | 136,386 B | 53% raw / 39% gzipped |
| Rest of page JS (43 chunks) | 585,534 B | 210,590 B | |
| Total | 1,250,221 B | 346,976 B | |

Largest offenders inside that closure: `javascript` 175 KB, Shiki core 107 KB, `github-dark` 80 KB, `python` 70 KB, `shellsession` 42 KB, then java, sql, groovy, terraform, yaml, hcl, xml and friends.

### Why it is eager

The chain is static all the way down:

```
layout.astro  ->  <Search client:load />
Search.vue    ->  AiChatDialog.vue
AiChatDialog  ->  MarkdownRenderer.vue
MarkdownRenderer.vue:17  ->  ~/markdown/marked-shiki
marked-shiki.ts:3        ->  ~/components/plugins/schema/shikiToolset
shikiToolset.ts:1-26     ->  shiki/core + 4 themes + 16 grammars
```

`getHighlighterCore()` is already `async`, but the grammars are pulled in by the top-level imports of `shikiToolset.ts`, so awaiting it buys nothing: the chunk is already in the page's static closure. `Search` is `client:load` in `layout.astro`, so a syntax highlighter for search results downloads on every page of the site.

The plugins routes reach the same closure by a second path, through `KsMarkdown` and `PluginsMarkdownRender`.

Docs prose is unaffected: fenced code in `.md` is highlighted at build time by `astro-expressive-code`. This is only the client-side path for API-sourced markdown (search results, AI chat, plugin docs, blueprint descriptions).

### Second, related problem: nothing renders until it lands

`MarkdownRenderer.vue` awaits `getMarked().parse()`, and that parse awaits the highlighter. Until the whole 136 KB arrives and the highlighter initializes, the component shows a skeleton. The text of a search result or an AI answer is held hostage by the highlighting of code that may not even be in it.

### Requirements

1. **Code is readable before Shiki loads.** Markdown renders immediately with fenced blocks as plain `<pre><code>`, escaped and legible. Highlighting is an upgrade applied when the highlighter resolves, not a precondition for showing anything. `CodeTabs.vue` already does this with `.shiki-fallback` (`src/components/common/CodeTabs.vue:90`); the same shape should apply here.
2. **No layout shift on upgrade.** The fallback and the highlighted output must agree on font family, font size, line height, padding and border radius, so swapping one for the other moves nothing. Worth an explicit check at both themes, since the highlighted markup carries inline colours the fallback does not.
3. **No grammar or theme chunk in any route's initial static closure.** `shikiToolset.ts` must be reached only through `await import(...)`, and no static import may pull it. Verify by walking the built chunks, not by reading the source: the whole bug is that the source already looks lazy.
4. **Grammars load on demand, not as a block.** Registering sixteen languages up front to highlight one YAML fence is most of the weight. A per-language dynamic import keyed on the fence's `lang` would cut the common case to core plus one grammar.
5. **Themes too.** Four themes ship today (`github-light`, `github-dark`, `github-light-default`, `github-dark-default`) at ~14-80 KB each. Two of them are addressed by name in `marked-shiki.ts`; the other pair comes from `shikiToolset.ts` serving a different consumer. Worth confirming all four are still needed.
6. **SSR path unchanged.** `MarkdownRenderer.astro`, `MarkdownRendererSSR.vue`, `SnippetsSSR.vue` and `SnippetsIsland.astro` highlight on the server and cost the client nothing. They must keep working, including the workerd constraint documented in `marked-shiki.ts` (the JavaScript regex engine, not the Oniguruma WASM one, which workerd refuses to compile).

### Acceptance

Rebuild, serve `dist/client`, load a page headless and count `.js` responses after network idle. No Shiki chunk in the set. Open search, type a query that returns a result containing a fenced block: the result text appears without waiting on the highlighter, and the code is readable in plain form before it is coloured.

### Rough sizing

Item 3 alone (one dynamic import boundary) should take the 136 KB off every route. Items 4 and 5 are follow-ups that shrink what a page pays when it does open search.

Found while measuring #5632, where removing bootstrap's global script saved 24.5 KB gzipped per route. This is five and a half times that.

elevatebart:
Closes #5636

Shiki's core, four themes and sixteen grammars sat in the static closure of every route. `Search` is `client:load`, the chain from there down to `shikiToolset.ts` was static all the way, and so a syntax highlighter for search results downloaded on every page of the site whether or not a code block was ever shown. `getHighlighterCore()` was already `async`, which bought exactly nothing.

### What changed

**`shikiToolset.ts`** no longer statically imports anything. Core, the JS regex engine, both themes and each grammar come through `import()`, and `getHighlighterCore(langs)` registers only what the caller names. Per-grammar loading forfeits Shiki's own alias registration at fetch time, so `LANG_ALIASES` routes `yml`/`sh`/`tf`/`console` and friends to the right chunk; without it those fences would have quietly downgraded to plain text. Core and grammar promises clear themselves on rejection, so a failed chunk fetch doesn't disable highlighting for the page's lifetime, and a grammar that fails leaves its own fence plain rather than failing the whole parse.

**`marked-shiki.ts`** reaches the toolset through `await import()` and asks for one grammar per fence. New `getPlainMarked()` renders fences as escaped `<pre class="shiki-fallback">`, parsed synchronously.

**`MarkdownRenderer.vue`** renders plain first, then swaps in the highlighted markup once it resolves. A failed upgrade keeps the readable fences instead of surfacing an error. The upgraded block carries no background of its own (marked-shiki strips Shiki's outer `<pre>`), so the fallback sets none either; the only rule it needs is the `pre-wrap` that Shiki puts on its per-line spans.

**Plugins and blueprints:** `SchemaToHtml` walks its schema for the languages its examples actually name, `SchemaToCode` falls back to `text` for a grammar not yet loaded rather than throwing out of a computed, and `CodeTabs` uses the core toolset instead of `await import("shiki")` (every grammar in the registry, to highlight one `lang`). `github-light` and `github-dark` are dropped: only reachable through defaults that every call site already overrode. The four components that parse before their first render start the fetch at island module scope, so its round trips overlap hydration instead of following it.

### Measured

`/about-us`, static closure walked over the built chunks:

| | raw | gzipped |
|---|---|---|
| before | 1,315,640 B | 367,114 B |
| after | 697,867 B | 244,448 B |

Zero Shiki chunks on that route, confirmed both by the chunk walk and headless after network idle. A page that does open search now pays core plus the fences' own grammars: about 211 KB raw / 62 KB gzipped for YAML, against 664 KB / 136 KB for the lot.

### Caveats

The SSR components (`MarkdownRendererSSR`, `BlueprintMarkdownBody`, `SnippetsSSR`) still re-parse during hydration, so on plugin and blueprint routes the grammars are fetched then rather than preloaded with the route. The module-scope warm-up overlaps that fetch with hydration, but it doesn't remove it. Killing it properly means passing the server HTML in as a prop (the `SnippetsIsland` -> `Snippets` shape) instead of re-deriving it on the client.

**On `KsMarkdown`** (an earlier revision of this description claimed plugin routes still load the design system's Shiki eagerly — that was wrong, and the correction is in the comments). `@kestra-io/design-system` does bundle core, 2 themes and 36 grammars into its `KsMarkdown` chunk, 628,210 B raw / 104,850 B gzipped, and upstream has not moved: 0.4.0 is npm latest and `kestra-io/kestra` HEAD carries the identical grammar list. But nothing here imports it directly — it arrives through `Topology.client.vue` → `@kestra-io/topology` → the design system, and both Topology usages are `client:only`, on blueprint-detail and orchestrate routes. It is in no route's static closure, so it doesn't qualify the goal; it's deferred weight on two route families, worth an upstream issue on its own merits.

### Verification

The build's prerender step fails in the environment this was developed in (blocked outbound calls to `api.kestra.io`), so the numbers above come from the client bundle and the two pages that did emit. `astro check` reports 0 diagnostics at every severity, the unit suite passes and oxlint is clean. Nine of those tests are new, covering the fallback markup, the alias table, the per-language registration and the copy-button injection, on the theory that a bug whose whole trick is looking lazy in the source deserves a test that doesn't take the source's word for it.

On Lighthouse, read the per-run CPU index before the deltas. The first two runs were unusable for the hydration-time question — runner indexes ranged 2227–4307 against a 2494 baseline, hiding deltas on most pages, and single-run pages swung by more than the effect being measured (Plugin Debug Return went 534 ms → 0 ms TBT between them). The third run landed close to baseline (2260–2642 vs 2475, 3 pages hidden). There, plugin debug TBT — the number that prompted the module-scope warm-up — reads 187 ms and is no longer flagged, having been 213 ms when it was. The three regressions that run flags don't reproduce across runs, and one of them (Blueprint Audit Logs TBT at 109 ms) is flagged while being lower than both earlier runs, so the baseline moves too. Treat any single flag here as a prompt to re-measure rather than a result.

https://claude.ai/code/session_01RLBsDEf2WdPpnW41sVAxCy


iitzIrFan (changes requested):
The lazy-loading approach looks good overall. I only see `import type` references to `shiki` left in `src/`, with the runtime paths behind `await import()`.

I found two things that I think should be addressed before merging:

**1. Dark mode fallback**

`_markdown-renderer.scss:23` hard-codes `background-color: #fff`, while the fallback `<code>` inherits `color: var(--ks-content-primary)` from `app.scss`. In dark mode that makes the fallback text white on a white background.

Also, `app.scss:27` targets `pre code span`, but the fallback doesn't have those spans. Since `marked-shiki.ts` also strips Shiki's outer `<pre>`, the fallback and upgraded states don't have the same background treatment.

Could we reuse the existing `var(--ks-background-secondary)` / `var(--ks-content-primary)` pattern from `CodeTabs.vue:153`, or avoid setting the background here?

**2. Cached failed imports**

In `shikiToolset.ts`, `corePromise ??= createCore()` and the `langPromises` cache the Promise itself. If the import rejects once, that rejected Promise stays cached, so highlighting won't retry for the lifetime of the page.

This differs from the previous `highlighterCoreCache` behaviour, which was only populated after a successful load. Could we clear the cached Promise on rejection so a subsequent attempt can retry? It would also be useful to avoid the empty catch in `warmHighlighterCore` making this failure completely silent.

A couple of smaller points:

* `injectCopyButtons` currently matches the literal `<pre>`, so it won't match `<pre class="shiki-fallback">`. That means the copy button only appears after the block is upgraded.
* The `LANG_ALIASES` comment could be clarified: the table is mainly acting as the routing table for lazy-loading the correct grammar; `loadLanguage` still registers Shiki's own aliases.

One other scope question: `KsMarkdown` still pulls the design system's Shiki on plugin routes, so the "no Shiki chunk on any route" goal doesn't hold universally. Not sure if that's intended to be a follow-up or something we should account for here.

elevatebart:
Thanks — both main points were real. Fixed in e47ed45.

**1. Dark mode fallback.** Right, and worse than described. I checked what the two states actually render:

```
UPGRADED: <pre><code class="language-yaml"><span class="line"><span style="color:#116329;--shiki-dark:#7EE787">id</span>…
FALLBACK: <pre class="shiki-fallback"><code>id: hello</code></pre>
```

Because `marked-shiki.ts` strips Shiki's outer `<pre>`, the upgraded block has no background of its own. So `#fff` was wrong twice over: unreadable in dark mode, and it never matched the upgraded state either — the comment I wrote claiming it did was simply false. I dropped the declaration rather than picking a colour, since "no background" is what it is matching. Only the wrapping still needs a rule, because Shiki puts `pre-wrap` on its per-line spans and the fallback has none.

**2. Cached failed imports.** Correct, and a behaviour change from `highlighterCoreCache` that I missed. Both the core promise and the per-grammar promises now clear themselves on rejection, so the next attempt retries.

While fixing it I found an adjacent problem: a rejected grammar promise propagated out of `getHighlighterCore` and failed the whole `marked` parse, so one bad chunk fetch would blank a document rather than just leave one fence unhighlighted. Grammar failures are now non-fatal, and `marked-shiki` re-checks `getLoadedLanguages()` before use instead of trusting `resolveLanguage` — the same defensive check `SchemaToCode` and `renderVersionedDoc` already do. Preload failures log instead of disappearing into an empty catch.

**`injectCopyButtons`.** Confirmed — literal `"<pre>"`, so the button only arrived with the upgrade. Now matches the tag with attributes, with a test covering both the fallback and upgraded markup.

**`LANG_ALIASES` comment.** Reworded. It is the routing table for the lazy import; Shiki still registers its own aliases once the chunk lands.

**`KsMarkdown`.** Follow-up, and it has to be — that Shiki comes from `@kestra-io/design-system`'s own bundle (core plus ~30 grammars, statically imported inside the package), so nothing in this repo can make it lazy. You're right that "no Shiki chunk on any route" doesn't hold universally; the measured claim is narrower and I should state it that way. What this PR removes is the site-wide closure that `Search` being `client:load` put on *every* route, `/about-us` included — 0 B there now, verified by walking the built chunks and headless after network idle. Plugin routes still pay the design system's copy. Happy to open an issue against the design system if you think that's the right venue.

---

iitzIrFan (approved):
Both blockers addressed, and I verified the behaviour rather than just the diff — forcing `shiki/core` to reject once now retries and recovers on the next call, and a failing grammar chunk degrades the fence to readable plain text instead of taking the parse down. Dropping the background entirely rather than picking a token is the right call given the upgraded `<pre>` has none either. The `getLoadedLanguages()` re-check and the copy-button tests are nice additions. 337 tests pass locally. LGTM — `KsMarkdown` remains the one open scope question for @brian-mulier-p.

elevatebart:
@iitzIrFan @brian-mulier-p — I dug into the `KsMarkdown` question. Short version: **upstream hasn't moved, but my caveat was wrong and the goal does hold.** Correcting it below.

**Upstream status.** `@kestra-io/design-system` 0.4.0 is npm latest, and the source at `kestra-io/kestra` HEAD (`e6155c2`, today) still eagerly imports Shiki core, 2 themes and 36 grammars in `ui/packages/design-system/src/components/Data/KsMarkdown/shikiHighlighter.ts`. I diffed the grammar list against the shipped 0.4.0 dist: identical. Last change to that directory was 2026-09-02 (kestra-io/kestra#18860). So there is no unreleased fix waiting on a release.

**But it isn't on plugin routes, and it isn't eager.** I wrote "plugin routes still pay for it" — wrong on both counts:

- Nothing in `src/` imports `KsMarkdown`. It arrives transitively: `Topology.client.vue` → `@kestra-io/topology` (which peer-depends on the design system) → `KsMarkdown`.
- Both `Topology.client.vue` usages are `client:only` — `pages/blueprints/[id].astro` and `components/orchestrate/OrchPattern.astro`. That's a runtime dynamic import, and those are blueprint-detail and orchestrate routes, not plugin routes.
- Walking the built graph confirms it: `KsMarkdown` is in no page's static closure, and every static path into it tops out at a chunk whose only importers are dynamic.

So requirement 3 of #5636 — no grammar or theme chunk in any route's initial static closure — holds without the qualification I added.

What's true is narrower: on a blueprint detail or orchestrate page, opening the topology island pulls a 628,210 B / 104,850 B gzipped chunk that is mostly those 36 grammars. Deferred, on two route families, and owned by the design system. Worth an upstream issue on its own merits, but it isn't a caveat on this PR and it doesn't block it. I've rewritten the PR body accordingly.

One aside: the design system's `loadLanguageOnDemand` drops its memo on failure with the same reasoning @iitzIrFan asked me to apply here ("a cached rejection would fail every later load"). Converged independently, which is mildly reassuring about the fix.

---

PR #5647, opened 2026-09-16
fix(build): stop duplicate dark logos churning the customers cache key

elevatebart:
## What

Seven customer stories ship a `logo-dark.svg` that is byte-identical to their `logo.svg`. This one-line-each change points `logoDark` at `./logo.svg` in those seven, which gets all 29 `/customers/<slug>` pages restoring from the incremental build cache instead of re-rendering on every single deploy.

## Why they never restored

Vite dedupes assets by content hash, so two byte-identical source files collapse into one emitted asset. That asset inherits the `[name]` of whichever import won the emit race, and the winner varies between builds. So `entry.data.logo.src` and `entry.data.logoDark.src` flip between `/_astro/logo.<hash>.svg` and `/_astro/logo-dark.<hash>.svg` depending on the mood of the module graph that day.

`customers/[slug].astro` folds `collectionContentDigest(allEntries)` into every page's `cacheKey`, and that digest stringifies `entry.data` for the whole collection into a single scalar. One flipped logo anywhere in the collection invalidates all 29 pages.

Two consecutive builds with no source edits, `dependencyHash` identical, no `contentHashes` recorded for the route:

```
cacheKey a: 1d4b4e9908a111ae|4eb73979e16a3e69|l2.0|4e3403cdea35a3a5
cacheKey b: 1d4b4e9908a111ae|a613ca69a63396c9|l2.0|4e3403cdea35a3a5
                             ^^^^^^^^^^^^^^^^ collectionContentDigest
```

29/29 paths differed, all on segment 2. Dumping the digest rows across two builds narrowed it to exactly two fields:

```
rows changed: 4 / 29
fields that churn: { logo: 4, logoDark: 4 }

airpaz  logo A: /_astro/logo.DtA9YSGA.svg
        logo B: /_astro/logo-dark.DtA9YSGA.svg   # same hash, different name
```

Same content hash, different filename. A repo-wide scan found 7 such pairs; only 4 flipped in that particular build pair, because the race is probabilistic.

Worth recording what this was **not**: the random `related` sibling pick is innocent (`canSkip()` never hashes props, and `contentHashes` is not even recorded for this route), `dependencyHash` is stable, and `collectionContentDigest` is doing its job correctly. Its input was the problem.

## Verification

Clean seed, then a warm build, on a machine that can reach `api.kestra.io`:

```
customers lines:     30
customers restored:  29
total restored:      2239 -> 2268  (+29)
```

The 30th line is `/customers/index.html`, a separate route with no `cacheKey`. All 29 cache keys are now byte-identical across builds, and all 95 `/_astro` image references across the 30 pages resolve to emitted assets, so the asset-existence check stays green.

## Output safety

Every pair was checked with `cmp` before editing. Identical bytes means identical content hash means pixel-identical rendering. The only change is that the filename stops racing between two names, which is strictly more deterministic than what we had.

The other 17 stories have genuinely distinct dark logos and are untouched.

## Second commit

The seven `logo-dark.svg` files are unreferenced once `logoDark` points elsewhere, so nothing imports them and nothing emits them. Deleted (~640KB, one of them 296KB on its own).

Re-ran the seed + warm pair afterwards and got the same numbers: 29/29 restored, 2268 total, 0 of 95 image references missing. The 18 `logo-dark` assets still in `dist/_astro` belong to the stories that have real dark logos.

## Follow-ups, deliberately not in here

- `src/pages/docs/changelog/[tag].astro` is the route behind the 101 always-fresh changelog pages, and it still has no `cacheKey`. It also hits `api.github.com` once per page, which is the host that has been rate-limiting CI. Bigger win than this PR, and independent of it.
- `src/contents/blogs/2024-10-15-deploying-kestra-in-clever-cloud/` has the same duplicate-content pattern between two step screenshots. It does not affect any cache key, but a restored blog page can carry the losing filename while the build emits the winner. Fixing it means editing markdown that references two semantically distinct steps, so it wants a human.


elevatebart:
Superseded by `elevate-bot/trusting-ride-1gyb8x` (commit 0dfdd5ee1a), which reaches the same root cause and covers more of it:

- 9 duplicate pairs rather than the 7 here, adding the story logos that `use-cases/public-services` and `use-cases/financial-services` duplicate under `src/assets`, and the `hero.jpg` shared by the two government stories
- drops the `logoDark` field outright (it is passed around but never rendered) instead of repointing it at `logo.svg`
- also keys `vs/[slug]` and `orchestration/[slug]` on the live API data they render, which was missing from their keys entirely and could leave those pages showing stale plugin counts
- makes `entryCacheKey` fail closed when any part of its scope is undefined

Closing in favour of that branch. The analysis here still stands if it is useful for review: two consecutive no-edit builds, `dependencyHash` identical and no `contentHashes` recorded, with only `collectionContentDigest` moving.

```
cacheKey a: 1d4b4e9908a111ae|4eb73979e16a3e69|l2.0|4e3403cdea35a3a5
cacheKey b: 1d4b4e9908a111ae|a613ca69a63396c9|l2.0|4e3403cdea35a3a5
                             ^^^^^^^^^^^^^^^^ collectionContentDigest
```

Worth flagging for whoever reviews the surviving branch: the race is intermittent. A later build pair on unfixed `main` restored 29/29 customer pages simply because both builds happened to pick the same filename, so a single green build pair does not show the bug is gone.

The changelog cache key has moved to #5648, which is independent of all of this.

PR #5649, opened 2026-09-16
fix(build): key pages on the API data they render, and stop duplicate images churning asset URLs

elevatebart:
## What the branch does

One commit, `0dfdd5ee1a`, covering two independent bugs in `experimental.incrementalBuild`:

1. **Cache keys missing inputs.** `vs/[slug]` prints the live plugin count but keyed only on the entry and layout, so a plugin release left every comparison page showing a stale number. `orchestration/[slug]` fetches blueprint sources, graphs, a blueprint count and plugin icons, none of it keyed. Adds `apiPayloadDigest` and `pluginIconDigest`, and makes `entryCacheKey` fail closed when any scope element is undefined.
2. **Byte-identical images.** Vite folds two identical sources into one emitted file and keeps whichever name arrived first, which varies per build. Nine pairs removed at source.

It also memoizes plugin icons through `$fetchApiTextCached`, which were previously refetched per page (~86 of them).

## State

- Merges cleanly with current `main` (verified with `git merge-tree`), though it is 8 commits behind
- `npx vitest run src/utils/incrementalCacheKey.test.ts`: 19 passed
- `astro check`: 0 errors, 0 warnings, 0 hints
- `scripts/check-dist-asset-links.mjs`: all `/_astro/` references resolve

## Bug 2 verified

Confirmed independently against live APIs. Two consecutive no-edit builds on unfixed `main`, `dependencyHash` identical and no `contentHashes` recorded for the route:

```
cacheKey a: 1d4b4e9908a111ae|4eb73979e16a3e69|l2.0|4e3403cdea35a3a5
cacheKey b: 1d4b4e9908a111ae|a613ca69a63396c9|l2.0|4e3403cdea35a3a5
                             ^^^^^^^^^^^^^^^^ collectionContentDigest
```

Only `collectionContentDigest` moved, and only `logo` / `logoDark` inside it, with the same content hash under two different filenames. On this branch the customer pages restore 29/29.

One caveat for whoever reviews: **the race is intermittent**. A later build pair on unfixed `main` restored 29/29 purely because both builds happened to pick the same winner. A single green build pair does not show the bug is gone.

Three byte-identical pairs remain on the branch, all inert: `quotes-companies/Foundation.svg` is never imported, `trusted-companies/JPMorgan.svg` is a `?raw` import so emits no asset, and the two `bloomberg.svg` copies sit outside any collection's `entry.data` and so cannot reach a collection digest.

## Bug 1 regresses orchestration, and should block merge

`orchestration/[slug]` goes from restoring to **zero restored**, all 38 pages, every build. The new key is the cause, not the images:

```
src/pages/orchestration/[slug].astro
  depHash: SAME | cacheKey diff: 38/38
  a: 42b37f552bc8ba77|da4d8fefefc36df7|l2.0|4e3403cdea35a3a5|7d55cc54bf98b54c|4f384f599347def1
  b: 42b37f552bc8ba77|da4d8fefefc36df7|l2.0|4e3403cdea35a3a5|39b5654eca95a276|4f384f599347def1
                                                             ^^^^^^^^^^^^^^^^ apiPayloadDigest(blueprintPayloads)
```

The reason is that `/blueprints/{id}/versions/latest/graph` mints fresh node UIDs on every single request:

```
A: {"nodes":[{"uid":"root.root-4I7fnx8HTFKLCR4OS0U92R", ...
B: {"nodes":[{"uid":"root.root-OcdIE0yZQoJEUFVaBno2i",  ...
```

Probing the 264 URLs these pages digest, twice back to back, 113 differed, and every one was a `/graph`. The other payloads (`versions/latest`, the `?q=&size=1` counts) are stable.

So `apiPayloadDigest` over a graph payload can never produce a stable key. The commit's own verification, "restored is 2266/2398 on three consecutive builds", was measured against a deterministic API stub, which is exactly what hides this.

Worth fixing before merge, probably by digesting only the fields the page actually renders from the graph, or dropping the graph from the key and keying on the blueprint source it is derived from. Net effect as it stands is a trade of 30 always-fresh customer pages for 38 always-fresh orchestration pages.

`vs/[slug]` is fine: it dipped to 0 on one build because `/plugins/subgroups` genuinely changed, then returned to 38/38 with stable keys on the next pair. That is the key doing its job.


elevatebart:
Pushed 98e5338641, which fixes the orchestration regression flagged above and a second instance of the same defect on `vs`.

Neither endpoint `apiPayloadDigest` is used on returns a byte-stable response, so both keyed routes churned every build.

**`/plugins/subgroups`** returns the same plugins in a different array order per request. Five calls gave three distinct responses, all of identical length (3840951 bytes), all identical once canonicalized. That was holding `vs/[slug]` at 0/38 restored, which I had wrongly written off above as the key doing its job.

**`/blueprints/{id}/versions/latest/graph`** mints fresh ids for its synthetic root/end/finally nodes on every request, and varies its edge order as well. Of the 264 payloads these pages digest, probed twice back to back, 113 differed and every one was a graph.

Two changes:

- `canonicalPayload` orders object keys and array elements before hashing, so a digest tracks content rather than arrival order. Sufficient for the plugin payload.
- Not sufficient for the graph, whose ids are genuinely new each time, so the graph drops out of the key and `/blueprints/{id}/versions/latest` stands in for it. The graph is derived from that source server-side, so the source is the real input.

Three consecutive builds against the live API:

```
build2: orch 38 | vs 38 | customers 29 | total 2266
build3: orch 37 | vs 38 | customers 29 | total 2265
```

The single orchestration miss is `/orchestration/gcs`, whose per-tool blueprint count legitimately changed between builds; all three of its blueprint sources verified stable. Tests 24 passed, `astro check` clean, `check-dist-asset-links.mjs` green.

Worth noting for the record: the original verification missed all of this because it ran against a deterministic API stub. A stub cannot show you that the endpoint you are keying on is non-deterministic.

elevatebart:
Merged `main` in (2d361804d3), resolving the conflicts.

All six were modify/delete on `logo-dark.svg`: #5651 ("cut SVG payload from 111 MB to 12 MB") rewrote the files this branch deletes. I took the deletion in each case, which matches the branch's rationale that `logoDark` is never rendered. Confirmed on current `main`: `logoDark` appears only as an optional prop type in `FeaturedCarousel.vue` and `StoryGallery.astro` and in frontmatter, never in a template.

**One judgement call worth a look.** After #5651, `sopht-scales-its-green-itops-platform-with-kestra/logo-dark.svg` is no longer byte-identical to its `logo.svg` (the other five still are). So that deletion drops a file that is now a distinct asset rather than a duplicate. It changes nothing rendered, for the reason above, but it is the one file here that is not a pure dedup. Easy to restore if you would rather keep it.

The branch is still needed: `main` currently has seven byte-identical pairs in `customer-stories` (six logo pairs plus the government `hero.jpg`), all of which this branch removes. After the merge there are none.

Three consecutive builds on the merge result:

```
build2: orch 35 | vs 38 | customers 29 | total 2277
build3: orch 36 | vs 38 | customers 29 | total 2278
```

Tests 24 passed, `astro check` clean, `check-dist-asset-links.mjs` green.

**Known residual, not fixed here.** Two or three orchestration pages still re-render per build. Probing the 151 payloads these pages digest, twice, exactly two differ, and both are the `/blueprints/versions/latest?q=<tool>&size=1` count endpoints. Their `total` is stable (GCS was 42 in both passes); what moves is the single `results` item that `size=1` returns, which the page never reads, since it only uses `search.total`. So the key is over-broad rather than unstable. Narrowing it to `total` would close the gap; happy to do that here or leave it as a follow-up.

@iitzIrFan added as reviewer.

elevatebart:
Merged `main` again (3bf20fa285). #5648 landing plus two asset/count PRs put seven fresh conflicts on this branch; all resolved.

**Content (5 files).** #5685 ("WebP conversion redo") rewrote the story frontmatter to `.webp` while this branch was removing the `logoDark` line. Took main's `.webp` filenames minus `logoDark`.

**`vs/[slug].astro`.** #5446 ("single source of truth for plugin and blueprint counts") replaced this branch's `apiPayloadDigest("/plugins/subgroups")` with ```p${await fetchTotalPluginsCount()}```. Took main's version outright: it keys on the count the page actually prints instead of digesting the whole 3.8 MB payload, which is narrower, cheaper, and immune to the array-ordering churn that the canonicalization in 98e5338641 was working around. Dropped the now-unused `apiPayloadDigest` import. `canonicalPayload` stays, since `orchestration/[slug]` still digests blueprint payloads through it.

**`src/utils/fetch.ts`.** Textual collision only, two unrelated functions landing at the same offset: this branch's `$fetchApiTextCached` and main's `errorStatus` helper for `$fetchApiCachedOptional`. Kept both.

**Orphans.** The WebP conversion recreated the five `logo-dark` files this branch had deleted, under the new extension. Removed the `.webp` copies too, so the frontmatter change does not leave them dangling.

One thing that resolves itself: sopht's two logos are byte-identical again after the WebP conversion, so the judgement call I flagged last round is moot. All five are plain duplicates now. `customer-stories` has zero byte-identical pairs after this merge.

Three consecutive builds:

```
build2: orch 38 | vs 38 | customers 29 | changelog 100 | total 2380
build3: orch 35 | vs 38 | customers 29 | changelog 100 | total 2377
```

Tests 24 passed, `astro check` clean, `check-dist-asset-links.mjs` green, no conflict markers left in `src/`.

The orchestration wobble is the same known residual as before, unchanged by this merge: the key digests the whole `?q=<tool>&size=1` response when the page reads only `search.total`. Still worth narrowing, still not done here.

src/pages/orchestration/[slug].astro
```diff
@@ -88,6 +91,25 @@ const LOGO_MAP: Record<string, string> = {
...
+    // module, so anything this needs has to be defined inside it.
+    // The graph is derived from the flow source, and the API regenerates its
+    // node ids and edge order per request, so the source stands in for both.
+    const blueprintPayloads = (data: OrchestrationData) => [
+        ...data.blueprints.items.flatMap((item) =>
+            item.blueprintId
+                ? [`/blueprints/${item.blueprintId}/versions/latest`]
+                : [],
+        ),
+        `/blueprints/versions/latest?q=${encodeURIComponent(data.tool.name)}&size=1`,
```
Comment (iitzIrFan) [inline src/pages/orchestration/[slug].astro]: The page reads only `search.total` from this endpoint, but the key digests the whole response, which is why two or three orchestration pages still re-render every build. `vs/[slug]` keys on the rendered count with `p${count}`; the same pattern here (fetch via `$fetchApiCached`, key on `b${total}`) would close that residual.

src/utils/incrementalCacheKey.ts
```diff
@@ -74,7 +74,74 @@ export async function layoutDigest(): Promise<string> {
...
-    ...scope: string[]
+    ...scope: (string | undefined)[]
 ): string | undefined {
-    return entry.digest === undefined ? undefined : [entry.digest, ...scope].join("|")
+    return entry.digest === undefined || scope.includes(undefined)
+        ? undefined
+        : [entry.digest, ...scope].join("|")
+}
+
+/** Canonical form of a payload, with object keys and array elements ordered, so
```
iitzIrFan, line 84: Worth stating here that this is lossy: sorting array elements means two payloads that differ only in element order hash identically. Fine for the current inputs (a flow string and a total), but a future caller keying a page on an ordered list would silently miss changes.

iitzIrFan (changes requested):
Reviewed against the branch head (3bf20fa28) locally, with the dev server running. Tests, CI, and rendering all check out; two small changes requested before merge.

## Verified

- `npx vitest run src/utils/incrementalCacheKey.test.ts`: 24 passed. All CI checks green.
- `/orchestration/*`, `/use-cases/public-services`, `/use-cases/financial-services`, `/customers`, `/customers/dataport` and `/customers/government-it-provider` render on the dev server. The Dataport and JPMorgan logos now resolve to the story `logo.svg` and load. Both government stories reference the single shared `hero.jpg`, so the cross-directory frontmatter path works.
- Probed the 43 plugin classes the orchestration key digests (`tool.pluginClass` plus competitors) against the live icon endpoint: all 200. So the fail-closed `pluginIconDigest` does not leave any page permanently uncached today.
- `logoDark` is passed through but never rendered (`FeaturedCarousel.vue:127` falls back to `logo`), so the seven deletions change nothing visible.

## Fits the codebase

- New digests live in `src/utils/incrementalCacheKey.ts` with tests alongside, and the `getStaticPaths` shape matches the other eleven `entryCacheKey` callers. The fail-closed `entryCacheKey` change is in the spirit of the helper's own doc comment and affects no existing caller.
- `$fetchApiTextCached` mirrors `$fetchApiCached` exactly. Side effect worth a line in the description: `fetchPluginIcon` now gets retries on 5xx and per-build memoization, where it previously used a bare `fetch`.
- Importing a story `logo.svg` from a use-case page has precedent (`automotive.astro`, `orchestration/[slug].astro`).

## Requested changes

1. **Key the blueprint count on the rendered value, not the payload.** `src/pages/orchestration/[slug].astro` reads only `search.total` from `/blueprints/versions/latest?q=<tool>&size=1`, but `blueprintPayloads()` digests the whole response. Your own build pairs show two or three orchestration pages re-rendering every build because of it. This PR already moved `vs/[slug]` to the narrower `p${count}` pattern; doing the same here with something like `b${total}` (fetch via `$fetchApiCached`, read `.total`) is a few lines and closes the residual you flagged twice.

2. **Document that `canonicalPayload` is lossy.** Sorting array elements means two payloads that differ only in element order hash identically. Safe for the current inputs (a flow string and a total), but it is an exported utility, and a future caller keying a page on an ordered list would silently miss changes. A one-line caveat in the doc comment is enough; making the array sort opt-in would be stricter but is not needed now.

elevatebart:
Both changes made in 838a87ab1f, plus the description note.

**1. Blueprint count keyed on the rendered value.** `blueprintPayloads()` no longer digests the search response; `blueprintCount()` fetches it via `$fetchApiCached`, reads `.total`, and contributes ```b${total}``` to the key, matching the `p${count}` pattern. It mirrors the page's own `?? 0` fallback rather than failing closed, so the key tracks what actually renders: if the request fails the page shows 0 and the key says `b0`, and the next build that succeeds invalidates it. That is also what `fetchTotalPluginsCount` does with `PLUGIN_COUNT_FLOOR`, so the two counts now behave the same way on failure.

This closes the residual exactly as you predicted. Three consecutive builds:

```
build2: orch 38 | vs 38 | customers 29 | total 2380
build3: orch 38 | vs 38 | customers 29 | total 2380
orchestration cacheKey diff: 0/38
```

Previously 35-38 and never twice the same. Also dropped "graphs" from the comment above the key, which had been stale since the graph came out of it.

**2. `canonicalPayload` caveat.** Added to the doc comment, and pinned with a test so the limitation is asserted rather than only described:

```ts
it("is lossy about order, which reordering alone cannot be detected through", ...)
```

Left the array sort unconditional as you suggested. Worth noting it now has exactly one caller: `vs/[slug]` moved to `fetchTotalPluginsCount()` when `main` merged, so the only payloads still going through `apiPayloadDigest` are the blueprint flow sources. If `orchestration` ever stops needing it too, the whole helper can go.

**Description.** Added the `fetchPluginIcon` line: it used a bare `fetch`, so it picks up `internalFetch`'s 5xx retries and per-build memoization along with the digest work.

Thanks for probing all 43 plugin classes against the live endpoint. That was the part I had not checked, and a permanently-uncached page from a 404 there would have been easy to miss. Tests now 25 passed, `astro check` clean, `check-dist-asset-links.mjs` green.

iitzIrFan (approved):
Both requested changes landed in 838a87ab1, re-verified locally.

- The orchestration key now carries `b${total}` from the count endpoint instead of digesting the whole `size=1` payload, and the `b0` fallback matches the page's own fallback, so the key tracks what renders. The full-payload digest still covers the blueprint sources, which are the other input the page bakes in.
- `canonicalPayload` documents its order-lossiness and has a test pinning it.

`npx vitest run src/utils/incrementalCacheKey.test.ts`: 25 passed. Lint & Astro Check and CodeQL green on the new head. Approving.

PR #5651, opened 2026-09-16
perf(assets): cut SVG payload from 111 MB to 20 MB

elevatebart:
Audit of every SVG over 5 KB, then the three changes it justified.

**`public/` + `src/` SVG: 110.85 MB across 727 files → 20.15 MB across 588.** No `.webp` files are added.

| Commit | Files | Before | After |
|---|---:|---:|---:|
| Delete unreferenced | 138 | 9.2 MB | 0 |
| Figma dot meshes → `<pattern>` | 20 | 75.0 MB | 1.0 MB |
| SVGO on disk | 218 | 10.1 MB | 7.4 MB |
| Grid backgrounds → `<pattern>` | 9 | 236 KB | 32 KB |

Assets only — `tests/` is identical to `main`. The added visual regression coverage moved to #4588.

## The dot meshes

Twenty use-case illustrations carried a dot field Figma exported one element at a time: 11,552 circles drawn as beziers, every one referencing its own `<linearGradient>`, and all 11,552 gradient definitions byte-identical. Thirteen were 5.6 MB each. Roughly half a million DOM nodes to draw a texture.

The field is now one `<pattern>` tile masking a rect filled with the original gradient. That gradient is `userSpaceOnUse` and vertical, so a rect masked to the dot pattern reproduces each dot's colour by construction rather than by approximation.

## SVGO

Nothing under `public/` reaches Astro's `experimental.svgOptimizer`, which only sees `astro:assets` imports. `scripts/svgo-batch.mjs` covers both roots and `npm run svgo:check` fails when a file drifts.

Four `preset-default` plugins are off, each for a defect found by rasterising all 249 files and diffing:

- **`collapseGroups`** flattened gradients on the isometric artwork. Collapsing a group changes the bounding box a `gradientUnits="objectBoundingBox"` gradient resolves against, so the cube faces in `use-cases/capabilities` lost their shading.
- **`convertTransform`** rounds to five significant digits. Figma wraps rasters in patterns scaled by factors like `0.000381825`, where five digits drops two of them and moves the image. The plugin saved 0.01 MB.
- **`convertShapeToPath`** lost rounded-corner geometry on rects inside a transformed group.
- **`cleanupIds`** stays off: several files are inlined into one document with `?raw`, where renamed ids collide.

Precision is per file — three decimals except where a generated pattern is present, keyed off `patternUnits` rather than a path list so a future run cannot undo the pattern work.

## Verification

Every changed file rasterised before and after in Chromium and pixel-diffed, rendered as `` the way the site consumes them — the composed SVG, not its component parts.

- Dot meshes: foreground illustrations pixel-identical, peak delta 61/255 confined to dot antialiasing.
- SVGO: 122 of 249 pixel-identical, 127 differ only in edge antialiasing, none structurally.
- Grids: peak delta 28/255, at most 0.4% of pixels above 16.

The residual is inherent to tiling — a pattern quantises every dot to the same sub-pixel phase where the original let each land where it fell.

## Corrections made during review

**A WebP conversion was reverted in full.** It extracted the raster embedded in 23 SVGs and re-encoded it, but all 23 compose that payload through a `viewBox` crop and most add masks, clipPaths, filters or blend modes. Lifting the payload discarded the composition. Caught by @iitzIrFan; redoing it means rasterising the composed SVG, and belongs in its own PR.

**13 `/vs/*` competitor logos were restored.** `getRawSvg(logoFile ?? slug)` in `vs/[slug].astro` resolves a bare `${slug}.svg` and returns `""` on a miss, so deleting those files emptied 13 pages with a green build. The orphan sweep had accounted for the `-img` and `-monograme` lookups but not this one.

## Two corrections to the original audit

The audit called the dot meshes and the grid backgrounds CSS-gradient candidates. Neither is.

Each dot mesh is a content illustration whose field sits behind foreground artwork under its own mask geometry. Each grid pairs a repeating field with a scatter of hand-placed highlight cells, and eight of the nine use that scatter as a mask over blurred colour ellipses. `gridbg.svg` looked like the one genuine exception until a side-by-side render showed the 19 highlight cells CSS would have dropped. A gradient reproduces the lines and discards the artwork, so both stay SVG and only the repeating part changes.

## Review notes

Worth a look on the preview: `/use-cases/ci-cd` (widest converted hero), `/use-cases/monitoring`, `/use-cases/change-data-capture`, `/use-cases/modern-data-stack`. The dot fields should look exactly as before, which is the entire point. `/vs/camunda` and `/vs/luigi` confirm the restored logos.

https://claude.ai/code/session_01XPuURKJjvfJhd3gq2ko1qu


iitzIrFan (changes requested):
Requesting changes. The SVGO and `<pattern>` work is excellent and I verified it independently — the dot meshes and grids are faithful under native-resolution pixel diffing (amplified diffs are essentially black), the 111 MB → 11.5 MB number holds (`public/` + `src/`: 110.77 MB / 721 files → 11.55 MB / 551), and the four disabled `preset-default` plugins are each correctly justified. That part I'd merge as-is.

Two things need to come out or be redone first, and neither trips CI — all 12 checks are green because both failure modes are silent.

## 1. Thirteen `/vs/*` pages lose their competitor logo

`getRawSvg()` in `src/pages/vs/[slug].astro` resolves the glob by runtime string match and returns `""` on a miss, so a deleted file is never a build error:

```js
const key = Object.keys(rawSvgModules).find((k) => k.endsWith(`/${filename}.svg`))
return key ? rawSvgModules[key] : ""
```

`logoFilename = logoFile ?? slug`, and the 13 files removed from `src/components/versus/assets/` are exactly 13 live `/vs/` slugs:

`apache-nifi`, `broadcom`, `camunda`, `chef`, `control-m`, `hp-operations-orchestrator`, `ibm-workload-automation`, `luigi`, `pipedream`, `simplyask`, `stonebranch`, `tines`, `workato`

On `main`, only `ansible-automation-platform` and `rundeck` lack a slug-named file — both have an explicit `logoFile:` override. On this branch, 15 miss.

The logo is inlined via `set:html` into `<div class="picto tool" aria-label="…">` and reused across Philosophy / BusinessValue / SetupSpeed / EaseOfUse / VisualPreview / FeatureGrid / DecisionGuide. Production `/vs/camunda` contains 53 occurrences of the Camunda brand colour; on this branch every one becomes an empty div.

`src/pages/orchestration/[slug].astro:52` has the same `?? ""` pattern — its `logoAsset` values don't currently overlap, but it's the same trap.

A basename grep can't see these, which is the gap worth closing before re-running the audit.

## 2. The WebP conversion emits a different image

The premise that these were "an `<svg>` wrapper around a single base64 PNG, no vector content" doesn't hold for the converted set. These SVGs wrap the raster in `<mask>`, `<clipPath>`, `<filter>`, blend modes and background rects, and crop it via `viewBox`. The converter extracted the embedded payload and re-encoded it, discarding all of that.

**20 of 23 conversions are visibly wrong.** Rendering each `main` SVG next to its replacement WebP:

| File | `main` | this branch |
|---|---|---|
| `assets/layout/started/cloud.svg` | purple Kestra cloud | **AICPA SOC compliance badge** |
| `components/cloud/assets/cloud.svg` | purple Kestra cloud, transparent | blue checkmark cloud on a pedestal, opaque dark bg |
| `public/hero-bg-effects.svg` | soft blurred light-leak | the glossy cube cluster |
| `assets/landing/see_how.svg` | 600×600 composed scene (`#131313` rect, `mix-blend-mode` glow, transformed raster) | the naked cube cluster |
| `assets/teams/*.svg` (9 headshots) | circular avatar crops | uncropped rectangles with grey/pink/green photo backgrounds |
| `assets/teams/techcrunch.svg` | transparent | opaque white box |
| customer-story `icon.svg` (4) | square icon marks | full wordmark lockups (`Ntico`, `sopht`, `VÍSSIMO group`, `Regulated COMPANY`) |
| `plugins/assets/{kestra,conapi}.svg` | 24×24 transparent marks | 200×200 marks on solid purple / yellow |
| `infrastructure/arrow.svg` | 401×340 padded frame | 166×178 tight crop — renders far larger in its box |

Live surfaces hit: the `/cloud` page header, the Get Started hero, the `SeeHow` CTA block (site-wide), the plan card in `GetStarted.astro`, the investor grid in `Investors.astro`, `Explore.astro`, and the customer-story `logoIcon` slots.

The clearest evidence of the mechanism: `hero-bg-effects.svg` (1720×818) and `see_how.svg` (600×600) share one embedded payload (`md5 329c12c5…`) but compose it completely differently — and both produced the **byte-identical** 1641×1697 WebP.

`Visuals.webp` and `julia-capabilities.webp` are the only two that came through clean; both are genuine passthrough wrappers. Tellingly, the three files kept as SVG-with-swapped-payload (`axeleo-new`, `isai-new`, `west-bend`) all render correctly, for exactly the reason the extracted ones don't — the wrapper survived.

The fix is to rasterise the **composed SVG** at its `viewBox` dimensions rather than lift the embedded payload.

## Minor

- `modernizing-mission-critical-e-commerce-integrations-with-kestra/icon.webp` is added but unreferenced — that story has no `logoIcon:` in its frontmatter, which is why its `icon.svg` was correctly flagged as an orphan. The commit that deletes orphans introduces a new one.
- `npm run svgo:check` isn't wired into any workflow, so nothing enforces it. Worth adding to `astro-check.yml`.
- `MIN_BYTES = 5 KB` means the check can't catch drift in files under 5 KB — fine, but worth a line in the script's docstring since the description presents it as covering "a file".

## What I checked and found clean

The other 162 deletions are genuinely unreferenced — I verified every deleted path against explicit imports, absolute string paths, SCSS `url()`, template-literal paths (`/icons/${…}.svg`, `/meta/plugins/…`), and every `import.meta.glob` consumer in the repo. The `teamsImages.ts` glob widening to `webp` and the corresponding `Investors.astro` / `NewsGrid.astro` key updates are all correct and complete.

Given the review effort already in this branch, splitting the SVGO and `<pattern>` commits into their own PR would let roughly 85% of the win land now while the WebP pass is redone.

elevatebart:
Both findings confirmed and acted on in `0e778bd`. Thank you for catching them — neither would have surfaced from CI.

## 1. The 13 `/vs/*` logos

Confirmed exactly as described. I checked `${slug}-img.svg` and `${slug}-monograme.svg` when building the held-back set and missed the third lookup: `getRawSvg(logoFile ?? slug)` resolving a bare `${slug}.svg`. Because it returns `""` rather than throwing, the build stayed green.

Reproduced against both refs: 13 of 38 slugs miss their logo on this branch, 0 on `main`. All 13 restored, and all 38 now resolve.

You're right that the basename grep is the gap. The rule I should have applied: a directory read through `import.meta.glob` with a runtime string match has no static reference to find, so nothing in it can be judged unreferenced without enumerating the key space of every consumer. I had that rule for `logoAsset` and `logoFile` and applied it incompletely.

## 2. The WebP conversion — reverted in full

Your diagnosis is right and the scope is worse than 20 of 23. I checked all 23: **every one** composes its payload through a `viewBox` crop, and most add masks, clipPaths, filters, blend modes or backing rects. The premise in my commit message, "no vector content at all", is false for the entire set. The `teams/*` avatars are a `0 0 50 51` viewBox over a 400×400 payload — a circular crop replaced with the full photo.

Your `hero-bg-effects` / `see_how` example is the proof, and I verified it: shared payload `md5 80f540ad…`, different composition (1720×818 vs 600×600, different clips and rects), and byte-identical output `md5 e3ba3534…`.

The verification was structurally incapable of catching this. I diffed the extracted payload against its own re-encoding — comparing the converter's input to its output, which is circular. For the dot meshes and the grids I rasterised the composed SVG before and after, which is why those hold up. I used the weaker method precisely where the transform was lossy.

Kept from that commit: the 11 unreferenced files deleted alongside it, and the three where the SVG wrapper survived and only the payload was re-encoded (`isai-new`, `axeleo-new`, `west-bend`) — correct for exactly the reason you identified.

Redoing it properly means rasterising the composed SVG at its `viewBox` size, and deciding per file whether a 24px icon or a 50px avatar should be a raster at all. Separate PR.

## Minor items

- `modernizing-mission-critical-e-commerce.../icon.webp` is gone with the revert.
- `npm run svgo:check` not wired into `astro-check.yml`, and the `MIN_BYTES` caveat in the docstring: both fair. Happy to add them here or leave them for the follow-up, whichever you prefer.

## What remains

SVGO, the dot-mesh and grid `<pattern>` work, and the deletions you verified. `main` → branch is 110.77 MB → 11.55 MB. Ready for another look.

---

iitzIrFan (approved):
Approving. Both blockers are resolved in `0e778bd`, and I re-verified everything from scratch against the new head rather than trusting the revert.

## 1. The 13 `/vs/*` logos

`src/components/versus/assets/` is now a **byte-identical file set to `main`** — all 13 restored, nothing else in the directory touched. Resolving `logoFile ?? slug` the way `vs/[slug].astro` does across every entry in the `vs` collection: **38 of 38 resolve**. The two without a slug-named file are `ansible-automation-platform` and `rundeck`, both carrying an explicit `logoFile:` override, exactly as on `main`.

## 2. The WebP conversion

Reverted in full, with no residue:

- Zero `.webp` files anywhere in the diff — all 23 gone
- All 36 reverted SVGs are byte-identical to `main` (checked by blob hash, not just presence)
- Every source import restored: `cloud/Header.astro`, `common/GetStarted.astro`, `common/SeeHow.astro`, `company/Investors.astro`, `company/NewsGrid.astro`, `get-started/Hero.astro`, `layout/Explore.astro`, `plugins/versions/OverviewPanel.vue`, both `code-in-any-language` pages, `use-cases/software-providers.astro`
- `src/utils/teamsImages.ts` back to `main` exactly — the `webp` glob widening is gone
- All five customer-story `icon.svg` / `logo.svg` restored and all four `index.md` frontmatter changes reverted

Keeping `isai-new`, `axeleo-new` and `west-bend` is the right call — those are the three where the wrapper survived and only the payload was re-encoded, which is why they render correctly.

Your note about the verification being circular — diffing the extracted payload against its own re-encoding — is the useful takeaway here, and it explains cleanly why the dot meshes and grids hold up while this didn't.

## What remains, re-verified

- **Deletions: 175 → 139.** All re-checked against explicit imports, absolute string paths, SCSS `url()`, template-literal paths, and every `import.meta.glob` consumer in the repo. One glob'd directory still has deletions (`src/assets/teams`, 6 files) and each has zero references.
- **SVGO and `<pattern>`: untouched by the revert.** 238 changed SVGs, 0 unparseable. I re-ran the native-resolution render diffs and got numbers identical to my first pass — `cicd/bg` 1.73% of pixels >48, `gridbg` max delta 50, `microservice-management` max 26, all confined to antialiasing.
- **Size: 110.77 MB → 20.15 MB**, matching the updated description.
- CI green on build, lint/astro-check, CodeQL and the Cloudflare publish.

## Two non-blocking notes

- Your comment above says `110.77 MB → 11.55 MB`; that's the pre-revert figure. The PR description is correct at 20.15 MB.
- On `svgo:check`: I'd wire it into `astro-check.yml` in this PR rather than the follow-up. The script lands here, and an unenforced check tends to stay unenforced. The `MIN_BYTES` docstring line can ride along or wait — no strong view.

Nice work on the turnaround.

PR #5672, opened 2026-09-16
perf: halve the /blogs LCP, cut 44 KB of render-blocking CSS, benchmark real images

elevatebart:
Four commits, each measured.

## 1. `/blogs` LCP (0cb3456)

The LCP element on `/blogs` was never blog content. It is the hero `section.mesh` background, `header-grid.svg`, and a CSS background is invisible to the preload scanner: the fetch could not start until the render-blocking stylesheet had arrived and layout had run. `HeroGridPreload.astro` puts a `<link rel="preload" as="image" fetchpriority="high">` in the head so the request goes out with the document. Lighthouse now passes all three LCP-discovery rows on the page (priority hinted, discoverable in the initial document, not lazy).

The page also shipped every blog twice. The `BlogGrid` island serialised all 202 entries into its `props` attribute (184 KB raw, 31 KB brotli, **47% of the compressed document**) carrying `description`, JSON-LD `schema`, `faq` and full author profiles that `BlogCard` never reads. It then rendered all 206 cards so CSS could hide 194 of them. The island now gets the six fields the card uses and renders `visibleCount` of them.

Lighthouse mobile throttling (1.6 Mbps, 150 ms RTT, 4x CPU), medians over interleaved runs:

| | document (br) | FCP | LCP |
| --- | --- | --- | --- |
| `/blogs` before | 65 KB | 1212 ms | 1748 ms |
| `/blogs` after | 42 KB | 796 ms | **852 ms** |
| `/blogs/<post>` before | 46 KB | 1332 ms | 1728 ms |
| `/blogs/<post>` after | 36 KB | 1016 ms | **1084 ms** |

Island props 184 KB -> 91 KB raw. Cards in the HTML: 206 -> 12. Isolating the halves: preload alone 1748 -> 1180 ms, payload trim alone 1748 -> 1348 ms. The preload also goes on the other three pages built on `Mesh`: `blogs/[...slug]`, `resources/[tag]`, `videos/[slug]`.

## 2. The render-blocking stylesheet (cbb6597)

`layout.css` was 338 KB raw on every page, and `/blogs` used 8.8% of it.

`layout.scss` called `@include markdown-styles` twice, once for `.bd-content` and once for `.bd-markdown`, emitting the same ~16 KB of rules under two selectors. One grouped selector emits them once; the two `.bd-*`-specific bits keep their own blocks and their place in the cascade.

`vanilla-cookieconsent`'s 27 KB stylesheet came in through `vendor.scss` on every page although only EU visitors on their first page ever see the consent UI. The consent script now injects it at runtime and awaits the load before building the modal, so it cannot flash unstyled. The `#cc-main` overrides moved from `app.scss` into the same file, after the vendor rules, so the `--cc-*` variables still resolve to the branded values (verified: the Accept button computes to `#631bff` on white).

It is imported as `?url` and attached with a `<link>` rather than imported as a module, because Astro hoists a script's CSS dependencies into the page head, dynamic imports included. The first attempt did exactly that and just renamed the blocking file.

`layout.css` 338.6 -> 294.6 KB raw, 45 -> 40 KB brotli. On `/blogs` the stylesheet lands 120 ms earlier (1031 -> 911 ms) and LCP follows it down (1164 -> 1040 ms).

## 3. The benchmark was measuring originals (11c3547, a7ba981)

The first Lighthouse run on this PR still showed `/blogs` at LCP 3.46 s, against ~340 ms of observed LCP subparts. The cause is the harness: it builds with `NO_IMAGE_OPTIM=true`, which sets the image service to passthrough, so `getImage({format: "webp", width: 800})` returns the source file. The first blog card, the one the grid loads eagerly, was a **4.9 MB PNG**; under the simulated 10 Mbps that single fetch starves the paint. In production the same card is **17 KB** through Cloudflare's image service.

`IMAGE_OPTIM_AT_BUILD` resizes with sharp at build time and leaves the runtime on passthrough, which is all `NO_IMAGE_OPTIM` was doing for this job. The benchmark run on this commit is the first one measuring real images:

| | performance | LCP |
| --- | --- | --- |
| Blog Index | 76 -> **98** | 3.46 s -> **1.07 s** |
| About Us | 84 -> **96** (🟢 +12) | 2.87 s -> **1.20 s** (a 5 MB `team.png`) |

Preview deploys keep `NO_IMAGE_OPTIM` and are untouched: verified, still passthrough, still a 2m34s build. Only the benchmark pays, and only for images: ~12k sharp transforms, ~9 min on a cold cache and ~0 on a warm `node_modules/.astro/assets`, which this workflow does not cache (deploy.yml drops that directory deliberately to stay inside the 10 GB quota).

The build job's timeout goes 20 -> 25 minutes, against the 14m12s the first real run took. Enough for a runner half again as slow, short enough that a wedged build still fails instead of holding a runner.

## Two trades worth a second opinion

- The 194 `display:none` blog links are no longer in the HTML, so crawler discovery of blog posts rests on `sitemap/blogs.xml`. Happy to render the overflow as bare `<a>` links instead.
- Moving the markdown CSS out of the global bundle entirely would save another 16 KB, but `.bd-markdown` is needed on every page: `AiChatDialog` is statically imported by `Search.vue`, which is `client:load` site-wide and mounts the dialog into a hidden modal on load. That needs the dialog to become genuinely lazy first, which is a change to the AI path rather than to a stylesheet import.

## Not in this PR

`bootstrap/utilities/api` is 76 KB of what remains in `layout.css`, `bootstrap/forms` 22 KB, `@kestra-io/topology` 11 KB. Trimming the `$utilities` map is the single biggest remaining win and the one with real regression surface.

## Checks

`astro check` 0 errors, 369 unit tests pass, oxlint clean, `/docs` re-rendered and spot-checked against computed styles. The visual suite was not run locally: this sandbox cannot reach Google Fonts, so the build falls back to system fonts and every snapshot would differ on text metrics for reasons unrelated to the diff. CI runs it with real fonts.

https://claude.ai/code/session_014XxNg7KHEiBhQAuC1EQmbb


iitzIrFan:
Thanks Bart, went through the PR and checked the main changes locally.

The `/blogs` payload reduction looks good. I also checked the CSS deduplication and the cookie-consent change, and both look consistent with the existing implementation. On the card cap — the overflow list keeps the same crawl posture as before, since those cards were already `display: none`, so that reads fine to me.

One small question: the inline `blogs` prop type in `BlogGrid.vue` only declares `path`, `title`, `category` and `date`, but the mapping also passes `image`, `author` and `authors`, which `BlogCard` renders. Since `BlogCard` types its own prop as `any`, nothing would flag it if one of those three got dropped from the mapping later.

Would it make sense to list all of them? Given the goal here is to narrow the island props, having the type cover everything that's actually passed would make that contract self-enforcing.

Other than that, nothing blocking from my side.

elevatebart:
Good catch, and it turned out to be worse than the type being incomplete: widening it does not enforce anything on its own.

`BlogList.astro` sits between the mapping and the grid with an untyped `Astro.props`, so `blogs` arrives at `BlogGrid` as `any` and satisfies whatever the prop declares. I tested it: with the type widened to all seven fields but `BlogList` left alone, deleting `image: blog.data.image` from the mapping still gives `astro check` **0 errors**.

So `87d2c4d` types both ends against a shared `BlogCardEntry` in `src/components/blogs/types.ts`, and gives `BlogList.astro` a `Props` interface. Same test now: **1 error**, pointing at the mapping.

One extra from checking the serialised props rather than guessing: `date` was declared `string` but Astro serialises it with its date tag (`[3, "2026-09-17T10:00:00.000Z"]`) and revives it as a `Date` in the island, so it is typed `Date` now. Nothing read it in `BlogGrid`, and `BlogCard` does `new Date(date)`, which is happy either way.

`externalNews` and `videos` on that interface are still `any[]` — they feed `BlogCard` and `LatestVideos`, which are outside what this PR touches.

---

iitzIrFan (approved):
LGTM!

PR #5679, opened 2026-09-16
fix(home): drop deprecated-only plugins from the plugin logo carousels

vfanucci:
The plugin logo carousels (`PluginsBlock`, `PluginsBox`, `EveryPlugin`) build `/plugins/<artifactId>` from `/plugins/metadata`. That endpoint still lists `plugin-notifications`, whose 35 elements are all deprecated since the split into per-channel plugins; its page has nothing to render and the router 301s it to `/plugins`.

The 16 Sept crawl counts 174 such links (home, `/features/*`, `/use-cases/*`, `/partners`, `/orchestration`, `/enterprise`).

Adds `fetchDeprecatedOnlyPlugins()` (plugins whose every element is deprecated, from `/plugins`) and filters those artifacts out of the three carousels, alongside the existing `excluded` list.


src/utils/plugins/deprecatedOnlyPlugins.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+    return new Set(
+        [...total.entries()]
+            .filter(([name, count]) => count > 0 && (live.get(name) ?? 0) === 0)
+            .map(([name]) => name),
+    )
+}
+
+export async function fetchDeprecatedOnlyPlugins(): Promise<Set<string>> {
+    const plugins = await $fetchApiCachedOptional<Plugin[]>("/plugins")
```
iitzIrFan, line 34: Non-blocking, but worth knowing: because the set is keyed on `name` from `/plugins` while the carousels filter on `artifactId` from `/plugins/metadata`, any metadata artifactId with no `/plugins` counterpart can never be flagged here.

There are five of those today: `script`, `storage-cloudflare`, `storage-seaweedfs`, `plugin-ee-openlineage`, and one `undefined`. Four are caught by the `excluded` regex — `script` isn't, and `https://kestra.io/plugins/script` 301s to `/plugins`. It's the one artifactId out of the 201 survivors that still redirects.

It only lands in the rendered set on a fraction of builds (random sort + slice), and it predates this PR, so fine as a follow-up. If you'd rather close it here it's a one-liner:

```
/^(core$|script$|plugin-kestra$|plugin-debezium$|plugin-ee|secret-|storage-)/
```

Note `PluginsBox.astro` spells the same exclusions out as a chain of `!==`/`startsWith` rather than using the regex, so it'd need the equivalent line too.

src/utils/plugins/deprecatedOnlyPlugins.ts
```diff
@@ -0,0 +1,36 @@
...
+import { isEntryAPluginElementPredicate, type Plugin } from "./plugin"
+
+/**
+ * Names of plugins whose every element is deprecated. The metadata endpoint
+ * still lists them (e.g. `plugin-notifications`, split into per-channel
+ * plugins), but their page has nothing left to render and the plugin router
+ * 301s it to the /plugins index — so anything linking to `/plugins/<name>`
+ * from that list produces a redirected link.
+ */
+export function collectDeprecatedOnlyPlugins(plugins: Plugin[]): Set<string> {
```
iitzIrFan, line 11: Consistency note, nothing failing today. `filterPluginsWithoutDeprecated` in `plugin.ts` already encodes this same "every element is deprecated" rule, and it's the one the router actually redirects on (`[...slug].astro` → `buildPluginPageProps` → `filterPluginsWithoutDeprecated`).

The two rules differ on plugins that expose no elements at all: `filterPluginsWithoutDeprecated` drops those (so the router 301s), while the `count > 0` guard here leaves them unflagged (so the carousel keeps linking). Zero plugins in the current payload hit this, so it's latent — but the two can drift as the payload changes.

Deriving the set from the existing helper — names in the input minus names it returns — would keep the carousel filter and the redirect decision in lockstep by construction.

iitzIrFan (approved):
Approving. I checked the new filter against the live API and the deployed site rather than just reading the diff:

- `collectDeprecatedOnlyPlugins` over the real `/plugins` payload flags exactly three plugins — `plugin-notifications`, `plugin-script-nashorn`, `plugin-script-jython` — and all three do 301 to `/plugins`. No false positives, so nothing that belongs in the carousels gets dropped.
- `/plugins` and `/plugins/subgroups` produce an identical result set, so the endpoint choice is fine.
- `plugin.name` in `/plugins` lines up with `artifactId` in `/plugins/metadata`, and all three flagged names exist as metadata artifactIds (21 metadata rows removed in total).
- Resolving all 201 artifactIds that survive the filter: 200 return 200. That's 173 of the 174 reported redirected links gone.
- Failure mode is right — `$fetchApiCachedOptional` returning nothing yields an empty set, so the carousels render unfiltered instead of breaking the page.

Two non-blocking notes inline: one remaining 301 the filter structurally can't catch, and a consistency point against `filterPluginsWithoutDeprecated`. Neither is worth holding the fix for.

PR #5680, opened 2026-09-16
fix(plugins): 301 bare element slugs instead of rendering a duplicate subgroup page

vfanucci:
`/plugins/<plugin>/<subgroup>/<bareName>` (e.g. `/plugins/plugin-ai/tool/stdiomcpclient`) is not a page: element pages end with the fully qualified class name. The router rendered the subgroup page there anyway, self-canonical, with every element card linking to `/plugins/plugin-ai/tool/stdiomcpclient/<fqcn>`, which 301s.

The 16 Sept crawl found 12 such pages under `plugin-ai/tool`, i.e. 144 redirected links plus 12 duplicate pages.

The router now 301s that URL shape to the element page when the bare name matches an element of the plugin, and to the subgroup page otherwise. Known pages are left untouched. Unit tests cover the three cases.


fdelbrayelle (approved):

fdelbrayelle:
Let's wait for @iitzIrFan review @vfanucci :pray:

iitzIrFan (approved):

iitzIrFan:
@vfanucci one follow-up for a separate issue — not blocking this PR.

The `if (!versionState.version)` guard exempts versioned URLs, and the same duplicate exists on them:

```
GET https://kestra.io/plugins/plugin-ai/v1.16.0/tool/stdiomcpclient
→ 200
→ <meta name="robots" content="index, follow, ...">
→ <link rel="canonical" href=".../plugin-ai/v1.16.0/tool/stdiomcpclient">
→ 12 links of shape /plugins/plugin-ai/v1.16.0/tool/stdiomcpclient/<fqcn>
```

Same shape as the bug this PR fixes — indexable, self-canonical, 12 redirected outlinks — but multiplied by the number of released versions rather than just the latest.

The guard is the right call here: `pageList` holds unversioned paths, so reusing `resolveShortElementSlug` as-is would 301 to a URL with the version segment dropped. A proper fix needs the version re-inserted into the target, which is more than this PR should carry.

Could you open an issue for it so it doesn't get lost?

vfanucci:
@iitzIrFan New issue :) https://github.com/kestra-io/docs/issues/5690

PR #5681, opened 2026-09-16
fix(docs): resolve ./index.md and ../index.md links without a trailing slash

vfanucci:
Two relative-link shapes in docs sources resolved to a directory URL with a trailing slash, which the site 301s:

- `./index.md` from a leaf file (`api-reference/03.problems/*.md` → `/docs/api-reference/problems/`, 35 links)
- `../index.md` from an index file (`triggers/`, `migration-guide/v2.0.0/`, `task-runners/`, `scripts/`, `cicd/`, `rbac/`, 6 links)

The remark link replacer moves out of `astro.config.mjs` into `src/markdown/remark/resolve-doc-link.ts` and resolves those two shapes to the directory segment explicitly (`../problems`, `../triggers`). Other relative links keep their previous resolution, which the tests pin.


iitzIrFan (approved):
Approving — the resolution logic checks out.

I ran the old `astro.config.mjs` replacer and the new `resolveRelativeDocLink` side by side over every relative markdown link in `src/contents/**` (~1500 links, all collections, since the markdown processor is global). Exactly 41 outputs change: the 35 `./index.md` links under `api-reference/03.problems/` and the 6 `../index.md*` links in index files. Everything else resolves identically, so no collateral change to blogs/orchestration/resources.

I then resolved each rewritten href against its own page URL (computed with the same `generateId` the content loader uses) and matched it against the set of doc ids — all of them land on an existing page, including the two that could have gone wrong in `generateId`: `v2.0.0` survives intact (`../v2.0.0`) and `16.scripts` loses only its ordering prefix (`../scripts`).

The `climb = ups + 1` for leaf files vs `ups` for index files is right, since `generateId` only strips ordering prefixes and never collapses a segment — URL depth mirrors filesystem depth. `astro sync` loads the config fine, `path` is still used after the `generateId` import was dropped, and the new test is picked up by vitest's `src/**/*.test.ts` include.

---

One non-blocking nit, in content rather than in this diff:

`src/contents/docs/05.workflow-components/07.triggers/05.realtime-trigger/index.md:11`

```md
Most Kestra [triggers](./index.md) poll external systems at a fixed interval
```

`./index.md` from an index file means "this directory's index", so it resolves to `realtime-trigger` — the page linking to itself. The prose means the parent triggers page, so it should be `../index.md`, which this PR now resolves to `../triggers`.

It's a pre-existing typo and the preserved `./index.md`-from-an-index-file branch is correct as written (the test at `resolve-doc-link.test.ts:44` pins the right behaviour). But since this PR is about making index links land on the right page, it's the one index link in the corpus still pointing at the wrong one — a natural one-line addition here, or a follow-up if you'd rather keep the diff to the resolver.

iitzIrFan:
Concrete version of the nit from my review, if you want to fold it in — and a correction: I wrote `../index.md` above, but the parent index is `index.mdx`, so it's `../index.mdx`.

`src/contents/docs/05.workflow-components/07.triggers/05.realtime-trigger/index.md:11`

```diff
-Most Kestra [triggers](./index.md) poll external systems at a fixed interval, which works well for batch processing.
+Most Kestra [triggers](../index.mdx) poll external systems at a fixed interval, which works well for batch processing.
```

That matches what the sibling page already does — `06.mcp-tool-trigger/index.md:88` links to `../index.mdx#trigger-common-properties`.

| | link | resolves to |
|---|---|---|
| now | `./index.md` | `realtime-trigger` → `/docs/workflow-components/triggers/realtime-trigger` (itself) |
| after | `../index.mdx` | `../triggers` → `/docs/workflow-components/triggers` |

Worth noting the fix depends on this branch: on `main`, `../index.mdx` from an index file resolves to `.` — the 301 you're removing here. So it only lands correctly once this PR is in, which is why it seemed a natural addition rather than a separate commit. Still non-blocking either way.

Issue #5611, opened 2026-09-14
/orchestration cleanup: API-only blueprints, canonical docs links, content-owned categories


iitzIrFan:
### Why

Team flagged "trash code" on the orchestration pages, with the blueprint handling as his example. An audit on 2026-09-14 confirmed the issues below against the repo, git history (`803af4739`, `ac97b6e91`), the live API and rendered pages.

### Files involved

| Role | Path |
|---|---|
| Per-tool template | `src/pages/orchestration/[slug].astro` |
| Catalog page | `src/pages/orchestration/index.astro` |
| Blueprint section | `src/components/orchestrate/OrchPattern.astro` |
| Schema | `src/schemas/orchestration.ts` |
| Content | `src/contents/orchestration/*.yaml` (38 files) |
| Existing utils to reuse | `src/utils/fetch.ts` (`$fetchApiCachedOptional`), `src/utils/plugins/taskUrls.ts` (`buildTaskUrls`), `src/utils/blueprints/*.ts` + `*.test.ts` as the pattern |
| Consumer to protect | `src/pages/blueprints/[id].astro:198` (regex-parses the YAML `tool:` block) |

---

### PR 1: blueprints come only from the API

**Problem.** `gemini.yaml` item 3 ("Multimodal extraction") has `blueprintPlaceholder: true` + inline `code:` + `plugins:`. The blueprint now exists in the API as `gemini-multimodal-receipt-extraction` and the inline copy has drifted (model names, trigger credentials, payload). The page also does its blueprint fetching inline with hand-rolled try/catch (`[slug].astro:118-149`) even though the repo pattern is a util in `src/utils/blueprints/` using `$fetchApiCachedOptional`. Commit `803af4739` ("remove fallback for blueprint and trust API") removed inline code from 20 files in May but left `code`/`plugins` optional in the schema, which let this come back.

**Tasks**

- [ ] `src/contents/orchestration/gemini.yaml`: on item 3 set `blueprintId: "gemini-multimodal-receipt-extraction"`, delete `blueprintPlaceholder`, `plugins`, `code`.
- [ ] `src/schemas/orchestration.ts`: in `blueprintSchema` remove `code`, `plugins`, `flowDiagram`, `blueprintPlaceholder`; make `blueprintId` required. Delete `flowNodeSchema` and `flowDiagramSchema` (no consumers, 0 YAML uses).
- [ ] New `src/utils/blueprints/orchestrationBlueprints.ts`:
[...]

PR #5683, opened 2026-09-16
fix(orchestration): derive the plugin docs link from the catalog

vfanucci:
`/orchestration/<slug>` guessed the plugin docs link as `/plugins/plugin-<4th class segment>`, which gives `/plugins/plugin-jdbc` for `io.kestra.plugin.jdbc.duckdb.Query` and `/plugins/plugin-microsoft` for the Fabric class. Neither is a page: both 301 to the `/plugins` index (5 links in the 16 Sept crawl: duckdb, postgres, snowflake, sql-server, microsoft-fabric).

The link is now derived from the catalog through `buildTaskUrls` (group segment of the element's canonical URL), with the previous guess kept as fallback when the class is unknown.


src/pages/orchestration/[slug].astro
```diff
@@ -162,13 +163,18 @@ try {
     console.warn(`Could not fetch plugin icon for ${tool.pluginClass}`, e)
 }
 
+// Resolve the plugin group page from the catalog: guessing `plugin-<segment>`
+// from the class name yields `/plugins/plugin-jdbc` for `io.kestra.plugin.jdbc.duckdb.Query`,
+// a page that only 301s to the /plugins index (the artifact is `plugin-jdbc-duckdb`).
+const canonicalTaskUrl = (await buildTaskUrls([tool.pluginClass]))[tool.pluginClass]
```
Comment (iitzIrFan) [inline src/pages/orchestration/[slug].astro:169]: nit: this line is not formatted per the repo's prettier config. `main` is clean, so this PR introduces the drift. CI only runs oxlint and astro check, which is why it passed. `npm run format` wraps it.

src/pages/orchestration/[slug].astro
```diff
@@ -162,13 +163,18 @@ try {
...
 const pluginParts = tool.pluginClass.split(".")
-const pluginDocsUrl =
-    pluginParts.length > 4 && pluginParts[3] === "ee"
-        ? `/plugins/plugin-ee-${pluginParts[4]}`
-        : pluginParts.length > 3
-          ? `/plugins/plugin-${pluginParts[3]}`
-          : null
+const pluginDocsUrl = canonicalTaskUrl?.startsWith("/plugins/")
```
Comment (iitzIrFan) [inline src/pages/orchestration/[slug].astro:171]: nit: `canonicalPluginUrl` only ever yields `/plugins/...` paths, so this guard is redundant. A plain `canonicalTaskUrl ? ... : fallback` reads the same.

More broadly, the repo keeps plugin URL logic in `src/utils/plugins/` with unit tests, and `canonicalUrl.ts` already has `canonicalPluginPath` computing exactly the plugin page path. A small `buildPluginPageUrls(classes)` next to `buildTaskUrls` would keep this `split/slice/join` out of the page and give it a test. Happy to leave as is for a one-line fix.

src/pages/orchestration/[slug].astro
```diff
@@ -162,13 +163,18 @@ try {
...
+    ? canonicalTaskUrl.split("/").slice(0, 3).join("/")
```
Comment (iitzIrFan) [inline src/pages/orchestration/[slug].astro:172]: suggestion: `slice(0, 3)` drops the subgroup segment. 22 of the 38 tools live in multi-subgroup plugins, so BigQuery, Pub/Sub, GCS, Dataproc and Vertex AI all land on the generic `/plugins/plugin-gcp` root, while the catalog gives `/plugins/plugin-gcp/google-cloud-bigquery`, which resolves 200. Dropping only the last segment instead would link each tool to its most specific page for free:

```ts
canonicalTaskUrl.split("/").slice(0, -1).join("/")
```

Not a regression versus `main`, so your call on whether the plugin root is the intended landing page.

iitzIrFan (approved):
Verified locally against the branch on a dev server: all 38 `/orchestration/<slug>` pages now render a docs-card link that returns 200 with no redirect. The five formerly broken ones (duckdb, postgres, snowflake, sql-server, microsoft-fabric) resolve to `plugin-jdbc-duckdb`, `plugin-jdbc-postgres`, `plugin-jdbc-snowflake`, `plugin-jdbc-sqlserver` and `plugin-microsoft-fabric`. Terraform, gemini and netbox are unchanged, and the old guesses `/plugins/plugin-jdbc` and `/plugins/plugin-microsoft` still 301, so the premise holds.

This is "PR 2" of #5611. Could you add `Part of #5611` to the description? Parts 1 and 3 are still open, so it should not close the issue.

Two nits and one suggestion inline. None blocking.

FYI, unrelated to this change: `docker.yaml` declares `io.kestra.plugin.docker.Build`, which the catalog no longer has (it is `io.kestra.plugin.docker.cli.Build` now). Docker is the one page that hits the fallback branch. The link still works, so I will file it separately.

PR #5685, opened 2026-09-17
perf(assets): WebP conversion redo, plus the svgo:check gate

elevatebart:
Two pieces of follow-up to #5651, from @iitzIrFan's review there. #5687 merged into this branch, so both now land together.

**SVG: 20.15 MB → 9.23 MB.** WebP grows 22.51 MB → 24.24 MB (+1.73 MB), so ~9.2 MB net off the repo.

## 1. WebP conversion, redone

#5651 reverted a conversion that lifted the base64 payload out of each wrapper and re-encoded it. Every one of those files composes that payload — a `viewBox` crop, usually masks, clipPaths, filters or blend modes — so lifting it discarded the composition. A `0 0 50 51` avatar crop became the full 400×400 photo.

Each of the 36 files is now rasterised **from its composed SVG** in Chromium, at 2× its `viewBox` (capped at 2400px on the long side), alpha preserved so anything compositing behind it still does. The browser resolves the composition, so nothing can be silently dropped.

The old collision doubles as a regression check: `hero-bg-effects.svg` and `see_how.svg` share one embedded payload and previously produced byte-identical WebP. They now render 2400×1141 and 1200×1200.

### Verification

Compares the **original SVG rendered at display size** against the **shipped WebP scaled into the same box**.

That is precisely the check the first attempt lacked — it diffed the extracted payload against its own re-encoding, the converter's input against its output, which is circular and structurally incapable of seeing a discarded wrapper.

| | |
|---|---|
| Worst file | 1.5% of pixels differ by >32 |
| Median | 0.2% |
| Above 2% | none |

Quality is per file, the lowest that holds the comparison under budget, falling back to lossless. Most landed on lossless: logos and UI art at this size cost more in lossy artefacts than they save.

### Six candidates excluded

| File | Why |
|---|---|
| `teams/isai-new.svg` | inlined with `?raw` + `set:html`; cannot be a raster |
| `orchestrate/west-bend.svg`, `teams/axeleo-new.svg` | already carry a WebP payload in a surviving wrapper, and are smaller as they are |
| copines-de-voyage `icon`/`logo`, `teams/techcrunch.svg` | 5–17% of pixels differ. Content is right and it's resampling softness at ~50px, but they're 196 KB of 11 MB and not worth asserting an equivalence I can't cleanly show |
| `public/og-bg.svg` | feeds a `foreignObject` template in `ogImage.ts` |

## 2. `svgo:check` wired into CI

Added as a hard gate in `astro-check.yml` next to the unit tests. Nothing under `public/` reaches Astro's `experimental.svgOptimizer`, so without this nothing notices an SVG landing unoptimised.

Running it locally failed on 26 files, which is the useful part: #5651's revert restored them from `main`, but its SVGO batch had already run against a tree where the WebP commit had deleted them, so they came back unoptimised with nothing to re-process them. Optimised here. Of those 26, 22 are superseded by the conversion above and are now deleted.

`svgo` is also pinned to an exact `4.1.0`: this PR makes the tool's output a gate, so a floating range would let a minor bump silently redefine "optimised" and red every open PR at once.

## Checks

`astro check` 0 errors across 503 files, `svgo:check` exits 0, no stale `.svg` references to converted files, every rewritten reference resolves.

References were rewritten by resolving each token the way the bundler would (`~/`, `./`, `/`, bare sibling names inside `src/contents`) rather than by basename — plus the `TeamsImages["x.svg"]` string keys and the `teamsImages.ts` glob.

## Worth a look on the preview

`/cloud`, `/get-started`, `/about-us` (investor grid), and any customer story with a logo — those are the surfaces the reverted conversion broke.

https://claude.ai/code/session_01XPuURKJjvfJhd3gq2ko1qu


iitzIrFan (approved):
Approving. Small, self-consistent, and I verified the premise rather than taking it on trust.

## The premise reproduces exactly

Ran the check against merged `main` and against this branch:

```
main:   211 files, 19.28 MB -> 19.26 MB
        26 file(s) are not optimised     exit 1
branch: 210 files, 19.25 MB -> 19.25 MB  exit 0
```

The 26 files it names are **precisely** the 26 this PR changes. Your explanation holds: #5651's SVGO batch ran against a tree where the WebP commit had already deleted those files, so the revert restored them in their unoptimised `main` state with nothing left to re-process them. Running the gate before wiring it up — and finding it red — is the right order of operations.

## The diff is exactly what the script produces

Ran `npm run svgo` on a clean checkout of merged `main` and compared file by file:

**26 of 26 byte-identical to this PR, 0 differing**, and it touched exactly those 26 — no more, no less. So the diff is mechanical output of the committed script, not hand-tuned.

## Renders

All 26 rasterised in Chromium from the composed SVG at native `viewBox` size, before and after: **9 pixel-identical, 17 differ only at edges, 0 structural.** Most maxima are 1–6/255. The two largest, zoomed with an amplified diff:

- `versus/assets/simplyask.svg` — at 8× the diff is a single-pixel outline tracing each glyph and the logomark. Nothing filled, shifted or dropped; invisible at its 127×26 render size.
- `cloud/assets/cloud.svg` — amplified diff is near-black with a faint mask-edge trace.

My 9/17 split differs from your 23/3 only because antialiasing counts are render-scale dependent — you rasterised at display size, I used native `viewBox`. Same conclusion.

## CI wiring

Confirmed the step actually executed rather than just that the job was green: job `105103721466`, step 12 `Check SVG optimisation` → success. Placement is right — after `npm ci`, after the reviewdog report so lint feedback isn't suppressed, and `if: ${{ !cancelled() }}` matching the neighbouring unit-test step. The `MIN_BYTES` docstring note is there too.

## Three non-blocking notes

1. **`svgo` is `^4.1.0`.** CI is deterministic through the lockfile, but a Dependabot minor bump can change SVGO's output and turn *every* open PR red until someone runs `npm run svgo`. Either pin it exact, or have the svgo bump PR run `npm run svgo` as part of the bump.
2. The workflow's header comment enumerates what the job runs; the SVG check isn't in that list.
3. Pre-existing, not yours: `astro-check.yml` runs `npm run test:unit` twice (lines 32 and 112, both already on `main`). Free cleanup while you're in the file, but out of scope here.

---

With this merged, the three PRs together take `public/` + `src/` SVG from **110.85 MB across 727 files to 20.13 MB across 588**. The part that reaches visitors: `/use-cases/disaster-recovery`, `/monitoring` and `/provisioning-and-deployment` each shipped ~16.5 MB of background texture and now ship under 80 KB. And `monitoring/quick-alerting.svg` alone went from **46,300 elements and 11,554 byte-identical gradient definitions to 99 elements and 1**.

Worth noting the Lighthouse benchmark doesn't cover a single use-case page, so the largest win on the site is currently invisible to it. Adding those eight pages to the page list looks like a cheap follow-up.

elevatebart:
Took notes 1 and 2 in `1c7150a`. Sorry for costing you the approval on non-blocking items — note 1 seemed worth fixing in the PR that creates the exposure rather than after it.

**`svgo` pinned to `4.1.0`.** Your reasoning is the deciding part: this PR turns the tool's output into a gate, so a floating range means a minor bump silently redefines "optimised" and reds every open PR at once, with the cause a version range rather than anything in the diff. Pinning converts that into one explicit bump PR that runs `npm run svgo` alongside it. Two lines, `package.json` and `package-lock.json`.

Worth flagging a trap I hit doing it: `npm pkg set devDependencies.svgo` **adds** a second declaration rather than editing the existing one — `svgo` lives in `dependencies` here, so I briefly had it declared twice. Corrected; it's declared once, and `npm ci --dry-run` resolves clean.

**Workflow header comment** now lists the SVG check.

**Note 3** left alone as you suggested — `test:unit` at lines 32 and 112 is pre-existing on `main` and unrelated to this diff.

On your closing point: agreed, and it's a real gap. The benchmark samples 18 routes and not one use-case page, so the biggest win on the site — `quick-alerting.svg` going from 46,300 elements and 11,554 gradient definitions to 99 and 1 — is invisible to it. Those eight pages are in the 42 routes added to #4588 for the screenshot suite; adding them to `PAGES` for Lighthouse is a different call, since that list is deliberately small and each entry costs benchmark time. Happy to open it as its own PR if you think the eight are worth the runtime.

---

iitzIrFan:
Re-verified `1c7150a`. Both notes addressed, and the change is as small as it can be.

`svgo` is declared exactly once, at `4.1.0`, in `dependencies` where it already lived — the duplicate you hit with `npm pkg set` isn't present. package.json and the lockfile agree, and `npm ci` passed in CI, which is the real proof there. Worth noting the lockfile diff is a single line: the resolved tree is unchanged, so `npm ci` installs exactly what it did before. No nested copies either — Astro 7.3.1 (`^4.0.1`) and `@iconify/tools` 5.0.14 (`^4.1.0`) are both satisfied by the pin.

One thing to know for later, not to act on: because Astro also depends on `svgo`, an Astro bump that moves to `svgo ^5` will quietly nest a second copy rather than fail. Harmless — `svgo-batch.mjs` resolves the root one — but it means the pin won't block an Astro upgrade, it'll just stop deduping.

No SVGs touched by this commit, `svgo:check` still passes, full CI green. Approval stands. Nice discipline keeping the pin in the PR that creates the exposure rather than after it.

On the Lighthouse page list: agreed it's a separate call, and your reasoning on runtime cost is fair. I'd leave `PAGES` as is — the screenshot suite in #4588 covering those 42 routes is the better place for use-case coverage, since what we want there is visual regression, not scores. Not worth its own PR unless someone asks for the numbers.

PR #5686, opened 2026-09-17
fix(marketing): replace hardcoded plugin counts with dynamic count

fdelbrayelle:
## Summary
- `/enterprise` and `/pricing` (via `PriceHeader`) plus 8 other marketing pages/components each hardcoded a frozen plugin count (1200/1400/1700/1800/2000+) instead of using `fetchTotalPluginsCount()`, which already exists in `src/utils/plugins/pluginCount.ts` and is what `PluginsBlock.astro` correctly uses.
- Counts had drifted out of sync across the site depending on which PR last touched each page (see files below).
- Wired all of them to the same live source so the number stays correct as plugins are added.

Files fixed:
- `src/components/price/Header.astro` (pricing page)
- `src/components/common/SchemaSoftwareApplication.astro` (SEO schema description, shared default across pages incl. enterprise/pricing)
- `src/components/home/Hero.astro`, `src/components/twozero/Hero.astro`, `src/pages/two-zero.astro`
- `src/pages/cloud.astro`
- `src/components/lp/workflow-orchestration/Platform.astro`, `Stack.astro`
- `src/pages/use-cases/public-services.astro`, `software-providers.astro`

Out of scope: dated blog posts and the whitepaper under `src/contents/blogs` / `src/contents/resources` (historical content reflecting the count at time of writing), and one docs MDX page (`src/contents/docs/ai-tools/index.mdx`) since content-collection MDX doesn't currently have a pattern for build-time dynamic values.

## Test plan
- [x] `npx astro check` — 0 errors across 503 files
- [ ] Visual check of `/pricing`, `/enterprise`, `/`, `/cloud`, `/two-zero`, `/use-cases/public-services`, `/use-cases/software-providers` after deploy preview

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


iitzIrFan:
Thanks for catching the `/two-zero` ones — heads up that this overlaps #5446 ("single source of truth for plugin and blueprint counts"), which has been open since Aug 30 and is in review.

**8 of the 10 files here are already in #5446** with the same change: `SchemaSoftwareApplication.astro`, `home/Hero.astro`, `lp/workflow-orchestration/{Platform,Stack}.astro`, `price/Header.astro`, `cloud.astro`, `use-cases/{public-services,software-providers}.astro`.

The two genuinely new ones are `twozero/Hero.astro` and `pages/two-zero.astro` — the 2.0 page (#5553) landed after #5446 opened. I've picked those up on that branch in d4f7f11.

The reason for folding them in rather than merging this separately: #5446 rewrites `fetchTotalPluginsCount()`, and on current `main` it still reads

```ts
export async function fetchTotalPluginsCount(): Promise<number> {
    try { ... return Math.floor(calculateTotalPlugins(g) / 100) * 100 }
    catch (e) { console.error(...); return 0 }
}
```

So against `main` these conversions have two side effects:

1. **`0+ Plugins` becomes reachable again.** The `catch` returns `0`, so a blip on `/plugins/subgroups` renders `0+ Plugins` on `/pricing`, `/`, `/cloud`, `/two-zero` and inside the shared `SoftwareApplication` JSON-LD. That's #5387, which #5446 closes with a `PLUGIN_COUNT_FLOOR = 2000` fallback. Today's hardcodes are stale but never zero, so this trades stale for occasionally-zero in copy and structured data.
2. **The thousands separator is lost.** `main` returns a bare number, so `/pricing` goes `1,200+` → `2000+` and `/two-zero` goes `2,000+` → `2000+`. #5446 centralises `formatPluginCount()` so every surface reads `2,000+`.

#5446 also covers the SSR paths that currently render `0+` (`enterprise/Dashboard.vue`, `onezero/Header.vue`, `AiCopilot.vue`, `EveryPlugin.astro`) and ~105 hardcoded totals across the `vs`/`orchestration` YAMLs.

Both branches touch the same lines in those 8 files, so whichever lands second will conflict. Suggest closing this one as superseded — but happy to be wrong if you'd rather land the quick fix first and I rebase #5446 on top.

iitzIrFan:
closing https://github.com/kestra-io/docs/pull/5446

PR #5705, opened 2026-09-18
test(visual): odiff comparison plus a diff report artifact

elevatebart:
Two changes, one problem: when the nightly snapshot refresh rewrites 40 baselines, nobody can tell what actually moved.

## Comparison now runs through odiff

`playwright-odiff` replaces the built-in screenshot comparison. Same snapshot folder, same `--update-snapshots=changed` behaviour, roughly an order of magnitude faster on a CI runner, and its anti-aliasing detection is less twitchy than pixelmatch.

Caveats worth knowing before review:

- It registers `toHaveScreenshotOdiff`, it does not override `toHaveScreenshot`. The `expect.toHaveScreenshot` block in `playwright.config.ts` therefore stopped applying, so `maxDiffPixelRatio: 0.01` moved to the call site. That config block is now a comment explaining why it is gone.
- It reads Playwright's private `_projectInternal`. Peer range is `^1.56.0` against our pinned `1.61.0`, fine today, worth re-checking on the next Playwright bump.

Verified locally against a scratch spec: writes missing baselines, tolerates a sub-1% change, fails on a 99% one, and rewrites the baseline under `--update-snapshots=changed`.

## The report

`scripts/visual-diff-report.mjs` (`npm run test:visual:report`) reads `git status` for the snapshot folder, pulls each baseline out of `HEAD`, and runs odiff against the rewritten file. Output is `visual-diff-report/`: `index.html`, `report.json`, and baseline/current/diff PNGs.

It uses `ODiffServer`, so 180 comparisons cost one process rather than 180 spawns. `failOnLayoutDiff: false` means a page that grew taller still gets a diff image instead of an opaque "layout-diff".

The HTML has no script tag. Radio inputs drive the per-card view switch (Diff / Slider / Before / After) and the global status filter; the before/after wipe is a `resize: horizontal` overlay whose inner image is pinned at `100cqw` so the two halves line up exactly.

Tested against five seeded changes: a small localised edit, a full-page hue shift, a 300px-taller page, one added snapshot and one deleted.

## Workflow wiring

`update-linux-snapshots.yml` builds the report and uploads it as the `visual-diff-report` artifact, and the generated PR body links to it with the change counts.

The report step runs **before** the commit step, and has to. The commit moves `HEAD` onto the new snapshots, at which point there is nothing left to diff against. There is a comment on it so nobody helpfully reorders it later.

## Two things to flag

The script does not produce a zip. `actions/upload-artifact` zips whatever you hand it, so shipping our own would put a zip inside a zip; the download is still `visual-diff-report.zip` either way. Easy to change if you would rather have the literal file.

Worked around an upstream bug: `ODiffServer.stop()` in odiff-bin 4.5.0 clears its `exiting` guard synchronously, before the async exit event lands, so a clean shutdown prints `odiff server exited unexpectedly with code null`. Two lines hold the guard open, with a comment naming the version.

## Ignore the Lighthouse regression markers here

The benchmark comment on this PR flags 🔻 on Get Started, About Us and Plugin Debug Return. None of them are real: this PR touches 7 files, all tooling, nothing under `src/` or `public/`, so the built site is identical to main.

This PR happens to be decent evidence for that, which is now filed as #5707. The two benchmark runs here bracket a single trailing newline added to a script that never executes at build time, so the measured site was byte-identical between them, and scores still moved by up to 22 points. The split falls exactly along the `runs` count in `tests/fixtures/page-sample.mjs`: single-run pages averaged a 6.1 point swing, `runs: 3` pages averaged 1.3.

https://claude.ai/code/session_01PStmR9iSDAvvDSV7rD6DyA


elevatebart:
`Lighthouse shard 3` failed ([job 107187403357](https://github.com/kestra-io/docs/actions/runs/35861260109/job/107187403357)). Not this PR's, and here is the working:

```
Runs     : 8, 1 per page except /docs/workflow-components/flow x3, /blueprints x3

Fatal error: Error: connect ECONNREFUSED 127.0.0.1:33357
```

Port 33357 is the Chrome DevTools port. The script's next log line, `Chrome launched on port …`, never printed, so this died inside `chromeLauncher.launch()` 25 seconds in, before a single page was measured. Chrome failed to come up on the runner.

Two reasons it is not attributable to this branch:

- The diff is 7 files, all tooling: a workflow, `.gitignore`, the two package files, `playwright.config.ts`, `scripts/visual-diff-report.mjs` and the visual-regression spec. Nothing under `src/`, `public/`, or the Lighthouse scripts.
- The same commit `25b8ebe` passed all four Lighthouse shards 90 minutes earlier, in [run 35844337939](https://github.com/kestra-io/docs/actions/runs/35844337939). This run was triggered by a force-push that removed an unrelated commit and left this SHA as the head.

There is no fix to port: a browser that does not start is not something a change here can address. Re-running the failed job is the right move, and I will, but the API currently refuses with `This workflow is already running` because the remaining shards and the report job are still going. I will re-run it once that finishes.

One thing worth a separate look, since it made this failure fatal rather than slow: `scripts/lighthouse-benchmark.mjs` wraps each Lighthouse pass in `runWithRetry` (two retries), but `chromeLauncher.launch()` has no retry at all, so a single failed browser start takes out the whole shard. Out of scope here.

---

.github/workflows/update-linux-snapshots.yml
```diff
@@ -147,6 +147,21 @@ jobs:
             echo "No Linux snapshot changes to commit."
           fi
 
+      # odiff compares each rewritten baseline against its committed version, so
+      # this has to run before the commit step moves HEAD onto the new snapshots.
+      - name: Build visual diff report
```
flcarre: **The report step now gates the commit and the PR.**

This step has no `continue-on-error`, and both `Commit and push updated Linux snapshots` and `Create PR with updated Linux snapshots` run after it. So any failure here throws away the entire refresh.

That is not hypothetical: `scripts/visual-diff-report.mjs` calls `server.compare(..., { timeout: 60_000 })`, and `ODiffServer` *rejects* that promise on timeout (`odiff-bin/server.js:207-214`). The rejection is not caught inside the `try` block, so the script exits non-zero. One slow comparison on a 2.8 MB full-page PNG — or an `ODiffServer` spawn failure — turns the nightly run red and the 40 regenerated baselines are never committed. They survive only in the `visual-snapshots-linux` artifact, which nobody is watching.

The step is a reporting nicety, not the workflow's product. Suggest:

```yaml
      - name: Build visual diff report
        id: diff_report
        continue-on-error: true
        if: steps.snapshot_changes.outputs.has_changes == 'true'
        run: npm run test:visual:report
```

The PR body already degrades gracefully — the two `steps.diff_report.outputs.*` interpolations just render empty.

flcarre (commented):
Reviewed the branch against `main` with the repo checked out, and verified the claims in the description against the actual sources — including unpacking `playwright-odiff@4.5.0` and `odiff-bin@4.5.0` from npm.

**What holds up.** The `ODiffServer.stop()` workaround is correctly diagnosed: `stop()` sets `exiting = true`, calls `cleanup()`, then resets `exiting = false` synchronously (`server.js:379-391`), all before the async `exit` handler at `server.js:111` runs — so holding the guard open is the right fix and the version note is accurate. `playwright-odiff` is published by the odiff author from the same repo, and `maxDiffPixelRatio` / `stylePath` are genuinely supported and typed (`dist/index.d.ts`). The `failOnLayoutDiff: false` reasoning checks out too — odiff still exposes `reason: "layout-diff"` in its result union, and the script handles both branches.

Seven findings below, the first three worth acting on before merge.

---

**Not inline — scope note**

Nothing blocking, but `tests/visual-regression.spec.ts-snapshots` is 186 MB across 180 files (largest single baseline 2.8 MB). See the comment on the upload step.

.github/workflows/update-linux-snapshots.yml
```diff
@@ -147,6 +147,21 @@ jobs:
...
+        id: diff_report
+        if: steps.snapshot_changes.outputs.has_changes == 'true'
+        run: npm run test:visual:report
+
+      - name: Upload visual diff report
+        if: steps.snapshot_changes.outputs.has_changes == 'true'
```
flcarre: **Artifact size on a large refresh.**

`tests/visual-regression.spec.ts-snapshots` is 186 MB across 180 files. The report copies baseline + current + diff uncompressed for every change, so a refresh that rewrites most baselines — the "40 baselines" case scaled up, e.g. after a font or global spacing change — puts `visual-diff-report/` near 550 MB, uploaded on top of the existing 186 MB `visual-snapshots-linux` artifact, both at `retention-days: 14`, in a job with `timeout-minutes: 75`.

Not a blocker at 40 changes (~120 MB), but worth a thought: downscaling the baseline/current copies, or shipping only the diff PNGs and building the slider against the snapshot artifact that is already uploaded, would cut this by two thirds.

.github/workflows/update-linux-snapshots.yml
```diff
@@ -192,6 +207,13 @@ jobs:
             Content dated after ${{ env.SNAPSHOT_CUTOFF }} was excluded from
             the build, and the sampled API responses were replayed from
             tests/fixtures/api.
+
+            ${{ steps.diff_report.outputs.report_total }} snapshot(s) changed,
```
flcarre: **This can read "0 snapshot(s) changed".**

`has_changes` is computed at line 140 over **two** paths:

```
git status --porcelain --untracked-files=all -- tests/visual-regression.spec.ts-snapshots tests/fixtures/api
```

but `changedSnapshots()` (`scripts/visual-diff-report.mjs:25-30`) only stats `SNAPSHOT_DIR`.

So when an upstream API response drifts and **only** `tests/fixtures/api` changes — precisely what the fixture replay exists to catch — this step still runs, writes `summary.total = 0`, and the generated PR reads "0 snapshot(s) changed, 0 differing pixels" above a link to an empty `index.html`.

Either scope the report to the same two paths, or set a separate snapshot-only flag in the `snapshot_changes` step and gate these three body lines on it.

playwright.config.ts
```diff
@@ -18,11 +18,8 @@ export default defineConfig({
         ? [["github"], ["html", { open: "never" }]]
         : "html",
 
-    expect: {
-        toHaveScreenshot: {
-            maxDiffPixelRatio: 0.01, // Allow up to 1% pixel difference for minor rendering variations
-        },
-    },
+    // No expect.toHaveScreenshot block: the suite uses toHaveScreenshotOdiff,
```
flcarre: **This comment is not correct — the block did still apply.**

`playwright-odiff` reads that exact config key as its option base:

```js
// dist/index.js:431
const configOptions = testInfo._projectInternal?.expect?.toHaveScreenshot || {}
```

and merges it *under* the call-site options (`SnapshotHelper`, `dist/index.js:76-79`), so `maxDiffPixelRatio: 0.01` would have kept working from here, with call-site values still winning.

Moving the tolerance to the call site is a defensible choice for locality. The problem is the stated reason: a maintainer who believes this comment and adds a second `toHaveScreenshotOdiff` call site elsewhere will silently inherit odiff's `DEFAULT_ODIFF_OPTIONS` (`maxDiffPixels: 0`, no ratio tolerance, `dist/index.js:23-28`) and get a suite that fails on a one-pixel anti-aliasing shift.

Either restore the config block and drop the per-call option, or reword to say the block still applies and the tolerance was moved deliberately.

scripts/visual-diff-report.mjs
```diff
@@ -0,0 +1,334 @@
...
+
+            let diff = null
+            if (status === "changed") {
+                const diffPath = join(outDir, "images/diff", file)
+                diff = await server.compare(
+                    join(outDir, "images/baseline", file),
+                    join(outDir, "images/current", file),
+                    diffPath,
+                    {
+                        threshold: 0,
```
flcarre: **The report compares more strictly than the suite that produced these diffs.**

`tests/visual-regression.spec.ts` passes only `fullPage` / `animations` / `timeout` / `maxDiffPixelRatio` / `stylePath`, so the matcher falls back to `DEFAULT_ODIFF_OPTIONS = { threshold: 0.2, antialiasing: true }` (`playwright-odiff/dist/index.js:23-28`). This report then recomputes every diff at `threshold: 0` with anti-aliasing detection off.

Consequence: a font-rasterisation shift the suite treated as within tolerance shows up here as a large "N px / X %" figure, and the aggregated `report_diff_pixels` in the PR body is not the number that gated the update. A reviewer reading "12,000 px differ" has no way to relate it to the 1 % tolerance that actually decided the baseline needed rewriting.

Either reuse the suite's comparison options here, or print both option sets in the report header so the numbers are interpretable.

scripts/visual-diff-report.mjs
```diff
@@ -0,0 +1,334 @@
...
+import { mkdirSync, copyFileSync, writeFileSync, rmSync } from "node:fs"
+import { basename, join } from "node:path"
+import { ODiffServer } from "odiff-bin"
+
+const SNAPSHOT_DIR = "tests/visual-regression.spec.ts-snapshots"
+const DEFAULT_OUT = "visual-diff-report"
+const DIFF_COLOR = "#ff0055"
+
+const args = process.argv.slice(2)
+const flag = (name, fallback) => {
```
flcarre: Minor: the fallback only applies when the flag is **absent**, not when its value is missing, so a trailing flag yields `undefined`.

`node scripts/visual-diff-report.mjs --out` gives `outDir === undefined`, and `main()` then calls `rmSync(undefined, { recursive: true, force: true })` → `ERR_INVALID_ARG_TYPE`. Same shape for a trailing `--dir` (git receives `undefined` as a pathspec) and `--base` (`git show undefined:path`).

```js
const flag = (name, fallback) => {
    const i = args.indexOf(`--${name}`)
    return i === -1 || args[i + 1] === undefined ? fallback : args[i + 1]
}
```

scripts/visual-diff-report.mjs
```diff
@@ -0,0 +1,334 @@
...
+
+    return entries.sort((a, b) => a.path.localeCompare(b.path))
+}
+
+/** Snapshots are named `<Label>-<project>-<platform>.png` by Playwright. */
+function describe(path) {
+    const file = basename(path, ".png")
+    const m = /^(.*)-(desktop|tablet|mobile)-(.+)$/.exec(file)
+    return m
+        ? { label: m[1].replace(/-/g, " ").trim(), project: m[2], platform: m[3] }
```
flcarre: Minor: `replace(/-/g, " ")` also eats hyphens that belong to the label. `Use-Case-CI-CD-desktop-linux.png` renders as "Use Case CI CD" rather than "Use Case CI-CD".

Cosmetic, but this heading is the only place a reviewer identifies which page moved. Since the script already sits next to `tests/fixtures/page-sample.mjs`, matching the slug against the known `PAGES` / `VISUAL_ONLY_PAGES` labels would be exact and would also give you the page path for free.

playwright.config.ts
elevatebart: You are right and I was wrong, including in the PR description. Confirmed `dist/index.js:431` reads the block and `SnapshotHelper` merges it under the call-site options, with `maxDiffPixelRatio` surviving the `NonConfigProperties` filter.

Restored the block rather than rewording, since that removes the trap instead of documenting it. The tolerance now applies to any `toHaveScreenshotOdiff` call site by default and the per-call duplicate is gone.

Proven both directions locally on a scratch spec with no call-site options: a 0.13% change passes with the block, and fails without it at `1200 pixels (0.13% of all pixels) are different`, which is exactly the `maxDiffPixels: 0` default you described.

Fixed in 6249193.

---

.github/workflows/update-linux-snapshots.yml
elevatebart: Agreed, and confirmed the timeout path: `server.js:207-214` rejects the promise, and nothing in the script catches it, so the process exits non-zero and takes the commit and the PR with it.

Applied your suggested `continue-on-error: true` verbatim in 6249193. The report is a convenience; the regenerated baselines are the product.

The body degradation is now cleaner than empty interpolations too, see the reply on the `report_total` thread.

---

.github/workflows/update-linux-snapshots.yml
elevatebart: Confirmed. The second path arrived with the #5706 merge, after this step was written, so the two drifted apart without anyone noticing.

Took both halves of your suggestion in 6249193: a snapshot-only `has_png_changes` flag now gates the build and upload steps, and the script composes the whole sentence into a single `report_summary` output. On API-only drift, or if the step fails under its new `continue-on-error`, the PR body renders nothing at all rather than a sentence full of blanks.

Verified on an unchanged tree: `report_total=0`, `report_summary=` empty.

---

scripts/visual-diff-report.mjs
elevatebart: Good catch, the numbers were not the gating numbers.

Took the first option, with a bit of the second. Both tolerances now come from one place, `tests/fixtures/screenshot-options.mjs`, imported by `playwright.config.ts` and by this script, so they cannot drift. The report compares at the suite's `threshold: 0.2` / `antialiasing: true`, and the header prints `maxDiffPixelRatio 0.01, threshold 0.2, antialiasing on` so a reader can relate the counts to the gate without going digging.

On the seeded test case, About Us now reports 0.14%, which reads directly against the 1% that decided it.

I did consider keeping `threshold: 0` for a more sensitive diff image, but an uninterpretable headline number is the worse trade. Fixed in 6249193.

---

scripts/visual-diff-report.mjs
elevatebart: Reproduced: `node scripts/visual-diff-report.mjs --out` gave `ERR_INVALID_ARG_TYPE` out of `rmSync`. Applied your fix verbatim in 6249193; the same invocation now falls back to the default and completes.

---

scripts/visual-diff-report.mjs
elevatebart: Took the sample-lookup suggestion rather than patching the regex, since as you say the slug genuinely cannot carry that information: Playwright collapses every non-word run to `-`, so `CI-CD` and `CI CD` slug identically.

The script now builds a slug-to-page map from `PAGES` and `VISUAL_ONLY_PAGES` and keeps the unhyphenating fallback only for a snapshot with no matching entry. `Use-Case-CI-CD-desktop-linux.png` reads "Use Case CI-CD" again, and each entry carries its path as you predicted. Checked that all sample entries round-trip: zero misses.

Fixed in 6249193.

---

.github/workflows/update-linux-snapshots.yml
elevatebart: Your figures reproduce exactly. 185 MB across 180 files, mean 1.03 MB, largest `Resource-Airflow-2-EOL-mobile-linux.png` at 2.9 MB. Scaling by the mean: ~123 MB at 40 changes, ~278 MB at 90, ~555 MB at a full repaint, on top of the 186 MB `visual-snapshots-linux` already uploaded.

Not changing it in this PR, and here is the reasoning rather than just deferring.

Downscaling baseline and current defeats the slider, whose whole value is that you can see a two-pixel shift at native resolution. Building the slider against the snapshot artifact does not survive the thing the report is for either: you download one zip, open `index.html`, and it has to work offline without a second artifact unzipped beside it at a known path.

The 3x is close to inherent. Baseline and current are both needed for a wipe, and the diff is the cheapest of the three (2.5 MB against 4.7 MB for the pair, on the seeded set).

What I would do instead is bound it rather than shrink it uniformly: a byte budget, sorted by diff percentage, that keeps baseline and current for the loudest changes and falls back to diff-only once the budget is spent, with the report saying so. A 40-change refresh stays fully interactive; a full repaint lands around 60 MB and still shows every page that moved. That keeps the good case good instead of degrading it to protect the rare case.

Happy to do it here if you would rather it not merge without a bound, otherwise it is a clean follow-up and I will file it. Your call.

---

PR #5706, opened 2026-09-18
test(visual): stop content and API drift from repainting baselines

elevatebart:
Three things moved the screenshot baselines, and only one of them was the CSS the suite exists to watch. Blog-Index has the most refreshes in the snapshot history, and none of them were about layout.

### Dated content

`SNAPSHOT_CUTOFF`, a new optional server env field, filters the four date-sorted collection reads behind `/blogs` and `/docs`: blogs in `WhatsNew.astro` and `blogs/[category].astro`, external blogs, tutorial videos. `beforeCutoff` is a pass-through when the var is unset, which is everywhere except the snapshot workflow, so the site keeps rendering its full content.

The workflow sets it to `2026-09-17`. A date-only value resolves to midnight UTC, so the one post dated `2026-09-17T10:00:00` falls outside it and the sample runs current through `2026-09-16T13:00`. Pass a full timestamp if you want a day inclusive.

### Live API

`/plugins`, `/blueprints` and the community metrics are read at request time, so a plugin release repainted them. The snapshot workflow now builds and screenshots against `scripts/api-fixture-server.mjs`, which Lighthouse already used, with its own `FIXTURE_PATHS` adding the endpoints the sampled SSR routes need: `/plugins/definitions/*`, `/plugins/metadata*`, `/plugins/pluginsInformation*`, `/plugins/artifacts/*`, `/plugins/*/versions`, `/versions`, `/communities/github/*`.

`isRecorded` grew an inner wildcard matching a single path segment, so `/v1/plugins/<fqcn>/versions` can be recorded without `/v1/plugins*` swallowing the schema fetches sitting next to it. Trailing-`*` prefixes behave exactly as before.

Anything recorded during a run is committed with the refreshed baselines. `workflow_dispatch` gained a `refresh_api_fixtures` box; ticking it clears the whole directory, including the ~300 blueprint detail responses only Lighthouse reads, which its next run re-records into an artifact to commit. Called out in the input description and the fixtures README.

### Incidental content

`/docs` renders the four newest posts through `WhatsNew`, where the blog list is not what that baseline is for. A stylesheet injected via `stylePath` pins the cards to their thumbnail height and blanks the copy, so bumping the cutoff moves Blog-Index and leaves Docs-Landing alone.

### Verified

Full build with a cutoff and the fixture server, exit 0, 274 fixture hits. Post-cutoff content absent from the built `/blogs/all`, `/docs` showing the four newest posts that remain. 404 unit tests pass (3 new, covering the pass-through, the boundary and an unparsable date), oxlint clean.

### What to expect

One refresh PR when this first runs, then quiet until the CSS actually changes or someone bumps the date. Bumping it is the deliberate act, ideally quarterly, alongside the refresh it produces: the trade is that the sampled content ages, so a card redesign that only triggers on newer frontmatter would not show up until the next bump.

https://claude.ai/code/session_01GHRsDvXNS9LLTVhdmjA8or


src/utils/snapshotCutoff.ts
```diff
@@ -0,0 +1,23 @@
...
+    return parsed
+}
+
+// Unset outside the visual-snapshot workflow, where beforeCutoff is a
+// pass-through and the site renders its full content as usual.
+const cutoff = parseCutoff(SNAPSHOT_CUTOFF)
+
+// Drops entries dated after the cutoff. `date` reads the field the caller
+// sorts on, since the collections disagree on its name.
+export function beforeCutoff<T>(entries: T[], date: (entry: T) => Date): T[] {
```
iitzIrFan, line 20: **Non-blocking: two blog lists skip the cutoff.** Neither moves a baseline today, but both will later:

- `src/utils/plugins/fetchPluginPageData.ts:243`: `fetchSecondaryData` loads posts through `loadBlogPostsMetadata()`, and the plugin landing pages render them. No post uses the `plugins:` frontmatter yet, so the list is empty. The first post that does will move the `/plugins/core*` baselines.
- `src/pages/orchestration/[slug].astro:219`: the fallback list shows the 4 newest posts matching the tool name. `/orchestration/ansible` is safe only because `ansible.yaml` has curated `resources`. Sampling another orchestration page without them would bring the drift back.

Wrapping both would make the cutoff cover every blog list the sampled pages can show:
```ts
beforeCutoff(blogs, (b) => new Date(b.data.date))
```

.github/workflows/update-linux-snapshots.yml
```diff
@@ -61,6 +77,24 @@ jobs:
           path: ~/.cache/ms-playwright
           key: playwright-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
 
+      - name: Discard the recorded API responses
+        if: inputs.refresh_api_fixtures
+        run: rm -f tests/fixtures/api/*.json
```
iitzIrFan, line 84: **Non-blocking: this also resets Lighthouse.** It deletes all ~300 `blueprints-*` responses that only Lighthouse reads, and those come back only after someone commits Lighthouse's artifact. Could this delete only the paths in this workflow's own `FIXTURE_PATHS`? If not, maybe add a "don't tick this unless you'll also re-commit the Lighthouse fixtures" warning in the `refresh_api_fixtures` description.

.github/workflows/update-linux-snapshots.yml
```diff
@@ -19,6 +24,17 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+    env:
+      # Midnight UTC cut: content dated after it never reaches a baseline, so a
+      # new post stops repainting /blogs. Bump it with the refresh PR it makes.
+      SNAPSHOT_CUTOFF: "2026-09-17"
```
iitzIrFan, line 30: **Non-blocking: who bumps this?** Since the cutoff needs a deliberate quarterly bump, it would help to name an owner or add a calendar reminder so the sampled content doesn't go stale.

iitzIrFan (approved):
Nice work. This goes after the actual causes of baseline churn, and it fits the existing setup: it reuses the fixture server and startup steps from `lighthouse.yml`, declares `SNAPSHOT_CUTOFF` like the other env settings, and puts `snapshotCutoff.ts` next to its test in the same style as the other utils. I checked out the branch: the new path matching handles exact, trailing-`*` and inner-`*` paths correctly (`/v1/plugins/core` is still not recorded, `/v1/plugins/<fqcn>/versions` is), and the new tests pass.

Approving. The inline notes are non-blocking suggestions, fine here or in a follow-up.

src/utils/snapshotCutoff.ts
elevatebart: Both confirmed and fixed in `b0ba5e8`. Checked your supporting facts before touching anything: zero posts carry `plugins:` frontmatter today, and on the orchestration page `hasCurated` (line 425) wins over `relatedBlogs` (line 434), with `ansible.yaml` supplying `resources.items`. So neither moves a baseline now, and both would later.

`fetchSecondaryData` uses `new Date(b.data.date)` as you wrote it, since `loadBlogPostsMetadata` parses frontmatter with js-yaml rather than through the collection schema and the surrounding sort already coerces. The orchestration page passes `b.data.date` straight through, where `z.date()` has already done it.

One thing I checked before adding the import: `fetchPluginPageData.ts` reaches `astro:env/server` and is bundled into the SSR worker. Nothing under `.vue` imports it, and a full build with the cutoff set came back clean, server bundle included.

---

.github/workflows/update-linux-snapshots.yml
elevatebart: Took the warning option, in `b0ba5e8`. The input description now says it drops the ~300 blueprint detail responses only Lighthouse reads and not to tick it without committing the `api-fixtures` artifact afterwards, and the step carries a comment saying the same.

I looked at scoping the delete and decided against it. Fixture filenames are `<slug>.<sha1-prefix>.json`, so matching them back to this workflow's `FIXTURE_PATHS` means reimplementing the server's naming rule in shell, and it would not separate cleanly anyway: this workflow's paths include `/v1/blueprints*`, so the `blueprints-*` files are partly ours and partly Lighthouse's. A filter that kept them would leave the sampled `/blueprints` responses stale through a refresh, which is the opposite of what the box is for.

If the shared directory becomes a real nuisance, the honest fix is splitting the two jobs' fixtures into sibling directories with `FIXTURE_DIR`, which the server already supports. Happy to do that as a follow-up if you want it.

---

.github/workflows/update-linux-snapshots.yml
elevatebart: Fair, and this one I've left alone: naming an owner is yours to decide, not mine to write into the workflow.

Worth saying the staleness is visible rather than silent. Every refresh PR this job opens states the cutoff in its body, so a date drifting months into the past shows up on each one. If that turns out to be too easy to skim past, the mechanical version is a step that fails the run once the cutoff is older than some threshold, turning the nightly red instead of relying on someone noticing. Say the word and I'll add it, with whatever window you want.

---

elevatebart: Added the mechanical version in `97876e8`, since relying on someone reading the refresh PR body was the weak part.

The run now reports the cutoff's age to the job summary every time, warns past 90 days, and fails past 180. Both windows are env vars on the job (`WARN_DAYS`, `FAIL_DAYS`), so tighten them if a quarter feels too generous.

It runs last on purpose. A neglected cutoff still gets its baselines regenerated and committed before the job goes red, so the nagging never costs you a refresh. It also catches an unparsable date, though the build would already have thrown by then.

Still no owner named in the workflow, that part stays yours.

---

